PR #1, opened 2017-08-17
First pre-release

mexanick:
- Improved repo structure 
- added releasing tools 
- restructured some modules, notably optohybrid
- first testing RPC functions for configuration loading added

PR #2, opened 2017-10-03
Add calibration routines to build threshold scan

cbravo135:
### Types of changes

## Checklist:


src/calibration_routines.cpp
```diff
@@ -1,50 +1,179 @@
...
 };
 
-void *thresholdScanLocal(void * args)
+void dacMonConfLocal(localArgs * la, uint32_t ohN, uint32_t ch)
 {
-  struct scan_args *local_args = (struct scan_args *) args; 
+    writeReg(la->rtxn, la->dbi, "GEM_AMC.GEM_TESTS.VFAT_DAQ_MONITOR.CTRL.RESET", 0x1, la->response);
+    writeReg(la->rtxn, la->dbi, "GEM_AMC.GEM_TESTS.VFAT_DAQ_MONITOR.CTRL.ENABLE", 0x1, la->response);
+    writeReg(la->rtxn, la->dbi, "GEM_AMC.GEM_TESTS.VFAT_DAQ_MONITOR.CTRL.OH_SELECT", ohN, la->response);
+    if(ch==128)
```
bdorney: Are VFAT3 channels indexed [1,128] or are they [0,127] with 128 being the calibration channel?

cbravo135: Configuring the module to use "channel 128" will give an OR behavior of all 128 channels.

src/calibration_routines.cpp
```diff
@@ -1,50 +1,179 @@
...
-void *thresholdScanLocal(void * args)
+void dacMonConfLocal(localArgs * la, uint32_t ohN, uint32_t ch)
 {
-  struct scan_args *local_args = (struct scan_args *) args; 
+    writeReg(la->rtxn, la->dbi, "GEM_AMC.GEM_TESTS.VFAT_DAQ_MONITOR.CTRL.RESET", 0x1, la->response);
+    writeReg(la->rtxn, la->dbi, "GEM_AMC.GEM_TESTS.VFAT_DAQ_MONITOR.CTRL.ENABLE", 0x1, la->response);
+    writeReg(la->rtxn, la->dbi, "GEM_AMC.GEM_TESTS.VFAT_DAQ_MONITOR.CTRL.OH_SELECT", ohN, la->response);
+    if(ch==128) 
+    {
+        writeReg(la->rtxn, la->dbi, "GEM_AMC.GEM_TESTS.VFAT_DAQ_MONITOR.CTRL.VFAT_CHANNEL_SELECT", 0, la->response);
```
bdorney, line 12: I see here the "channel" argument is 0, but ch in the if statement is checked against 128.  Why do we pass 0 here instead of 128?

cbravo135: It doesn't matter what VFAT_CHANNEL_SELECT is since VFAT_CHANNEL_GLOBAL_OR is written to 1.

src/calibration_routines.cpp
```diff
@@ -1,50 +1,179 @@
...
+
+uint32_t vfatSyncCheckLocal(localArgs * la, uint32_t ohN)
+{
+    uint32_t goodVFATs = 0;
+    for(int vfatN = 0; vfatN < 24; vfatN++)
+    {
+        char regBase [100];
+        sprintf(regBase, "GEM_AMC.OH_LINKS.OH%i.VFAT%i",ohN, vfatN);
+        bool linkGood = readReg(la->rtxn, la->dbi, std::string(regBase)+".LINK_GOOD");
+        uint32_t linkErrors = readReg(la->rtxn, la->dbi, std::string(regBase)+".SYNC_ERR_CNT");
```
bdorney: Can we have `regBase`, `linkGood`, and `linkErrors` declared outside this loop (just below `goodVFATs`) rather than allocating the memory each iteration?

We can have them reset after line 38 if needed.

cbravo135: This is compiled C and not python, the memory is allocated on the stack at compilation anyways. These are just local buffers for building the strings needed to use the address table, it is important to think of scope of variables in C.

src/calibration_routines.cpp
```diff
@@ -1,50 +1,179 @@
 #include <pthread.h>
 #include "optohybrid.h"
 
-struct scan_args {
-  lmdb::txn & rtxn;
-  lmdb::dbi & dbi;
-  RPCMsg *response;
+struct localArgs {
```
bdorney: This is declared as a global variable, is it possible that we put this, and all other function definitions here, inside of a namespace, e.g. `calibration`?

cbravo135: This is C, so no namespaces.

src/calibration_routines.cpp
```diff
@@ -1,50 +1,179 @@
...
+{
+    auto env = lmdb::env::create();
+    env.set_mapsize(1UL * 1024UL * 1024UL * 40UL); /* 40 MiB */
+    env.open("/mnt/persistent/texas/address_table.mdb", 0, 0664);
+    auto rtxn = lmdb::txn::begin(env, nullptr, MDB_RDONLY);
+    auto dbi = lmdb::dbi::open(rtxn, nullptr);
+
+    uint32_t ohN = request->get_word("ohN");
+
+    struct localArgs la = {.rtxn = rtxn, .dbi = dbi, .response = response};
```
bdorney: At line 4 you've declared `localArgs` as a struct globally, why are you re-declaring locally a `localArgs` struct?

cbravo135: It is not instantiated at line 4, just defined. https://www.tutorialspoint.com/cprogramming/c_structures.htm

src/calibration_routines.cpp
```diff
@@ -1,50 +1,179 @@
...
+        sprintf(regBase, "GEM_AMC.OH_LINKS.OH%i.VFAT%i",ohN, vfatN);
+        bool linkGood = readReg(la->rtxn, la->dbi, std::string(regBase)+".LINK_GOOD");
+        uint32_t linkErrors = readReg(la->rtxn, la->dbi, std::string(regBase)+".SYNC_ERR_CNT");
+        goodVFATs = goodVFATs | ((linkGood && (linkErrors == 0)) << vfatN);
+    }
+
+    return goodVFATs;
+}
+
+void vfatSyncCheck(const RPCMsg *request, RPCMsg *response)
```
bdorney: I'm missing why this function is needed in addtion to `vfatSyncCheckLocal(...)`?  Is that because `vfatSyncCheckLocal(...)` local doesn't take RPCMsg request/response as arguments?  It's my ignorance but what is `vfatSyncCheck(...)` doing that `vfatSyncCheckLocal(...)` isn't?  Or how do they fit together in the bigger picture?

cbravo135: The non-local version of each function can be called externally by the host machine. The local version exists so we can use them locally without having to parse the address table again.

bdorney: I guess I have the same question about all the other function pairs below:

- `ttcGenConfLocal(...)`, `ttcGenConf(...)`
- `thresholdScanLocal(...)`, `thresholdScan(...)`,
- etc...

I guess there's some architecture reason/design that I missed from last night?

src/calibration_routines.cpp
```diff
@@ -1,50 +1,179 @@
...
+    uint32_t notmask = ~mask & 0xFFFFFF;
+    if( (notmask & goodVFATs) != notmask) 
+    { 
+        char errBuf[200];
+        sprintf(errBuf,"One of the unmasked VFATs is not Synced. goodVFATs: %x\tnotmask: %x",goodVFATs,notmask);
+        response->set_string("error",errBuf); 
+        return;
+    }
+    char regName[100];
+    for(int vfatN = 0; vfatN < 24; vfatN++) if((notmask >> vfatN) & 0x1)
```
bdorney: for coding style/readability can we have these on two separate lines with braces modified appropriately.

src/calibration_routines.cpp
```diff
@@ -1,50 +1,179 @@
...
+    { 
+        sprintf(regName,"GEM_AMC.OH.OH%i.GEB.VFAT%i.CFG_RUN",ohN,vfatN); 
+        writeReg(la->rtxn, la->dbi, regName, 0x1, la->response);
+        sprintf(regName,"GEM_AMC.OH.OH%i.GEB.VFAT%i.CFG_THR_ARM_DAC",ohN,vfatN); 
+        writeReg(la->rtxn, la->dbi, regName, 0, la->response);
+    }
+    for(uint32_t dacVal = dacMin; dacVal <= dacMax; dacVal += dacStep)
+    {
+        for(int vfatN = 0; vfatN < 24; vfatN++) if((notmask >> vfatN) & 0x1)
```
bdorney: same comment as above, split lines.

src/calibration_routines.cpp
cbravo135: This style of C is legitimate. This is easy to read as for every vfat not masked. I really don't like tons of extra unnecessary braces.

src/calibration_routines.cpp
cbravo135: I argue this is more readable than your proposal.

src/calibration_routines.cpp
```diff
@@ -1,50 +1,179 @@
...
+        writeReg(la->rtxn, la->dbi, "GEM_AMC.GEM_TESTS.VFAT_DAQ_MONITOR.CTRL.RESET", 0x1, la->response);
+        writeReg(la->rtxn, la->dbi, "GEM_AMC.TTC.GENERATOR.CYCLIC_START", 0x1, la->response);
+        bool running = true;
+        while(running) running = readReg(la->rtxn, la->dbi, "GEM_AMC.TTC.GENERATOR.CYCLIC_RUNNING");
+        for(int vfatN = 0; vfatN < 24; vfatN++)
+        { 
+            char regBuf[200];
+            int idx = vfatN*(dacMax-dacMin+1)/dacStep+(dacVal-dacMin)/dacStep;
+            sprintf(regBuf,"GEM_AMC.GEM_TESTS.VFAT_DAQ_MONITOR.VFAT%i.GOOD_EVENTS_COUNT",vfatN);
+            outData[idx] = readRawReg(la->rtxn, la->dbi, regBuf, la->response);
```
bdorney: Can we add a check somewhere that `outData` has this many elements?  e.g. This function should ensure the user passes it a pointer to an array of appropriate size.

src/calibration_routines.cpp
```diff
@@ -1,50 +1,179 @@
...
+    uint32_t dacStep = request->get_word("dacStep");
 
-  pthread_t thread;
-  configureScanModule(rtxn, dbi, request, response);
-  struct scan_args args = {.rtxn = rtxn, .dbi = dbi, .response = response};
-  pthread_create(&thread,NULL,&thresholdScanLocal,(void *)&args);
+    struct localArgs la = {.rtxn = rtxn, .dbi = dbi, .response = response};
+    uint32_t outData[24*(dacMax-dacMin+1)/dacStep];
+    thresholdScanLocal(&la, outData, ohN, mask, ch, nevts, dacMin, dacMax, dacStep, response);
+    response->set_word_array("data",outData,24*(dacMax-dacMin+1)/dacStep);
```
bdorney: How does the user access `outData`?  It looks like Line 155 is written `outData` to a buffer?  Where/how does the user get this?

src/calibration_routines.cpp
cbravo135: See line 153.

src/calibration_routines.cpp
jsturdy: The complaint is **not** about useless braces, rather about the "one statement per line" readability philosophy:
```c
for(int vfatN = 0; vfatN < 24; vfatN++)
  if((notmask >> vfatN) & 0x1) {
...
  } // closes if
```

src/calibration_routines.cpp
```diff
@@ -1,50 +1,179 @@
...
 #include "optohybrid.h"
 
-struct scan_args {
-  lmdb::txn & rtxn;
-  lmdb::dbi & dbi;
-  RPCMsg *response;
+struct localArgs {
+    lmdb::txn & rtxn;
+    lmdb::dbi & dbi;
+    RPCMsg *response;
```
bdorney: Can we stick with a white space convention here like we do with our other repo's?  e.g. looks like some `\t` characters are present here.

src/calibration_routines.cpp
cbravo135: I am still trying to get Misha to explain how that magic part works. The user will never access outData but rather what is packaged into the "data" word_array placed in response.

src/calibration_routines.cpp
```diff
@@ -1,50 +1,179 @@
...
+    response->set_word_array("data",outData,24*(dacMax-dacMin+1)/dacStep);
 
 }
 
 void *latencyScanLocal(void * args)
 {
-  struct scan_args *local_args = (struct scan_args *) args; 
-  startScanModule(local_args->rtxn, local_args->dbi, local_args->response);
+    struct localArgs *local_args = (struct localArgs *) args; 
+    startScanModule(local_args->rtxn, local_args->dbi, local_args->response);
```
bdorney: Where is `startScanModule(...)` defined?  Why is it now used in `thresholdScanLocal(...)`?

src/calibration_routines.cpp
```diff
@@ -1,50 +1,179 @@
...
+    auto rtxn = lmdb::txn::begin(env, nullptr, MDB_RDONLY);
+    auto dbi = lmdb::dbi::open(rtxn, nullptr);
 
-  pthread_t thread;
-  configureScanModule(rtxn, dbi, request, response);
-  struct scan_args args = {.rtxn = rtxn, .dbi = dbi, .response = response};
-  pthread_create(&thread,NULL,&latencyScanLocal,(void *)&args);
+    pthread_t thread;
+    configureScanModule(rtxn, dbi, request, response);
+    struct localArgs args = {.rtxn = rtxn, .dbi = dbi, .response = response};
```
bdorney: Same comment as above regarding local vs. global `localArgs`

src/calibration_routines.cpp
cbravo135: This is not operational code, it is historic from when Misha was putting this initial version of this. The latency scan code does not work yet.

bdorney (changes requested):
Changes requested and some clarifications needed.

src/calibration_routines.cpp
cbravo135: I will point out it is not used int threhsoldScanLocal, but rather latencyScanLocal.

src/calibration_routines.cpp
cbravo135: This is C, don't know what else to tell you. It was never instantiated globally as you are implying.

src/calibration_routines.cpp
bdorney: Let's add a comment that states this here

src/calibration_routines.cpp
bdorney: In that case can we remove line 17 entirely so that only the `VFAT_CHANNEL_GLOBAL_OR` is present?

src/calibration_routines.cpp
cbravo135: Def'n no \t char here. I changed the spacing to be 4 instead of 2 for the whole file, that is what you are seeing.

src/calibration_routines.cpp
bdorney: I know the scope of the variables here. I'm asking to make this more efficient rather than having to allocate memory every iteration

src/calibration_routines.cpp
bdorney: Okay I see...so we should view the non-local version as executed only on the CTP7?  And the non-local version as what would be executed from the DAQ PC?

For this system we write software that is executed from the DAQ PC and then is run on the CTP7?  Then the user never calls the local versions?

src/calibration_routines.cpp
bdorney: Sorry misread the line, didn't notice the `la =` part

src/calibration_routines.cpp
bdorney: I think this comes back to my comment above, how is `thresholdScanLocal(...)` used: 

1. Is it only ever called by `thresholdScan(...)`?  
2. Or is there some usage case where it would be called standalone?

If the usage case is #1 then yeah line 153 is sufficient, however if there would be a case where it is called standalone then a check of the array size is a good idea.

src/calibration_routines.cpp
bdorney: Okay got it.  We wait for @mexanick.

src/calibration_routines.cpp
bdorney: Okay.  Suggest to add a comment that states this, or probably better to comment it out/delete it for now.

src/calibration_routines.cpp
bdorney: Yeah, sorry I misread the line, didn't notice the `args =` part

src/calibration_routines.cpp
cbravo135: The local version exists so we can use them locally without having to parse the address table again. Parsing the address table takes several seconds.

cbravo135: It is impossible for the user to call the local versions, they are not registered in the library. See the end of the file to see the registered functions.

src/calibration_routines.cpp
cbravo135: The memory is allocated on the stack at compilation, and is only accessed each iteration. Your proposal would be not be any more efficient, and if ever reset for some reason it would become less efficient.

src/calibration_routines.cpp
cbravo135: Evaldas recommended to write it to 0 as I have done in this case.

src/calibration_routines.cpp
cbravo135: "for every vfat not masked" is a single statement expressed by this one line.

src/calibration_routines.cpp
cbravo135: We need a comment to say this library is a work in progress and is not complete yet?

cbravo135:
This code definitely can be cleaned and optimized. This PR is about functionality. If you don't want this code because of where I declare variables, it doesn't matter to me. I still have the code and can run the scans myself to get results, just thought others might want to have this new functionality.

cbravo135:
I would call this code more middleware than software. This goal here is to make it easier for the user to interact with the FW on the CTP7. We no longer have a scan module in fw for V3, so this piece of middleware that runs on the CTP7 will have to take the place of that fw module. The goal here is to provide a library for easy access by the developer of the user system. Modifying this is easier than modifying fw, but not as easy as changing the over arching python code. This is meant to be a module the python code can call, no DOC should ever directly interact with this code. We don't ask the DOCs to look into the fw do we?

src/calibration_routines.cpp
mexanick: Well, to be clear: what you guys say depends on how do you compile and which compilation flags are involved. If your compiler optimizes the code (what's done in this case), then Cameron's comment is spot on. Otherwise there could be options. 
Although this code can benefit from certain refactoring, I would postpone it to the future PRs.

src/calibration_routines.cpp
mexanick: I support Cameron in this case. No need for few extra lines and aligned blocks.

src/calibration_routines.cpp
mexanick: the thresholdScanLocal method can not be called standalone, it doesn't have appropriate interface for this. However, it can be used as a sub-call of other methods, not limiting to threshold scan only.
There's another reason for a such architecture: for the moment all the methods are so-called blocking transactions. I.e. you block and RPC connection during execution time. The goal for future is to spawn detached threads for the calibration routines (executing those local methods) and return the interfacing RPC call ASAP.

src/calibration_routines.cpp
mexanick: The communication between host module and remote method is based on exchange of RPCMsg objects. There're two such objects: _request_ and _response_. You can find the description of the RPCMsg class here: https://github.com/cms-gem-daq-project/ctp7_modules/blob/master/include/wiscRPCMsg.h#L1

src/calibration_routines.cpp
mexanick: @bdorney , the templates for functions which are not implemented aims V2 system design. Actually, certain things has been implemented and work. I can remove them, but I need a management decision that **we are not going to use RPC calls for V2 setup**

src/calibration_routines.cpp
jsturdy: `for` and `if` are functionally two separate statements; that they combine in your use case to provide some logical construct is irrelevant.
Code readability is **critical** for maintainability, especially in the case that the original developer leaves; please don't underestimate this

src/calibration_routines.cpp
jsturdy: I would stress that adding in RPC calls for V2 has always been "in the plan", as we have an actual system where it can be fully tested, and which will ensure that things can be as compatible as possible between V2 and V3 (which has also, always been the plan)

src/calibration_routines.cpp
bdorney: > Although this code can benefit from certain refactoring, I would postpone it to the future PRs.

Since we will be building on this as we move forward in the long run it would save time to commit things right the first time.  As I have said many times, in general we shouldn't do things that only benefit us this week.  

If the group feels this is not an issue then okay.

src/calibration_routines.cpp
bdorney: kk my mistake

src/calibration_routines.cpp
bdorney: Okay I understand, thanks to you both for the clarification (here and in the comment above).

src/calibration_routines.cpp
bdorney: I guess I'm still confused on how you write a script on the local PC that executes all this and reads back the data.  Maybe we are not there yet though?

src/calibration_routines.cpp
bdorney: My mistake.  Thanks for pointing this out.

src/calibration_routines.cpp
bdorney: I did not realize this was supposed to be for both v2 and v3.  Maybe it's better to have two different branches?  If this code is working for the v2b system but not yet for the v3 I think it's better to have this as a different branch and have this commented out/missing.  But if the group wants a different approach then no problem.

mexanick (approved):

src/calibration_routines.cpp
```diff
@@ -1,50 +1,147 @@
...
+        sprintf(regBuf,"One of the unmasked VFATs is not Synced. goodVFATs: %x\tnotmask: %x",goodVFATs,notmask);
+        la->response->set_string("error",regBuf); 
+        return;
+    }
+
+    if(ch >= 128)
+    {
+        if(enCal)
+        {
+            la->response->set_string("error","Brian, it doesn't make sense to calpulse all channels");
```
mexanick, line 72: Lol. Don't personalize the error message :D

src/calibration_routines.cpp
```diff
@@ -1,50 +1,147 @@
...
+
+    struct localArgs la = {.rtxn = rtxn, .dbi = dbi, .response = response};
+    ttcGenConfLocal(&la, L1Ainterval, pulseDelay);
+
+    return;
 
 }
 
-void *latencyScanLocal(void * args)
+void genScanLocal(localArgs *la, uint32_t *outData, uint32_t ohN, uint32_t mask, uint32_t ch, uint32_t enCal, uint32_t nevts, uint32_t dacMin, uint32_t dacMax, uint32_t dacStep, std::string scanReg)
```
mexanick, line 51: This function will benefit from logging. Would be nice to log at INFO level what we are scanning, limits and step.

src/calibration_routines.cpp
```diff
@@ -1,50 +1,147 @@
...
+    writeReg(la->rtxn, la->dbi, "GEM_AMC.GEM_SYSTEM.VFAT3.VFAT3_RUN_MODE", 0x1, la->response);
+    writeReg(la->rtxn, la->dbi, "GEM_AMC.TTC.GENERATOR.SINGLE_RESYNC", 0x1, la->response);
+    dacMonConfLocal(la, ohN, 0);
+    uint32_t goodVFATs = vfatSyncCheckLocal(la, ohN);
+    uint32_t notmask = ~mask & 0xFFFFFF;
+    char regBuf[200];
+    if( (notmask & goodVFATs) != notmask) 
+    { 
+        sprintf(regBuf,"One of the unmasked VFATs is not Synced. goodVFATs: %x\tnotmask: %x",goodVFATs,notmask);
+        la->response->set_string("error",regBuf);
```
mexanick, line 64: When error is detected I think it would be also good to log it at the same time. Here and in other places as well.

mexanick (changes requested):

include/vfat3.h
```diff
@@ -0,0 +1,95 @@
...
+    { 
+        char errBuf[200];
+        sprintf(errBuf,"One of the unmasked VFATs is not Synced. goodVFATs: %x\tnotmask: %x",goodVFATs,notmask);
+        la->response->set_string("error",errBuf); 
+        return;
+    }
+
+    for(uint32_t vfatN = 0; vfatN < 24; vfatN++) if((notmask >> vfatN) & 0x1)
+    {
+        std::string configFileBase = "/mnt/persistent/gemdaq/vfat3/config_OH"+std::to_string(ohN)+"_VFAT"+std::to_string(vfatN)+".txt";
```
mexanick, line 54: I would give some more freedom to the config file name/location. We may have few different configuration sets stored at the CTP7 (like in V2 we have different ztrims)

bdorney: Can we have the linux partition of the CTP7 mount an external location?  In this way we could use the identical structure to what we have now, e.g. `$DATA_PATH/configs`.

src/calibration_routines.cpp
bdorney: What are the list of possible values for `scanReg`?  I see this is designed to be an arbitrary scan application.  More of a syntax question. e.g.

- For VT1 `scanReg` = ???
- For latency `scanReg` = ??? 
- etc...

src/calibration_routines.cpp
```diff
@@ -1,50 +1,147 @@
 #include <pthread.h>
 #include "optohybrid.h"
+#include "vfat3.h"
 
-struct scan_args {
-  lmdb::txn & rtxn;
-  lmdb::dbi & dbi;
-  RPCMsg *response;
-};
+void dacMonConfLocal(localArgs * la, uint32_t ohN, uint32_t ch)
```
bdorney, line 5: Did we drop the non-local version of this function?

cbravo135: No it just lives in vfat3.h now.

src/calibration_routines.cpp
```diff
@@ -1,50 +1,147 @@
...
 
-void *latencyScanLocal(void * args)
+void genScanLocal(localArgs *la, uint32_t *outData, uint32_t ohN, uint32_t mask, uint32_t ch, uint32_t enCal, uint32_t nevts, uint32_t dacMin, uint32_t dacMax, uint32_t dacStep, std::string scanReg)
 {
-  struct scan_args *local_args = (struct scan_args *) args; 
-  startScanModule(local_args->rtxn, local_args->dbi, local_args->response);
+
+    writeReg(la->rtxn, la->dbi, "GEM_AMC.TTC.GENERATOR.CYCLIC_L1A_COUNT", nevts, la->response);
+    writeReg(la->rtxn, la->dbi, "GEM_AMC.GEM_SYSTEM.VFAT3.VFAT3_RUN_MODE", 0x1, la->response);
+    writeReg(la->rtxn, la->dbi, "GEM_AMC.TTC.GENERATOR.SINGLE_RESYNC", 0x1, la->response);
```
bdorney, line 56: Lines 54 why do we have this here instead of a call of `ttcGenConfLocal(...)`?  Is the user supposed to call `ttcGenConf(...)` before calling `genScan(...)`?

Line 56 why is this not in `ttcGenConfLocal(...)`?

include/vfat3.h
cbravo135: We need to meet and discuss this. Using CTP7 Zynq resources is a limited resource, and we need to agree on a philosophy about how we will use it. I think of it equivalent to an embedded system, and in that case all of this "software" (which I would rather call middleware) would be part of the "firmware" code. This is not really an embedded system, and it is able to run linux. I still think we should treat it as a glorified embedded system and all of this code should be super minimalist and leave the heavy lifting to the host computer. This softens up the fw on the CTP7 and makes it more usable by the sw layer on the host computer. We need to be careful what we ask the CTP7 to do. Adding a network mount, at least to me, is completely out of the question.

src/calibration_routines.cpp
bdorney: Was it renamed?  I don't see it.  Or maybe you made changes that weren't pushed to your remote yet?

src/calibration_routines.cpp
cbravo135: yes, just as it is done now in all of the scan sw.

bdorney: Okay, thanks.  What about line 56?

src/calibration_routines.cpp
cbravo135: Ah no, this configures the daq monitoring, I should rename it. This does not have a non-local version on purpose, only methods in this package should ever be calling this.

cbravo135: vfat configuration is now configureVFAT3s in vfat3.h

src/calibration_routines.cpp
cbravo135: Why would it be in ttcGenConfLocal? The question doesn't make any sense, we are sending a resync right before we start to run the actual scan, this is what we want.

cbravo135: It doesn't make sense to send a resync when we configure the ttc generator, why would we do that? We want this here, it isn't going away. Evaldas and I discussed this exact part of the procedure for at least 15 minutes last night.

bdorney: I don't know, that's why I'm asking.  What does a single resync here do and why is it appropriate here and not in `ttcGenConfLocal(...)`?

bdorney: I'm not asking for it to go away, I'm asking to understand what it does.

src/calibration_routines.cpp
cbravo135: VT1 no longer exists. This can be literally any dac on the VFAT3, so please refer to his address table if you want to know the names of the dacs on the VFAT3.

bdorney: Yeah I understood it could be any DAC on the VFAT3, do you have the link/filepath of the address table?  Thanks.

bdorney: What does `enCal` do or supposed to represent?  It doesn't look it's really used yet.  I guess a future feature?  What is the goal?

include/vfat3.h
mexanick: Network mount is not an option. Not because we can't do this or it is "too heavy for Zynq", but because it is slow by construction. Also we have more than enough space to keep couple of versions of the config. However not ALL the config versions will be kept there, just the ones we consider as often-used.

src/calibration_routines.cpp
cbravo135: Yeah, Evaldas and I were just trying to save time by not putting the ctp7 logging everywhere.

include/vfat3.h
bdorney: One thought: what about making a `struct` for the VFAT3 parameters if one does not already exist?  

Then `configureVFAT3sLocal(...)` could fill this struct by default from the input file but then additional arguments (words) could be added to `configureVFAT3sLocal` (`configureVFAT3s`) to allow for more configurable parameters at runtime from the host PC?

mexanick:
I'm going to do the forces merge of this PR. Please open issues/questions to discuss the remaning points. The reason of such action is that we need a quick code turnaround to avoid huge PRs.

mexanick (approved):

PR #13, opened 2017-11-06
New Feature: ultra scan modules for v2b electronics

bdorney:
## Description

I have expanded the scan tools for v3 electronics to also work with v2b electronics.  The modules identify at runtime what behavior should be executed.

### Types of changes

## Motivation and Context

If we are moving toward RPC/XHAL framework for CTP7 we should have the tools for both v2b and v3 electronics rather than having to move back and forth between libraries.

## How Has This Been Tested?

A complete working example is shown here:
http://cmsonline.cern.ch/cms-elog/1020178

For the v2b electronics I took a per channel threshold scan using the ipbus/uhal framework and then compared the results obtained when taking data with the rpc/xhal tools shown here.  As can be seen the behavior is identical.

Additionally as shown the tools also work on v3 electronics.

See attached plots in above elog

## Checklist:

### DIFF
--- a/include/optohybrid.h
+++ b/include/optohybrid.h
@@ -206,46 +206,92 @@ void configureVFATs(const RPCMsg *request, RPCMsg *response) {
   rtxn.abort();
 }
 
-/*
- *     Configure the firmware scan controller
-*      mode: 0 Threshold scan
-*            1 Threshold scan per channel
-*            2 Latency scan
-*            3 s-curve scan
-*            4 Threshold scan with tracking data
-*      vfat: for single VFAT scan, specify the VFAT number
-*            for ULTRA scan, specify the VFAT mask
- */
-void configureScanModule(lmdb::txn & rtxn, lmdb::dbi & dbi, const RPCMsg *request, RPCMsg *response)
-{
-  std::string oh_number = request->get_string("oh_number");
-  uint32_t mode = request->get_word("mode");
-  uint32_t vfat_mask = request->get_key_exists("vfat_mask")?request->get_word("vfatN"):0xFFFFFFFF;
-  bool ultra = request->get_key_exists("ultra")?true:false;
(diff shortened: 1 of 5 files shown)


bdorney:
I've decided to do some tests with a latency scan to make sure that no changes are needed when using an external trigger source.

include/optohybrid.h
```diff
@@ -206,48 +206,338 @@ void configureVFATs(const RPCMsg *request, RPCMsg *response) {
...
-  uint32_t stepsize = request->get_word("stepsize");
-  uint32_t numtrigs = request->get_word("numtrigs");
-  uint32_t channel = request->get_word("channel");
-  std::string scanBase = "GEM_AMC.OH.OH" + oh_number + ".ScanController";
-  (ultra)?scanBase += ".ULTRA":scanBase += ".THLAT";
-  // check if another scan is running
-  if (readRawReg(rtxn, dbi, scanBase + ".MONITOR.STATUS", response) > 0) {
-    LOGGER->log_message(LogManager::WARNING, stdsprintf("%s: Scan is already running, not starting a new scan", scanBase.c_str()));
-    response->set_string("error", "Scan is already running, not starting a new scan");
+void configureScanModuleLocal(localArgs * la, uint32_t ohN, uint32_t vfatN, uint32_t scanmode, bool useUltra, uint32_t mask, uint32_t ch, uint32_t nevts, uint32_t dacMin, uint32_t dacMax, uint32_t dacStep){
```
mexanick, line 209: What is the reason to put the comments block **inside** the function rather then **outside**? I would strongly insist on following the doxygen commenting style as described in [1]. I doubt doxygen will correctly recognize the comment as it is written below.
[1] https://www.stack.nl/~dimitri/doxygen/manual/docblocks.html#memberdoc

mexanick (approved):

PR #15, opened 2018-01-16
New Feature: SBIT Rate Scan vs. Arbitrary Scan Register

bdorney:
## Description

The sbit rate scan code that was originally written by Evaldas was done purely in python and could only be reliably executed from the CTP7 itself.  This PR includes an RPC module that provides functionality for performing an SBIT rate scan.

### Types of changes

## Motivation and Context

Allows an SBIT rate scan against an arbitrary scan register, the following table describes the input arguments:

| Argument | Type | Description |
| ---------- | ---- | ------------|
| `la` | `localArgs` | Struct containing the `rtxn`, `dbi` and RPC `response` objects. |
| `outDataDacVal` | `uint32_t *` | Pointer to an array that will hold the DAC values of each scan point of `scanReg`.  Note the size of the array is assumed to be `(dacMax - daxMin + 1) / dacStep`.  The `idx` position of this array represents the same scan point as the `outDataTrigRate` argument. |
| `outDataTrigRate` | `uint32_t *` | As `outDataDacVal` but for the y-values of the array. |
| `ohN` | `uint32_t` | Link on CTP7 to be scanned. | 
| `maskOh` | `uint32_t` | This is a 24 bit number that is used as the mask for  `GEM_AMC.OH.OHX.TRIG.CTRL.VFAT_MASK`.  The VFAT to be scanned `vfatN` is determined from this mask and it is assumed to be the only unmasked VFAT. |
| `ch` | `uint32_t` | Channel of `vfatN` to be used, all other channels will be masked.  The original mask settings will be returned after the scan has completed.  Expected to be in the range [0,127].  Note if `ch=128` then the channel mask settings will not be altered and the rate measurement will use the `OR` of all channels. |
| `dacMin` | `uint32_t` | Starting value of `scanReg`.|
| `dacMax` | `uint32_t` | Final value of `scanReg`.|
| `dacStep` | `uint32_t` | Step size used to move from `dacMin` to `dacMax` of `scanReg`.|
| `scanReg` | `std::string` | Register to be scanned, the leading `CFG_` of the register name is omitted when passing to this argument. |
| `waitTime` | `uint32_t` | The time the scan will spend measuring the rate at each scan point. |

Right now the code will only perform a scan using one VFAT of one link.  However discussing with @andrewpeck and @evka85 on how this could be parallelized at least inside of one link (e.g. 24 VFATs simultaneously).

## How Has This Been Tested?

I made measurements using [sbitThreshScanSeries.py](https://github.com/bdorney/vfatqc-python-scripts/blob/qualify-v3-el/sbitThreshScanSeries.py)

ratesummary

Please note in the above scan only VFATs: 0-2, 7-10, 16-18 were unmasked.  The remaining VFATs were masked.

## Checklist:

### DIFF
--- a/src/calibration_routines.cpp
+++ b/src/calibration_routines.cpp
@@ -497,24 +497,29 @@ void sbitRateScanLocal(localArgs *la, uint32_t *outDataDacVal, uint32_t *outData
     if(!((notmask >> maskOh) & 0x1)){
         sprintf(regBuf,"The requested VFAT is not synced; goodVFATs: %x\t requested VFAT: %x", goodVFATs, maskOh);
         la->response->set_string("error",regBuf);
-        return
+        return;
     }
 
     //Determine which VFAT we are using;
-    uint32_t vfatN;
+    uint32_t vfatN=1000;
     for (int vfat=0; vfat<24; ++vfat){
-        if ( 1 == ( (maskOh >> bit) & 1)){
+        if ( 1 == ( (maskOh >> vfat) & 1)){
             vfatN=vfat;
             break;
         }
+    }
+    if( vfatN > 23 ){
+        sprintf(regBuf,"I did not find a VFAT to perform this scan on, maskOh: %x", maskOh);
+        la->response->set_string("error",regBuf);
+        return;
     }
 
     //If ch!=128 store the original channel mask settings
     //Then mask all other channels except for channel ch
     std::map<uint32_t, uint32_t> map_chanOrigMask; //key -> reg addr; val -> reg value
     if( ch != 128){
-        chanMaskAddr[128];
-        for(int chan=0; chan<128; ++chan){ //Loop Over All Channels
+        uint32_t chanMaskAddr[128];
+        for(unsigned int chan=0; chan<128; ++chan){ //Loop Over All Channels
             uint32_t chMask = 1;
             if ( ch == chan){ //Do not mask the channel of interest
                 chMask = 0;
@@ -532,28 +537,28 @@ void sbitRateScanLocal(localArgs *la, uint32_t *outDataDacVal, uint32_t *outData
 
     //Get the scanAddress
     sprintf(regBuf,"GEM_AMC.OH.OH%i.GEB.VFAT%i.CFG_%s",ohN,vfatN,scanReg.c_str());
-    scanDacAddr = getAddress(la->rtxn, la->dbi, regBuf, la->response);
+    uint32_t scanDacAddr = getAddress(la->rtxn, la->dbi, regBuf, la->response);
 
     //Get the OH Rate Monitor Address
     sprintf(regBuf,"GEM_AMC.TRIGGER.OH%i.TRIGGER_RATE",ohN);
-    ohTrigRateAddr = getAddress(la->rtxn, la->dbi, regBuf, la->response);
+    uint32_t ohTrigRateAddr = getAddress(la->rtxn, la->dbi, regBuf, la->response);
 
     //Take the VFATs out of slow control only mode
     writeReg(la->rtxn, la->dbi, "GEM_AMC.GEM_SYSTEM.VFAT3.VFAT3_RUN_MODE", 0x1, la->response);
 
     //Place chip in run mode
     sprintf(regBuf,"GEM_AMC.OH.OH%i.GEB.VFAT%i.CFG_RUN",ohN,vfatN);
-    runAddr = getAddress(la->rtxn, la->dbi, regBuf, la->response);
+    uint32_t runAddr = getAddress(la->rtxn, la->dbi, regBuf, la->response);
     writeRawAddress(runAddr, 0x1, la->response);
 
     //Loop from dacMin to dacMax in steps of dacStep
-    for(uint32_t dacVal = dacMin; dacVal <= daxMax; dacVal += dacStep){
+    for(uint32_t dacVal = dacMin; dacVal <= dacMax; dacVal += dacStep){
         writeRawAddress(scanDacAddr, dacVal, la->response);
         std::this_thread::sleep_for(std::chrono::seconds(3));
 
         int idx = (dacMax-dacMin+1)/dacStep;
         outDataDacVal[idx] = dacVal;
-        outDataTrigRate[idx] = readRawAddress(ohTrigRateAddr);
+        outDataTrigRate[idx] = readRawAddress(ohTrigRateAddr, la->response);
     } //End Loop from dacMin to dacMax
 
     //Take chip out of run mode
@@ -566,7 +571,7 @@ void sbitRateScanLocal(localArgs *la, uint32_t *outDataDacVal, uint32_t *outData
         }
     } //End restore original channel masks if specific channel was requested
 
-    return
+    return;
 } //End sbitRateScanLocal(...)
 
 void sbitRateScan(const RPCMsg *request, RPCMsg *response){
@@ -584,6 +589,7 @@ void sbitRateScan(const RPCMsg *request, RPCMsg *response){
     uint32_t dacStep = request->get_word("dacStep");
     std::string scanReg = request->get_string("scanReg");
 
+    struct localArgs la = {.rtxn = rtxn, .dbi = dbi, .response = response};
     uint32_t outDataDacVal[(dacMax-dacMin+1)/dacStep];
     uint32_t outDataTrigRate[(dacMax-dacMin+1)/dacStep];
     sbitRateScanLocal(&la, outDataDacVal, outDataTrigRate, ohN, maskOh, ch, dacMin, dacMax, dacStep, scanReg);
--- a/src/calibration_routines.cpp
+++ b/src/calibration_routines.cpp
@@ -476,7 +476,7 @@ void genScan(const RPCMsg *request, RPCMsg *response)
     return;
 }
 
-void sbitRateScanLocal(localArgs *la, uint32_t *outDataDacVal, uint32_t *outDataTrigRate, uint32_t ohN, uint32_t maskOh, uint32_t ch, uint32_t dacMin, uint32_t dacMax, uint32_t dacStep, std::string scanReg, uint32_t waitTime){
+void sbitRateScanLocal(localArgs *la, uint32_t *outDataDacVal, uint32_t *outDataTrigRate, uint32_t ohN, uint32_t maskOh, bool invertVFATPos, uint32_t ch, uint32_t dacMin, uint32_t dacMax, uint32_t dacStep, std::string scanReg, uint32_t waitTime){
     //Measures the SBIT rate seen by OHv3 ohN for the non-masked VFAT found in maskOh as a function of scanReg
     //It is assumed that all other VFATs are masked in the OHv3 via maskOh
     //Will scan from dacMin to dacMax in steps of dacStep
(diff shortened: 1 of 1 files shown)


bdorney:
I had a lot of trouble trying to extract `vfatN` from `maskOh` last night.  Right now the current implementation is only using `maskOh` to make sure that `vfatN` is in the list of sync'd VFATs but it is not being used to determine `vfatN`.  

I was able to play with some hex logic to make from `maskOh` a hex that is all 0's except for the bit `vfatN` but it was unclear how to extract from this number an integer that could be used as `vfatN`.

e.g. if maskOh has every VFAT except for VFAT1 masked then:

```
~maskOh & 0xFFFFFF
```

Should give in binary: `0b000000000000000000000010`

But I didn't know how to manipulate this to get an integer that was `vfatN=1` without supplying it. The implementation could be simplified if I could get some advice for this.

Thanks.

src/calibration_routines.cpp
```diff
@@ -475,6 +475,124 @@ void genScan(const RPCMsg *request, RPCMsg *response)
...
+    char regBuf[200];
+    int iFWVersion = readReg(la->rtxn, la->dbi, "GEM_AMC.GEM_SYSTEM.RELEASE.MAJOR");
+    if (iFWVersion < 3){ //v3 electronics behavior
+        sprintf(regBuf,"SBIT Rate Scan is Presently only supported in V3 Electronics");
+        la->response->set_string("error",regBuf);
+        return;
+    }
+
+    uint32_t goodVFATs = vfatSyncCheckLocal(la, ohN);
+    if( !( (goodVFATs >> vfatN) & 0x1 ) ){
```
mexanick: shouldn't the maskOh be used here? Like
`if ( !(goodVFATs==maskOh)) {return error;}`
?

src/calibration_routines.cpp
```diff
@@ -475,6 +475,124 @@ void genScan(const RPCMsg *request, RPCMsg *response)
...
+        sprintf(regBuf,"The requested VFAT is not synced; goodVFATs: %x\t requested VFAT: %x", goodVFATs, maskOh);
+        la->response->set_string("error",regBuf);
+        return;
+    }
+
+    if( vfatN > 23 ){
```
mexanick: This check has to be done on the host PC and no inexistent VFAT number should be passed

src/calibration_routines.cpp
```diff
@@ -475,6 +475,124 @@ void genScan(const RPCMsg *request, RPCMsg *response)
...
+        sprintf(regBuf,"I did not find a VFAT to perform this scan on, maskOh: %x", maskOh);
+        la->response->set_string("error",regBuf);
+        return;
+    }
+
+    //If ch!=128 store the original channel mask settings
+    //Then mask all other channels except for channel ch
+    std::map<uint32_t, uint32_t> map_chanOrigMask; //key -> reg addr; val -> reg value
+    if( ch != 128){
```
mexanick, line 543: what should happen if the ch==128? IIRC this means in our convention to use all channels, but such case is invalid in the context of this function and should be checked against at the host PC.

src/calibration_routines.cpp
```diff
@@ -475,6 +475,124 @@ void genScan(const RPCMsg *request, RPCMsg *response)
...
+        uint32_t chanMaskAddr[128];
+        for(unsigned int chan=0; chan<128; ++chan){ //Loop Over All Channels
+            uint32_t chMask = 1;
+            if ( ch == chan){ //Do not mask the channel of interest
+                chMask = 0;
+            }
+
+            //store the original channel mask
```
mexanick, line 551: do we really need this? In our operation we usually reconfigure all the VFATs settings including channel registers every time we perform a new operation. My suggestion would be following then:
the default system has all the chan registers masked (or unmasked), depending what is set after the powercycle, then we know a priori what should be the mask to write to perform a scan and to what mask we should return the system.

src/calibration_routines.cpp
```diff
@@ -475,6 +475,124 @@ void genScan(const RPCMsg *request, RPCMsg *response)
...
+    //Take the VFATs out of slow control only mode
+    writeReg(la->rtxn, la->dbi, "GEM_AMC.GEM_SYSTEM.VFAT3.VFAT3_RUN_MODE", 0x1, la->response);
+
+    //Place chip in run mode
+    //sprintf(regBuf,"GEM_AMC.OH.OH%i.GEB.VFAT%i.CFG_RUN",ohN,vfatN);
+    //uint32_t runAddr = getAddress(la->rtxn, la->dbi, regBuf, la->response);
+    //writeRawAddress(runAddr, 0x1, la->response);
+
+    //Loop from dacMin to dacMax in steps of dacStep
+    for(uint32_t dacVal = dacMin; dacVal <= dacMax; dacVal += dacStep){
```
mexanick, line 584: the dacMin, dacMax and dacStep parameters has to be checked before the call to ensure that they have valid values and the resulting array size is integer (or certain flooring rule has to be applied)

src/calibration_routines.cpp
```diff
@@ -475,6 +475,124 @@ void genScan(const RPCMsg *request, RPCMsg *response)
...
+
+        int idx = (dacVal-dacMin)/dacStep;
+        outDataDacVal[idx] = dacVal;
+        outDataTrigRate[idx] = readRawAddress(ohTrigRateAddr, la->response);
+    } //End Loop from dacMin to dacMax
+
+    //Take chip out of run mode
+    //writeRawAddress(runAddr, 0x0, la->response);
+
+    //Restore the original channel masks if specific channel was requested
```
mexanick, line 598: see the comment above about possible simplification

src/calibration_routines.cpp
```diff
@@ -475,6 +475,124 @@ void genScan(const RPCMsg *request, RPCMsg *response)
...
+
+void sbitRateScan(const RPCMsg *request, RPCMsg *response){
+    auto env = lmdb::env::create();
+    env.set_mapsize(1UL * 1024UL * 1024UL * 40UL); /* 40 MiB */
+    env.open("/mnt/persistent/texas/address_table.mdb", 0, 0664);
+    auto rtxn = lmdb::txn::begin(env, nullptr, MDB_RDONLY);
+    auto dbi = lmdb::dbi::open(rtxn, nullptr);
+
+    uint32_t ohN = request->get_word("ohN");
+    uint32_t maskOh = request->get_word("maskOh");
```
mexanick, line 619: I didn't find where it is used...

src/calibration_routines.cpp
```diff
@@ -475,6 +475,124 @@ void genScan(const RPCMsg *request, RPCMsg *response)
...
+    uint32_t vfatN = request->get_word("vfatN");
```
mexanick: if you want to get the vfatN from the maskOh, you can hardcode an array with VFAT numbers, i.e. [0b001, 0b010, 0b100,... etc] and use the std::find() function which returns an index, which will correspond to the VFAT number

src/calibration_routines.cpp
bdorney: No, `maskOh` will have all but *one* VFAT masked, e.g. for `VFAT1` we have: `0b111111111111111111111101`.

While goodVFATs will have a 1 in each position that a VFAT is sync'd, e.g. the current case we have 0xffffcf (`0b111111111111111111001111`).

The original idea was to extract `vfatN` from `maskOh` however I couldn't think of a good implementation.  However, with your suggestion below I can improve thanks!

Right now these limitations are imposed by the firmware but I've asked Andrew & Evaldas if this can be migrated from the CTP7 to the OHv3 FPGA and then parallelized such that the OH can monitor 24 VFATs simultaneously.

src/calibration_routines.cpp
bdorney: With your suggestion below the check will be on the output of `std::find(...)`.  If the output is the `stlcontainer::end()` then it will terminate and log an error message in the RPC response e.g. unrecognized parameter for `maskOh` input.

src/calibration_routines.cpp
bdorney: The case of `ch==128` is 100% valid in the context of this function, in fact that's how the above screenshot was obtained.  You look at the rate the OH sees from the entire VFAT.

This function can then two different types of scans, a coarse over an entire VFAT scan *or* a fine scan looking at the rate from a single channel.

A check can be implemented if `ch >= 129`.  This would be invalid.

This function is useful beyond just setting "CFG_THR_ARM_DAC" for the context of configuring the chip.  It can also be used to measure particle rate as I have done for a few measurements already.

src/calibration_routines.cpp
bdorney: If you look at the `ultra*.py` scripts in most case they return the system to the configuration before the scan started.  This includes placing chips back into sleep mode, reseting the original trigger source configuration, etc...

I think it's dangerous to drop this.  For one, the channel registers right now are not configured by default (one has to supply `--chConfig` argument to the configure script).  Additionally you may not want to change the channel configuration each time you configure e.g. you just want to update the comparator mode, either `CFD` or `ARM`, or you want to write a different set of comparator DAC values (`CFG_THR_ARM_DAC`).  

If you remove this block of code and make a channel scan afterward the state of the system will have 127 channels masked and this will *not* by default be restored.

Additionally presently since we lack an RPC method for setting channel registers this would induce some time loss for re-configuring but that's another story.

**TL;DR:** it is dangerous to remove this block since channel registers are not configured by default and the system may be left in an "unuseable state" after this method is called for a given channel.  Strongly urge to keep.

src/calibration_routines.cpp
bdorney: Sounds like a good idea.  I will add some error statements inserted into the RPC response.

src/calibration_routines.cpp
bdorney: Yeah this has to do with my comment about not figuring out how to extract `vfatN` from `maskOh` but still needing a way to check that the VFAT was sync'd (e.g. bit at vfatN=1  in `goodVFATs`).  With your suggestion below I can improve this a bit so that only `maskOh` will be used and `vfatN` will be determined only from that.  Thanks!

src/calibration_routines.cpp
mexanick: I would make these checks at the host PC, so we don't spend precious computation time at the Zynq

src/calibration_routines.cpp
bdorney: Yeah I'll try to add something like that, thanks!

src/calibration_routines.cpp
bdorney: Okay, I'll try to have as many of the checks in the `xhal` method then.

src/calibration_routines.cpp
mexanick: I understand your reasons. Let's keep it as is right now, but discuss later with @evka85, @jsturdy and others. What I propose is a kind of different approach which might require some changes in the other routines, however will make them work "out-of-the-box" for the freshly powered up system.

src/calibration_routines.cpp
```diff
@@ -492,15 +493,44 @@ void sbitRateScanLocal(localArgs *la, uint32_t *outDataDacVal, uint32_t *outData
         return;
     }
 
-    uint32_t goodVFATs = vfatSyncCheckLocal(la, ohN);
-    if( !( (goodVFATs >> vfatN) & 0x1 ) ){
-        sprintf(regBuf,"The requested VFAT is not synced; goodVFATs: %x\t requested VFAT: %x", goodVFATs, maskOh);
+    //Hard code possible maskOh values and how they map to vfatN
+    std::map<uint32_t,uint32_t> map_maskOh2vfatN;
```
mexanick: std::unordered_map will be better performing in this case bc it creates a hash table, which is better for a lookup

bdorney: done

mexanick (approved):

PR #16, opened 2018-01-24
"Bug Fix" New VFAT3 Slow Control Only Mode Behavior

bdorney:
## Description

@evka renamed the register for placing the VFAT3s into slow control only mode and inverted it's meaning in CTP7 FW version v3.3.6 and beyond.  This PR ensures this new behavior is propagated to the SW.

For complete discussion see: http://cmsonline.cern.ch/cms-elog/1026503

### Types of changes

- [X] Bug fix (non-breaking change which fixes an issue)

## Motivation and Context

Presently if using CTP7 FW v3.3.6 or higher `genScanLocal(...)` will either:

1. throw a reg not found error, or
2. place VFAT3s into slow control only mode (no data taking possible).

This PR addresses this issue.

## How Has This Been Tested?

See above e-log.  Change was trivial, but tested anyway.

## Checklist:

### DIFF
--- a/src/calibration_routines.cpp
+++ b/src/calibration_routines.cpp
@@ -285,9 +285,6 @@ void genScanLocal(localArgs *la, uint32_t *outData, uint32_t ohN, uint32_t mask,
         //Configure VFAT_DAQ_MONITOR
         dacMonConfLocal(la, ohN, ch);
 
-        //Place VFATs out of slow control only mode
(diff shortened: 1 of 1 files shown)


mexanick (commented):
Should this call be just removed or replaced with another one?

bdorney:
> Should this call be just removed or replaced with another one?

@evka suggested removed.  The default value for the new register ensures they will not be in slow control only mode.  However, we could instead of removing the line change it to the new node and invert the value so that the "slow control only mode" is again explicitly set to false.

mexanick (approved):

Issue #7, opened 2017-10-06
Development flow implementation


mexanick:
## Brief summary of issue

Please make PRs to the develop branch of the repository. Master branch should remain stable at all times. When a sufficient confidence in the code correctness is received, the develop branch will be merged into master. The only reason to make a direct PRs to the master is hotfixes for bugs.

jsturdy:
```
sudo yum install git-flow
cd <repo>
git flow init
# profit
```

bdorney:
Further reading: https://danielkummer.github.io/git-flow-cheatsheet/

Branching model base: http://nvie.com/posts/a-successful-git-branching-model/

PR #17, opened 2018-02-12
New Feature: Parallel SBIT Rate Scan vs. Arbitrary Register

bdorney:
## Description

In Optohybrid v3 FW version [3.0.7.A](https://github.com/cms-gem-daq-project/OptoHybridv3/releases/tag/3.0.7.A) a set of new counters was added which measure the rate of sbits being sent from each VFAT individually.  I have added a new SBIT rate scan module `sbitRateScanParallelLocal(...)` which takes advantage of this new FW functionality.

However, due to the issues with Optohybrid FW versions v3.0.7.A (trigger link underflow error) and v3.0.8.A (VFATs kept in hard reset) it has not yet been possible to test yet.

### Types of changes

## Motivation and Context

In the previous sbit rate scan `sbitRateScanLocal(...)` it was only possible to scan one VFAT at a time which caused the scan to take an exceptionally long time (especially in the case of a channel scan).  I had added `sbitRateScanParallelLocal(...)` which uses the new FW functionality (OHv3 version 3.0.7.A or higher) to scan all VFATs in parallel.

### Changes to sbitRateScan(const RPCMsg *request, RPCMsg *response)

The previous functionality, described in https://github.com/cms-gem-daq-project/ctp7_modules/pull/15, has been preserved.  New functionality has been added:

| Argument | Type | Description |
| ---------- | ----- | ----------- |
| `maskOh` | `uint32_t` | If the the `isParallel` flag is true this acts as the standard `vfatmask` otherwise see https://github.com/cms-gem-daq-project/ctp7_modules/pull/15. |
| `isParallel` | `bool` | If true `sbitRateScan(...)` will call `sbitRateScanParallelLocal(...)` instead of the series version `sbitRateScanLocal(...)`. |
| `data_trigRatePerVFAT` | `uint32_t *` | Pointer to an array that holds the trigger rate values for each VFAT for each scan point of `scanReg`. Note the size of the array is assumed to be `24*(dacMax - dacMin + 1)/dacStep`.  The `idx` for `vfatN` is `N*(dacVal-dacMin+1)/dacStep` and the `dacVal` that corresponds to this trigger rate is then `data_dacVal[idx]`. |

Please note `invertVFATPos` is not used in `sbitRateScanParallelLocal(...)` since from OH FW version [3.0.5.A](https://github.com/cms-gem-daq-project/OptoHybridv3/releases/tag/3.0.5.A) the VFAT numbering in the OH matches the "software" convention;  However since OH FW version 3.0.4.A is still the last "stable" FW release I have kept this flag.

Also the `waitTime` is not used in `sbitRateScanParallelLocal(...)` for simplicity.  The counters will increment for a total of 1 second (40,079,000 clock cycles), thus the rate is simply the count.

If `isParallel` is true then `sbitRateScanParallelLocal(...)` will be called and an additional word array will be retuned in the RPC response `data_trigRatePerVFAT`.  If `isParallel` is false then the current functionality will persist.

###  Description of sbitRateScanParallelLocal(...)

| Argument | Type | Description |
| ---------- | ----- | ----------- |
| `la` | `localArgs` | Struct containing the `rtxn`, `dbi` and RPC `response` objects. |
| `outDataDacVal` | `uint32_t *` | Pointer to an array that will hold the DAC values of each scan point of `scanReg`.  Note the size of the array is assumed to be `(dacMax - daxMin + 1) / dacStep`.  The `idx` position of this array represents the same scan point as the `outDataTrigRateOverall` argument. |
| `outDataTrigRatePerVFAT` | `uint32_t *` | Pointer to an array that holds the trigger rate values for each VFAT for each scan point of `scanReg`. Note the size of the array is assumed to be `24*(dacMax - dacMin + 1)/dacStep`.  The `idx` for `vfatN` is `N*(dacVal-dacMin+1)/dacStep` and the `dacVal` that corresponds to this trigger rate is then `outDataDacVal[idx]`. |
| `outDataTrigRateOverall` | `uint32_t *` | As `outDataDacVal` but for the y-values of the array. |
| `ohN` | `uint32_t` | Link on CTP7 to be scanned. | 
| `vfatmask` | `uint32_t` | VFAT mask to be used, e.g. `0x30` means all VFATs except for 4 & 5 will be scanned. |
| `ch` | `uint32_t` | VFAT Channel to be used, all other channels will be masked.  The original mask settings will be returned after the scan has completed.  Expected to be in the range [0,127].  Note if `ch=128` then the channel mask settings will not be altered and the rate measurement will be an `OR` of all channels. |
| `dacMin` | `uint32_t` | Starting value of `scanReg`.|
| `dacMax` | `uint32_t` | Final value of `scanReg`.|
| `dacStep` | `uint32_t` | Step size used to move from `dacMin` to `dacMax` of `scanReg`.|
| `scanReg` | `std::string` | Register to be scanned, the leading `CFG_` of the register name is omitted when passing to this argument. |

## How Has This Been Tested?

Due to a FW bug it hasn't yet.  When @andrewpeck releases the new OHv3a FW version addressing  [OptohybridV3 Issue #7](https://github.com/cms-gem-daq-project/OptoHybridv3/issues/7) testing will resume.

sbitratescanparallel

## Checklist:

### DIFF
--- a/src/calibration_routines.cpp
+++ b/src/calibration_routines.cpp
@@ -294,9 +294,7 @@ void genScanLocal(localArgs *la, uint32_t *outData, uint32_t ohN, uint32_t mask,
             //Write the scan reg value
             for(int vfatN = 0; vfatN < 24; vfatN++) if((notmask >> vfatN) & 0x1)
             {
-                //writeRawAddress(scanDacAddr[vfatN], dacVal, la->response);
                 writeReg(la->rtxn, la->dbi, stdsprintf("GEM_AMC.OH.OH%i.GEB.VFAT%i.CFG_%s",ohN,vfatN,scanReg.c_str()), dacVal, la->response);
-                //std::this_thread::sleep_for(std::chrono::microseconds(50)); //Seen some evidence that
             }
 
             //Reset and enable the VFAT_DAQ_MONITOR
@@ -311,7 +309,6 @@ void genScanLocal(localArgs *la, uint32_t *outData, uint32_t ohN, uint32_t mask,
                 uint32_t l1aCnt = 0;
                 while(l1aCnt < nevts){
                     l1aCnt = readRawAddress(l1CntAddr, la->response);
-                    //std::this_thread::sleep_for(std::chrono::microseconds(50));
                     std::this_thread::sleep_for(std::chrono::microseconds(200));
                 }
 
@@ -329,21 +326,21 @@ void genScanLocal(localArgs *la, uint32_t *outData, uint32_t ohN, uint32_t mask,
 
 
             for(int vfatN = 0; vfatN < 24; vfatN++){
+                if ( !( (notmask >> vfat) & 0x1)) continue;
+
                 int idx = vfatN*(dacMax-dacMin+1)/dacStep+(dacVal-dacMin)/dacStep;
                 outData[idx] = readRawAddress(daqMonAddr[vfatN], la->response);
 
-                if((notmask >> vfatN) & 0x1){
-                    LOGGER->log_message(LogManager::DEBUG, stdsprintf("%s Value: %i; Readback Val: %i; Nhits: %i; Nev: %i; CFG_THR_ARM: %i",
-                                 scanReg.c_str(),
-                                 dacVal,
-                                 readReg(la->rtxn, la->dbi, stdsprintf("GEM_AMC.OH.OH%i.GEB.VFAT%i.CFG_%s",ohN,vfatN,scanReg.c_str())),
-                                 readReg(la->rtxn, la->dbi, stdsprintf("GEM_AMC.GEM_TESTS.VFAT_DAQ_MONITOR.VFAT%i.CHANNEL_FIRE_COUNT",vfatN)),
-                                 readReg(la->rtxn, la->dbi, stdsprintf("GEM_AMC.GEM_TESTS.VFAT_DAQ_MONITOR.VFAT%i.GOOD_EVENTS_COUNT",vfatN)),
-                                 readReg(la->rtxn, la->dbi, stdsprintf("GEM_AMC.OH.OH%i.GEB.VFAT%i.CFG_THR_ARM_DAC",ohN,vfatN,scanReg.c_str()))
-                        )
-                    );
-                }
-            }
+                LOGGER->log_message(LogManager::DEBUG, stdsprintf("%s Value: %i; Readback Val: %i; Nhits: %i; Nev: %i; CFG_THR_ARM: %i",
+                             scanReg.c_str(),
+                             dacVal,
+                             readReg(la->rtxn, la->dbi, stdsprintf("GEM_AMC.OH.OH%i.GEB.VFAT%i.CFG_%s",ohN,vfatN,scanReg.c_str())),
+                             readReg(la->rtxn, la->dbi, stdsprintf("GEM_AMC.GEM_TESTS.VFAT_DAQ_MONITOR.VFAT%i.CHANNEL_FIRE_COUNT",vfatN)),
+                             readReg(la->rtxn, la->dbi, stdsprintf("GEM_AMC.GEM_TESTS.VFAT_DAQ_MONITOR.VFAT%i.GOOD_EVENTS_COUNT",vfatN)),
+                             readReg(la->rtxn, la->dbi, stdsprintf("GEM_AMC.OH.OH%i.GEB.VFAT%i.CFG_THR_ARM_DAC",ohN,vfatN,scanReg.c_str()))
+                    )
+                );
+            } //End Loop over vfats
         } //End Loop from dacMin to dacMax
 
         //If the calpulse for channel ch was turned on, turn it off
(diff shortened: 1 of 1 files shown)


src/calibration_routines.cpp
```diff
@@ -605,6 +590,113 @@ void sbitRateScanLocal(localArgs *la, uint32_t *outDataDacVal, uint32_t *outData
...
+
+    char regBuf[200];
+
+    //Are we using v3 electronics?
+    int iFWVersion = readReg(la->rtxn, la->dbi, "GEM_AMC.GEM_SYSTEM.RELEASE.MAJOR");
+    if (iFWVersion < 3){ //v2b electronics behavior
+        sprintf(regBuf,"SBIT Rate Scan is Presently only supported in V3 Electronics");
+        la->response->set_string("error",regBuf);
+        return;
+    }
```
mexanick, line 610: Lines 604-610 are often used by various methods. Suggestion: add new method taking the caller's name as a parameter:
```
int fw_version_check(std:: string & caller_name, localArgs *la,)
{
    int iFWVersion = readReg(la->rtxn, la->dbi, "GEM_AMC.GEM_SYSTEM.RELEASE.MAJOR");
    if (iFWVersion < 3){ //v2b electronics behavior
        sprintf(regBuf,"%s is Presently only supported in V3 Electronics", caller_name.c_str());
        la->response->set_string("error",regBuf);
    }
    return iFWVersion;
}
```
use this method where appropriate.

src/calibration_routines.cpp
```diff
@@ -605,6 +590,113 @@ void sbitRateScanLocal(localArgs *la, uint32_t *outDataDacVal, uint32_t *outData
...
+                sprintf(regBuf, "GEM_AMC.OH.OH%i.GEB.VFAT%i.VFAT_CHANNELS.CHANNEL%i.MASK",ohN,vfat,chan);
+                chanMaskAddr=getAddress(la->rtxn, la->dbi, regBuf, la->response);
+                map_chanOrigMask[chanMaskAddr]=readReg(la->rtxn, la->dbi, regBuf);   //We'll write this by address later
+
+                //write the new channel mask
+                writeRawAddress(chanMaskAddr, chMask, la->response);
+            } //End Loop Over all Channels
+        } //End loop over all vfats
+    } //End Case: Measuring Rate for 1 channel
+
```
mexanick, line 646: Lines 621-646 are used in multiple places. Please move them to a separate method, which should set new mask and return original

src/calibration_routines.cpp
```diff
@@ -605,6 +590,113 @@ void sbitRateScanLocal(localArgs *la, uint32_t *outDataDacVal, uint32_t *outData
...
+            idx = vfat*(dacMax-dacMin+1)/dacStep+(dacVal-dacMin)/dacStep;
+            outDataTrigRatePerVFAT[idx] = readRawAddress(ohTrigRateAddr[vfat], la->response);
+        }
+    } //End Loop from dacMin to dacMax
+
+    //Restore the original channel masks if specific channel was requested
+    if( ch != 128){
+        for(auto chanPtr = map_chanOrigMask.begin(); chanPtr != map_chanOrigMask.end(); ++chanPtr){
+            writeRawAddress( (*chanPtr).first, (*chanPtr).second, la->response);
+        }
```
mexanick, line 694: lines 692-694 should be moved the function `setChannelMask` and called as needed

mexanick (changes requested):
Please move the commented pieces of the code to the separate methods to reduce code duplication

src/calibration_routines.cpp
bdorney: Here the `caller_name` string is the function name that is calling `fw_version_check` right?

Additionally is there a way for it to determine `caller_name` automatically (e.g. `this->...`?).  Or should this be hard coded?

src/calibration_routines.cpp
bdorney: There is no such function yet...while there is some code here that is "duplicated" it is not done completely.  For example in the series case there is no loop over all VFATs.

I think adding this function is beyond the scope of this PR and deserves more thought/it's own PR.

src/calibration_routines.cpp
bdorney: see comment below.  Argue this is outside the scope of this PR and it deserves it's own issue/PR/

src/calibration_routines.cpp
bdorney: Also since this affects multiple functions and not just `sbitRateScanParallelLocal(...)` this should warrant it's own PR and is beyond the scope here.

mexanick: My comment will apply to the cases below as well. I think it is a good idea to implement my suggestions now *and* make another issue about refactoring rest of the code. Just not make things even more duplicated. However if you insist on merging as is - we can do it. I let you decide. 
P.S. Yes, `caller_name` is the name of the function. It is not possible to pass it as `this`. In principle one can implement a so-called lazy callable, but it would be an over-complication here.

bdorney:
Baseline testing completed: http://cmsonline.cern.ch/cms-elog/1028798

Will need to consider adding requested functionality.

mexanick (approved):
Ok, I approve the PR since this is an important feature. However careful refactoring should be done in a very near future

PR #20, opened 2018-02-13
New Feature: genScanLocal uses current pulse mode of calibration module

bdorney:
## Description

VFAT3 design team has recommended to use the current pulse mode for injecting cal pulses into the system with the calibration module as it allows a finer granularity scurve than the voltage injection mode.

This PR adds functionality to `genScanLocal(...)` for toggling between the voltage and current pulse injection mode and for setting the current pulse scale factor to be used (`CFG_CAL_FS`).

The current pulse duration is hard coded to 1 clock cycle at the recommendation of the VFAT3 design team (additionally injecting for more than one clock cycle is probably equivalent to just changing the `CFG_PULSE_STRETCH` register).

### Types of changes

Two new arguments have been added:
- `bool currentPulse` if set to `true` (`false`) uses the current (voltage) pulse injection, and 
- `uint32_t calScaleFactor` the value to be written to the `CFG_CAL_FS` register, only applicable if `currentPulse == true`.

## Motivation and Context

Finer granularity scurves are possible.

## How Has This Been Tested?

See following e-logs: 
- http://cmsonline.cern.ch/cms-elog/1028418
- http://cmsonline.cern.ch/cms-elog/1028540

## Checklist:

### DIFF
--- a/src/calibration_routines.cpp
+++ b/src/calibration_routines.cpp
@@ -467,6 +467,7 @@ void genScan(const RPCMsg *request, RPCMsg *response)
     uint32_t dacMax = request->get_word("dacMax");
     uint32_t dacStep = request->get_word("dacStep");
     bool useCalPulse = request->get_word("useCalPulse");
+    bool currentPulse = request->get_word("currentPulse");
     std::string scanReg = request->get_string("scanReg");
 
     bool useUltra = false;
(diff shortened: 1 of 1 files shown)


src/calibration_routines.cpp
```diff
@@ -243,22 +243,38 @@ void genScanLocal(localArgs *la, uint32_t *outData, uint32_t ohN, uint32_t mask,
...
                 la->response->set_string("error","It doesn't make sense to calpulse all channels");
                 return;
-            }
-            else{
-                for(int vfatN = 0; vfatN < 24; vfatN++){
-                    if((notmask >> vfatN) & 0x1){
+            } //End Case: OR of all channels
+            else{ //Case: Pulse a specific channel
+                for(int vfatN = 0; vfatN < 24; vfatN++){ //Loop over all VFATs
+                    if((notmask >> vfatN) & 0x1){ //End VFAT is not masked
```
mexanick, line 260: A comment should be just `VFAT is not masked`

bdorney:
Quick update, unsure of the behavior of `CFG_CAL_DUR`, i.e. if `0x0` corresponds to an output pulse for 0 or 1 clock cycles.  Awaiting feedback from VFAT3 design team.

mexanick (approved):
As the code interface is proven to be working, we are merging this PR towards the legacy release of the xHAL. Further work on adjusting calibration routine parameters may be provided in other issues/PR

Issue #35, opened 2018-04-18
Bug Report: missing break statements when using switch statements.


bdorney:
## Brief summary of issue

In `ttcGenConf(...)` the original `if-else if-else` statement was replaced by a `switch` statement:

https://github.com/cms-gem-daq-project/ctp7_modules/blob/c2387633c1dd1bbbdcdeee03370e3d78f89ca192/src/calibration_routines.cpp#L212-L275

But this misses `break` statements.  As a result the `switch` statement will evaluate **all** of the statements after the first matching statement on the input variable.   This generates an RPC error if working with the v3 electronics since there is no `T1.Controller` block in the FW.

This additionally crops up in:
- `genScanLocal(...)`

### Types of issue

- [X] Bug report (report an issue with the code)
- [ ] Feature request (request for change which adds functionality)

## Expected Behavior

When using `switch` statements the `break` statement must be included at the end of the block for a particular case.

## Current Behavior

Right now all three cases statements are being executed if `fw_version_check(...)` returns `0x3`:

```
Apr 18 12:28:05 CTP7 local0.notice rpcsvc[10251]: rpcsvc: Client connected from 192.168.0.180
Apr 18 12:28:06 CTP7 local0.err rpcsvc[10251]: calibration_routines.ttcGenConf: Key: GEM_AMC.OH.OH0.T1Controller.RESET is NOT found
Apr 18 12:28:06 CTP7 local0.err rpcsvc[10251]: calibration_routines.ttcGenConf: Key: GEM_AMC.OH.OH0.T1Controller.MODE is NOT found
Apr 18 12:28:06 CTP7 local0.err rpcsvc[10251]: calibration_routines.ttcGenConf: Key: GEM_AMC.OH.OH0.T1Controller.MODE is NOT found
Apr 18 12:28:06 CTP7 local0.err rpcsvc[10251]: calibration_routines.ttcGenConf: Key: GEM_AMC.OH.OH0.T1Controller.DELAY is NOT found
Apr 18 12:28:06 CTP7 local0.err rpcsvc[10251]: calibration_routines.ttcGenConf: Key: GEM_AMC.OH.OH0.T1Controller.DELAY is NOT found
Apr 18 12:28:06 CTP7 local0.err rpcsvc[10251]: calibration_routines.ttcGenConf: Key: GEM_AMC.OH.OH0.T1Controller.INTERVAL is NOT found
[...]

Issue #34, opened 2018-04-18
Bug Report: fw_version_check() sets error keyword un-necessarily


bdorney:
## Brief summary of issue

The function `fw_version_check(...)` is setting the `error` word when it should not:

https://github.com/cms-gem-daq-project/ctp7_modules/blob/c2387633c1dd1bbbdcdeee03370e3d78f89ca192/src/calibration_routines.cpp#L19-L28

Here `iFWVersion` could be `0x1` without causing an error, e.g. see `ttcGenConf(...)`:

https://github.com/cms-gem-daq-project/ctp7_modules/blob/c2387633c1dd1bbbdcdeee03370e3d78f89ca192/src/calibration_routines.cpp#L177-L209 

This function has both v2b and v3 electronics behavior coded.  For example if `iFWVersion` is 3 then `ttcGenConf(...)` will configure the `TTC.GENERATOR` block in the `ctp7` firmware.  But if `iFWVersion` is 1 then `ttcGenConf(...)` will configure the `T1.Controller` block in the `OHv2b` firmware. 

### Types of issue

- [X] Bug report (report an issue with the code)
- [ ] Feature request (request for change which adds functionality)

## Expected Behavior

`fw_version_check(...)` should only raise an error if it fails to read the major fw release register of the card:

https://github.com/cms-gem-daq-project/ctp7_modules/blob/c2387633c1dd1bbbdcdeee03370e3d78f89ca192/src/calibration_routines.cpp#L21

Or if the value returned is `0x0` (indicative of a CTP7 timeout error).

## Current Behavior

`fw_version_check(...)` raises an error if major FW release is not equal to `0x3`.

### Steps to Reproduce (for bugs)

1. Run `fw_version_check(...)` on a `ctp7` with FW `v1.X.Y`

## Possible Solution (for bugs)

Rely on `readReg(...)` to set the `error` word in the case of a failure to read the register pointed to by the node.  Then change:

https://github.com/cms-gem-daq-project/ctp7_modules/blob/c2387633c1dd1bbbdcdeee03370e3d78f89ca192/src/calibration_routines.cpp#L23-L25

to:

```
if (iFWVersion == 0){ //CTP7 Timeout Error
        sprintf(regBuf,"Major FW Release is 0x0; CTP7 Timeout Error");
        la->response->set_string("error",regBuf);
    }
``` 

In this case the `caller_name` argument could be dropped.

## Context (for feature requests)

[...]

PR #36, opened 2018-04-18
Addressing issues #34, #35 and https://github.com/cms-gem-daq-project/xhal/pull/58 

mexanick:
Addressing issues #34, #35 and adding the LINKX_UNDERFLOW counter to TRIGGEROHMain

## Description

Fixing the `switch` statements and adding missing monitoring registers. Documentation is updated thanks to @evka85 
### Types of changes

## Checklist:

### DIFF
--- a/src/calibration_routines.cpp
+++ b/src/calibration_routines.cpp
@@ -12,7 +12,7 @@
 
 /*! \fn unsigned int fw_version_check(const char* caller_name, localArgs *la)
  *  \brief Returns AMC FW version
- *  in case FW version is below 3.X sets an error string in response
+ *  in case FW version is not 1.X or 3.X sets an error string in response
  *  \param caller_name Name of methods which called the FW version check
  *  \param la Local arguments structure
  */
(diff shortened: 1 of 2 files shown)


bdorney:
I'd like this to be merged after #37 to prevent repeat testing (again).  Additionally there are several PR's in the other repo's that #37 depends on.

src/calibration_routines.cpp
```diff
@@ -659,87 +681,98 @@ void genScan(const RPCMsg *request, RPCMsg *response)
...
+            uint32_t vfatN = (invertVFATPos) ? 23 - (*vfatNptr).second : (*vfatNptr).second;
 
-    uint32_t goodVFATs = vfatSyncCheckLocal(la, ohN);
-    if( !( (goodVFATs >> vfatN ) & 0x1 ) ){
-        sprintf(regBuf,"The requested VFAT is not synced; goodVFATs: %x\t requested VFAT: %i; maskOh: %x", goodVFATs, vfatN, maskOh);
-        la->response->set_string("error",regBuf);
-        return;
-    }
+            uint32_t goodVFATs = vfatSyncCheckLocal(la, ohN);
+            if( !( (goodVFATs >> vfatN ) & 0x1 ) ){
```
bdorney, line 724: I think your editor is inserting `\t` which can cause formatting issues.  Is it possible you update your `~/.vimrc` file or similar to replace `\t` with four `\s` characters?

bdorney (changes requested):
Aesthetic change requested to ensure no problems with different editors are introduced.

src/calibration_routines.cpp
mexanick: @bdorney I believe the new shift here is correct as I have changed a an `if then else` statement to a `switch` statement in this function which requires another indentation shift. See line 684. I have checked the file and it appears to have all the tabs converted to spaces, e.g. there's no \t but four \s everywhere

bdorney (approved):

PR #40, opened 2018-06-21
Packaging added 

mexanick:
## Description

ARM packaging is added and working. Temporary requires xhal to be under $BUILD_HOME. Will be updated later when wget-able libs become available
### Types of changes

## Checklist:


conf/conf.txt
```diff
@@ -1,49 +0,0 @@
-dacName/C:dacVal/I
```
bdorney: this file serves as an example for `configureVFAT3sLocal(...)`.

If this is removed it should be reloacted elsewhere until such time that `configureVFAT3sLocal(...)` no longer references this file.

bdorney (changes requested):
`conf/conf.txt` should not be removed.

bdorney:
@mexanick is this PR stale?

conf/conf.txt
bdorney: Additionally this file is now referenced by `setup_ctp7.sh`

mexanick:
Nope, but requires an update. Will be resolved in following days

conf/conf.txt
mexanick: why it is here then? It should rather be in that repository...

mexanick:
ready for review

conf/conf.txt
bdorney: Still against removal of this file.  Until `configureVFAT3sLocal(...)` is updated to no longer use it this file *should* stay in this repository.

Makefile
```diff
@@ -2,38 +2,59 @@ ifndef PETA_STAGE
...
+LongPackage  := ctp7_modules
+PackageName  := $(ShortPackage)
+PackagePath  := $(shell pwd)
+PackageDir   := pkg/$(ShortPackage)
+Arch         := arm
+Packager     := Mykhailo Dalchenko
+
+CTP7_MODULES_VER_MAJOR=1
+CTP7_MODULES_VER_MINOR=0
+CTP7_MODULES_VER_PATCH=0
```
jsturdy: could use the tag-based version detection as in `vfatqc`/`gempython`/`gemplotting`, e.g., 

https://github.com/cms-gem-daq-project/vfatqc-python-scripts/blob/19ed505e33ff483b0f570d041b6fdbfea50d5d1a/Makefile#L22-L24

jsturdy (changes requested):
Consider taking the version from a tag, rather than relying on hard-coding it into the `Makefile`

conf/conf.txt
mexanick: In my opinion, this file should go to the ctp7user repo

mexanick:
Makefile updated to provide tag-based versioning

conf/conf.txt
bdorney: It can be removed, but after `configureVFAT3sLocal(...)` is updated to no longer use these files.  I think this is a suitable compromise (because then it wouldn't even need to be placed in the `gemctp7user` repo.

Otherwise you're going to add a file there, edit the setup script, remove a file here, sometime later make a change here, then remove a file there, edit the setup script again.  It's confusing.

Additionally in the future new developers might join and miss where this file is and what it's supposed to be (e.g. after the three of us leave the project).

Keeping it here is the option that requires the least work and the least complexity.

jsturdy (commented):
can you verify that it's working as expected?

mexanick:
it produces `ctp7_modules-1.0.1-centos7.arm.rpm`

jsturdy (dismissed):

conf/conf.txt
jsturdy: my 2 cents, no one should seriously care about this file.  it has always been a temporary solution, and the *only* long term solution we should care about would be the removal of dependence (proper VFAT configuration, assuming i'm understanding the usage of this file correctly).
to wit, i agree with @bdorney re the removal in this instance, as the long term solution is not yet implemented

include/daq_monitor.h
```diff
@@ -10,6 +10,8 @@
 #include "utils.h"
 #include <unistd.h>
 
+const int NOH_MAX = 12;
```
bdorney, line 13: I'm opposed to hard coding this.  This should be read from the FW or the LMDB.  In the future other upgrades will feature a different number of optohybrids on the backend AMC (either uTCA or aTCA based).  If we are forward thinking and rpcmodules are used there then many of the present modules we have *could* work out of the box if we did not hard code things like `NOH` or `NVFAT`.

mexanick: this is `NOH_MAX` - a maximal supported number of optohybrids by one CTP7. It is 12 and will be 12 at all times. Please see how it is used. The number of supported links for which firmware was compiled is indeed read out of the FW and used accordingly.

bdorney: This is exactly my point, it may *not* be always 12 in the future.

If you look at the design of the OH for GE2/1 it uses about half the fibers than the current one (and reduced number of GBTx's per OH).  So *more* OH's could run off a CTP7 (or equivalent backend card) for the GE2/1 wheel.  Hardcoding to use 12 OH's or 24 VFATs (when the VFAT number for GE2/1 hardware will also be different) will cause us issues in the future.

We can be forward thinking, for once... :D

mexanick: I'm sorry, but right now this looks like "I just don't want this PR to be merged. Ever. "

bdorney: I'm sorry you have that impression.  It was not my intent.  But I feel I've raised valid concerns here.  If you don't want to address this here then please open an issue on the issue page for this, similar to https://github.com/cms-gem-daq-project/ctp7_modules/issues/55.  But we *should* try to be forward thinking to prevent requiring to go back and change things in the future (something I'm very much guilty of myself).

mexanick: You know very well, that in the current version of the firmware there's no register to say the max NOH. And we have discussed multiple times and agreed that the monitoring should expect 12 OH per AMC. You can open an issue and bring it again to discussion if you want, but this is targeting the GE1/1 system, not the GE2/1.

bdorney: Most likely this was a private discussion because I do not recall having participated.  

Can you describe the reason this implementation is necessary?  From the changes below why is max NOH needed?  What issue are you trying to solve?

Why we may be developing "today" for GE1/1 we will (and by we I mean the "CMS GEM Project") have to support GE2/1 & ME0 in the future.  And to me this implementation seems limiting.  And it's unclear why.

mexanick: The reason is that we have agreed that the monitoring would always expect 12 OH linked to one CTP7 board. So the monitoring methods should return information for 12 links, as this is expected at the host PC side. If less links are expected to work, the excessive number of links will be masked, you can check the code to see how it is implemented, e.g. lines 65-78 of daq_monitoring.cpp

bdorney: I think I can propose a solution, couldn't we change to:

```C++
  if (request->get_key_exists("NOH")){
     unsigned int NOH_requested = request->get_word("NOH");
     if (NOH_requested > NOH) {
       LOGGER->log_message(LogManager::WARNING, stdsprintf("NOH requested (%i) > NUM_OF_OH AMC register (%i)",NOH_requested,NOH));
       ohMask = ohMask & (0xfffffffff >> (36-NOH));
     }
     NOH = NOH_requested;
   }
```

The "real" max a CTP7 could do is the number of RX/TX pairs that exist in the CXP's, so that's 3*12=36.  So that makes the maximum number of GBT or LpGBT's one CTP7/AMC can control as 36.  So can't we use `36 - NOH` and then a 36 bit mask `0xfffffffff`?  Doesn't this also work for the case of `N_Max=12`?

And for reference you can see the layout of the GE2/1 OH on [slide 2](https://indico.cern.ch/event/756570/contributions/3137759/attachments/1715714/2767835/OH_status_09_18_1.pdf).  And here indeed there is one VTTx/VTRx pair for slow control and 2 VTTx pairs for trigger data.  This would then give 36 OH's for a CTP7 or "CTP7-like" back-end card.

bdorney: tl;dr, my proposal is to change `NOH_MAX = 36` and update the bit mask to be `0xfffffffff` and it would work for both of us.

mexanick: I disagree. The design of the monitoring back-end assumes 12 OH links per AMC (for GE1/1, for GE2/1 we will need a complete redesign of monitoring. Even V2b and V3 are quite different from the monitoring point of view). Transferring "dead" data from additional 24 links which will never be used for GE1/1 is just a waste of time/bandwidth.

bdorney: > (for GE1/1, for GE2/1 we will need a complete redesign of monitoring. Even V2b and V3 are quite different from the monitoring point of view)

The FW will share common blocks in which case some monitoring functions will work out of the box, e.g. as they do now.

> Transferring "dead" data from additional 24 links which will never be used for GE1/1 is just a waste of time/bandwidth.

You will not be transferring "dead" data from additional 24 links.  You're still looping over `NOH`, not over the bits of `ohMask`; the higher bits of `ohMask` will not affect this.  The value of `NOH` still depends on either "user input" or the FW build, whichever is smaller.  I could provide a 32 bit value to `ohMask` key and if the FW was compiled for 12 links I'll only get info back (including dead) for at most 12 links.

I don't really see why this cannot be accommodated; it's a transparent change.

jsturdy: This type of information has already been hard-coded (and is placed in a more restrictive header file, which can be moved) for a long time, so it would be better to use the current definitions.

https://github.com/cms-gem-daq-project/cmsgemos/blob/9af04e7164036dc51b2ee4fec65bc4c4993bca70/gembase/include/gem/base/GEMFSMApplication.h#L31-L33

which could (should) be moved to `GEMApplication.h`.

Then, since the precedent is already set, keeping the `static const`s until such a time as an expansion of the design is warranted (these hard coded bits are in **one** location, for exactly this reason, such that they could become application parameters at some future point) would be the solution.

> tl;dr, my proposal is to change `NOH_MAX = 36` and update the bit mask to be `0xfffffffff` and it would work for both of us.

As is, this would **not** work, as the CTP7 `Zynq` is `ARMv7` which is a 32-bit CPU....

My proposal is that @mexanick makes the modification I mention above (moving the defs from `GEMFSMApplication.h` to `GEMFSMApplication.h`, and uses one of these three already existing definitions in his current implementation

mexanick: @jsturdy, how do you see implementation of your solution? This is an RPC module, not the cmsgemos code. Also, this code will benefit of refactoring and all the definitions such as static consts could be moved in one place (but I don't see how to share it with `GEMApplication.h`)

bdorney: > As is, this would not work, as the CTP7 Zynq is ARMv7 which is a 32-bit CPU....

Fair point, hah! How did it go again?

> Move along, nothing to see here...

bdorney: @mexanick @jsturdy we have the `config/` directory of several repo's as a git submodule.  Is it possible to have `cmsgemos/gembase` as a submodule in this (or other repo's)?

In this case wouldn't we just need to checkout the submodule, src the `cmsgemos` `$ENV` in addition to the `ctp7_modules` `$ENV`, compile `gembase` with it's makefile, include `GEMApplication.h` to get the static val, and then link against the library when compiling `daq_monitor.so`?

Would this work...?

bdorney: This then preserves the central location @jsturdy mentioned...?

jsturdy: > Is it possible to have cmsgemos/gembase as a submodule in this (or other repo's)?

Sure it's possible, but that's not a good way to do it..., if you want this definition, it should be provided by an `#include <gemdefs>`, but this can also add some unnecessary circular dependencies.

> This is an RPC module, not the cmsgemos code.

Oops! Nothing to see here, move along :-)

mexanick: Resolving all possible issues with XDAQ-specific includes and definitions only to have few static const members doesn't looks like an easy way to go.

jsturdy: tl;dr, I view the current implementation as preserving the status quo, and not harmful (could add some comment like in the `cmsgemos` case about how we're ashamed of hard-coding those things), but I don't view changing this in accordance with @bdorney's comments as a critical priority

jsturdy (approved):

bdorney (approved):

bdorney:
`conf.txt` is back, and 36 is greater than 32 :D 

approved.

PR #43, opened 2018-08-03
New Feature: Cross-check SBIT Mapping & Rate Calculation in OH FPGA

bdorney:
## Description

I have long suspected there is an issue in the trigger data the v3 electronics are sending.  This provides several new calibration routines (and supporting infrastructure) to qualify the trigger data in the v3 electronics.  When a new set of electronics hardware, CTP7 firmware, or OH firmware is released two main questions should be asked:

1. Is the mapping of `vfat` channel to sbit in the OH correct?
2. Is the measurement of rate of sbits done by the OH/CTP7 correct?

This PR adds the required RPC methods for answering these questions.

#### `setChannelRegistersVFAT3SimpleLocal(...)`
The present RPC method for setting VFAT3 channel registers `setChannelRegistersVFAT3Local(...)` and it's non-local counterpart takes six separate arrays for:

- calEnable,
- mask,
- trim ZCC,
- trim ZCC polarity,
- trim ARM, and
- trim ARM polarity 

While this offers flexibility it doesn't pair well with the getting the channel registers using `getChannelRegistersVFAT3Local(...)` which returns a single array of channel register data where each element is the 16 bit channel register in question.  Additionally looking to the future the DB will most likely *not* store six different columns for the channel register data.

Hence an additional function has been added which takes only a single array and pairs well with the existing `get` method.

It has been incorporated into the same remote method and a check for the key `simple` toggles between `setChannelRegistersVFAT3Local(...)` and `setChannelRegistersVFAT3SimpleLocal(...)`.

See documentation in `include/vfat3.h` found in this PR for usage.

#### `confCalPulseLocal(...)`
Multiple local methods:

- `genScanLocal(...)`,
- `checkSbitMappingWithCalPulseLocal(...), and
- `checkSbitRateWithCalPulseLocal(...)`.

Require configuring the cal pulse for a particular channel.  This method provides the functionality to do so.  

See documentation found in `src/calibration_routines.cpp` for usage.  Not sure if this should be placed in `vfat3.cpp/h` but it seemed appropriate to be found in the calibration library.

#### `checkSbitMappingWithCalPulseLocal(...)`
This specifically addresses the first goal mentioned at the top of this PR.  This method will:

1. Store the current channel register settings,
2. Mask all channels and disable the calpulse for all unmasked vfats,
3. For the vfat of interest unmask a single channel and enable the calpulse,
4. Send `nevts` number of calpulses,
5. readout the sbits observed by the CTP7 and unpack the data,
6. Mask the channel again and disable the calpulse
7. repeat steps 3 through 6 for all channels on the vfat of interest
8. Revert to the original channel register settings

This allows the user to check that the (vfat,chan) being pulsed is the same as the (vfat,sbit) being observed by the OH/CTP7 and confirm the mapping.  It will also report the sbit cluster size.

See documentation in `src/calibration_routines.cpp` for usage.

#### `checkSbitRateWithCalPulseLocal()`
This specifically addresses the second goal mentioned at the top of this PR.  This method will:

1. Store the current channel register settings,
2. Mask all channels and disable the calpulse for all unmasked vfats,
3. For the vfat of interest unmask a single channel and enable the calpulse,
4. Send calpulses at a specific rate using the TTC Generator for a fixed time,
5. readout the rate counters on the  CTP7 and OH,
6. Stop sending calpulses, then mask the channel again and disable the calpulse,
7. repeat steps 3 through 6 for all channels on the vfat of interest
8. revert to the original channel register settings

This allows to confirm that the known rate pulsed matches the measured rate by the CTP7, specifically this is reading out registers:

- `GEM_AMC.OH.OHY.FPGA.TRIG.CNT.VFATX_SBITS` for vfat X of interest on OH Y,
- `GEM_AMC.OH.OHY.FPGA.TRIG.CNT.CLUSTER_COUNT` for OH Y
- `GEM_AMC.TRIGGER.OHY.TRIGGER_RATE` for OH Y

The first is the number of sbits for VFATX seen in `GEM_AMC.OH.OHX.FPGA.TRIG.CNT.SBIT_CNT_TIME_MAX` which then gives the rate.

The second is the sum of all sbit clusters from all vfats observed by the OH.

The third is the rate observed by the CTP7.  

In this case all three values should be the same. 

### Types of changes

## Motivation and Context

Goals 1 & 2 at the top of this page.

## How Has This Been Tested?

http://cmsonline.cern.ch/cms-elog/1058329

Channel pulsed, sbit observed
sbitobsvschanpulsed_validsbits_calenabled_sumofallrates

rate pulsed, rate observed
rateobservedvsratepulsed_validsbits

## Checklist:

### DIFF
--- a/src/calibration_routines.cpp
+++ b/src/calibration_routines.cpp
@@ -1244,7 +1244,7 @@ void checkSbitRateWithCalPulseLocal(localArgs *la, uint32_t *outDataCTP7Rate, ui
     //Prep the SBIT counters
     LOGGER->log_message(LogManager::INFO, stdsprintf("Preping SBIT Counters for ohN %i", ohN));
     writeReg(la, stdsprintf("GEM_AMC.OH.OH%i.FPGA.TRIG.CNT.SBIT_CNT_PERSIST",ohN), 0x0); //reset all counters after SBIT_CNT_TIME_MAX
-    writeReg(la, stdsprintf("GEM_AMC.OH.OH%i.FPGA.TRIG.CNT.SBIT_CNT_TIME_MAX",ohN), 0x02638e98); //count for 1 second
+    writeReg(la, stdsprintf("GEM_AMC.OH.OH%i.FPGA.TRIG.CNT.SBIT_CNT_TIME_MAX",ohN), 0x02638e98*int(waitTime/1000) ); //count for a number of BX's specified by waitTime
 
     if(!((notmask >> vfatN) & 0x1)){
         la->response->set_string("error",stdsprintf("The vfat of interest %i should not be part of the vfats to be masked: %x",vfatN, mask));
--- a/src/calibration_routines.cpp
+++ b/src/calibration_routines.cpp
@@ -63,7 +63,7 @@ std::unordered_map<uint32_t, uint32_t> setSingleChanMask(int ohN, int vfatN, uns
         //store the original channel mask
         sprintf(regBuf, "GEM_AMC.OH.OH%i.GEB.VFAT%i.VFAT_CHANNELS.CHANNEL%i.MASK",ohN,vfatN,chan);
         chanMaskAddr=getAddress(la, regBuf);
-        map_chanOrigMask[chanMaskAddr]=readReg(la, regBuf);   //We'll write this by address later
+        map_chanOrigMask[chanMaskAddr]=readReg(la, regBuf);
 
         //write the new channel mask
         writeRawAddress(chanMaskAddr, chMask, la->response);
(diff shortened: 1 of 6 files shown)


src/calibration_routines.cpp
```diff
@@ -933,6 +966,407 @@ void sbitRateScan(const RPCMsg *request, RPCMsg *response)
...
+ *  \param vfatN specific vfat position to be tested
+ *  \param mask VFATs to be excluded from the trigger
+ *  \param useCalPulse true (false) checks sbit mapping with calpulse on (off); useful for measuring noise
+ *  \param currentPulse Selects whether to use current or volage pulse
+ *  \param calScaleFactor
+ *  \param nevts the number of cal pulses to inject per channel
+ *  \param L1Ainterval How often to repeat signals (only for enable = true)
+ *  \param pulseDelay delay between CalPulse and L1A
+ */
+void checkSbitMappingWithCalPulseLocal(localArgs *la, uint32_t *outData, uint32_t ohN, uint32_t vfatN, uint32_t mask, bool useCalPulse, bool currentPulse, uint32_t calScaleFactor, uint32_t nevts, uint32_t L1Ainterval, uint32_t pulseDelay){
```
mexanick, line 986: Please add somewhere the sbit cluster description

mexanick (changes requested):

src/calibration_routines.cpp
bdorney: Done

mexanick: I meant put this things into description which is picked by documentation handler:
https://github.com/cms-gem-daq-project/ctp7_modules/blob/8d5d5bc6f5d7dd42007acdd21bf1146c2b6ee52f/src/calibration_routines.cpp#L1079

bdorney: I'm not sure I understand, I've added a description here:

> The sbit size is the number of trigger pads from the sbit reported that where part of this cluster, size ranges from 0 to 7.

But I think documentation here should really be in the address table of both the CTP7 and OH firmware (or some guide there in).  This is retrieving info from the `SBIT_MONITOR` block of the CTP7 FW which is being sent sbits on the GEM trigger link from the OH.

It would be better to have this in the FW documentation to prevent it being updated in one place and not forward propagated in another.  

Anyway I can add additional description but I think @andrewpeck and @evka85 should have some documentation for this.

mexanick: I think it is better to have it in two places rather then nowhere =) Please, add these three lines, I believe the code documentation will benefit of this and other software developers will be grateful to have the docs in one place =)

bdorney: okay will do.

mexanick (approved):

PR #47, opened 2018-09-18
New Feature: OH Sysmon from OH FW & Splitting Monitoring Into Dedicated Files 

bdorney:
## Description

Provided a local and non-local versions of function `getmonOHSysmon()` which monitors exposed sysmon registers in OH FW with direct register access rather than using JTAG path (bad).  This function will read the following error conditions:

```
eagle26 > kw GEM_AMC.OH.OH2.FPGA.ADC.CTRL
0x65084000 None         GEM_AMC.OH.OH2.FPGA.ADC.CTRL
0x65084000 r    GEM_AMC.OH.OH2.FPGA.ADC.CTRL.OVERTEMP                   0x00000000
0x65084000 r    GEM_AMC.OH.OH2.FPGA.ADC.CTRL.VCCAUX_ALARM               0x00000000
0x65084000 r    GEM_AMC.OH.OH2.FPGA.ADC.CTRL.VCCINT_ALARM               0x00000000
```

And their number of occurrences:

```
0x6508401c r    GEM_AMC.OH.OH2.FPGA.ADC.CTRL.CNT_OVERTEMP               0x00000000
0x6508401c r    GEM_AMC.OH.OH2.FPGA.ADC.CTRL.CNT_VCCAUX_ALARM           0x00000000
0x6508401c r    GEM_AMC.OH.OH2.FPGA.ADC.CTRL.CNT_VCCINT_ALARM           0x00000000
```

Along with reading the FPGA core temperature, voltage, and I/O voltage.

#### Creating Dedicated Monitoring Library: `daq_monitor.so`
After discussion with @mexanick I've broken `include/amc.h` and corresponding `*.cpp` file into two files:

- `include/daq_monitor.h`, and
- `include/amc.h`.

In the former case all monitoring functions from the old `amc.h` are now located here.

Changes to `Makefile` have been included.

#### Providing `ohMask` parameter for all monitoring functions
We may find in the future that monitoring is required on an AMC in which the number of valid OH's is not sequential, e.g. OH7 may be non-operable and this would need to be masked for whatever reason.  In this case I've provided all monitoring functions the ability to have an `ohMask`.  This mask follows the convention of the current `sca.py` tool, e.g. a 1 in the N^th bit means use the N^th OH.

Additionally to maintain backwards compatibility all monitoring functions will first check if the `ohMask` key exists in the RPC request.  If it does, it will be set using the `get_word` method, if the key does not exist a default value of `0xfff` will be set.  This will ensure that all OH's are considered.  This should maintain current functionality.

### Documentation
All function documentation has been preserved in header file.  For new function `getmonOHSysmon()`, local and non-local versions, please see function documentation in `include/daq_monitor.h`.

### Types of changes

Breaking change because any method in `xhal/cmsgemos` that was previously looking for monitoring functions in `amc.so` will now need to look in `daq_monitor.so`.

## Motivation and Context

Trying to access sysmon information of FPGA using JTAG path is not reliable and causes problems.  Here dedicated register reads are used.

In the future ability to mask an OH from monitoring may be desireable.  Functionality is included here.

Finally wanted to compartmentalise libraries, e.g. providing monitoring in dedicated monitoring library and amc functionality in dedicated amc library.

## How Has This Been Tested?

In progress.

## Checklist:

### DIFF
--- a/include/amc.h
+++ b/include/amc.h
@@ -2,7 +2,7 @@
  *  \brief RPC module for amc methods
  *  \author Cameron Bravo <cbravo135@gmail.com>
  *  \author Mykhailo Dalchenko <mykhailo.dalchenko@cern.ch>
- *  \author Brin Dorney <brian.l.dorney@cern.ch>
+ *  \author Brian Dorney <brian.l.dorney@cern.ch>
  */
 
 #ifndef AMC_H
--- a/include/amc.h
+++ b/include/amc.h
@@ -1,7 +1,7 @@
 /*! \file amc.h
- *  \brief RPC module for optohybrid methods
- *  \author Mykhailo Dalchenko <mykhailo.dalchenko@cern.ch>
+ *  \brief RPC module for amc methods
  *  \author Cameron Bravo <cbravo135@gmail.com>
+ *  \author Mykhailo Dalchenko <mykhailo.dalchenko@cern.ch>
  *  \author Brin Dorney <brian.l.dorney@cern.ch>
  */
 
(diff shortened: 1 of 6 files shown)


include/amc.h
```diff
@@ -1,7 +1,7 @@
 /*! \file amc.h
- *  \brief RPC module for optohybrid methods
- *  \author Mykhailo Dalchenko <mykhailo.dalchenko@cern.ch>
+ *  \brief RPC module for amc methods
  *  \author Cameron Bravo <cbravo135@gmail.com>
+ *  \author Mykhailo Dalchenko <mykhailo.dalchenko@cern.ch>
  *  \author Brin Dorney <brian.l.dorney@cern.ch>
```
mexanick: Mr Brin, you may want to correct this typo ;)

src/calibration_routines.cpp
```diff
@@ -1510,8 +1510,8 @@ void dacScan(const RPCMsg *request, RPCMsg *response){
 
 /*! \fn void dacScanMultiLink(const RPCMsg *request, RPCMsg *response)
  *  \brief As dacScan(...) but for all optohybrids on the AMC
- *  \details Here the RPCMsg request should have a "ohMask" word which specifies which OH's to read from, this is a 12 bit number where a 1 in the n^th bit indicates that the n^th OH should be read back.  Additionally there should be a "ohVfatMaskArray" which is an array of size 12 where each element is the standard vfatMask for OH specified by the array index.
- *  \param request rpc request message
+ *  \details Here the RPCMsg request should have a "ohMask" word which specifies which OH's to read from, this is a 12 bit number where a 1 in the n^th bit indicates that the n^th OH should be read back.
+ :  \param request rpc request message
```
mexanick: Typo: `:` should be `*`

src/daq_monitor.cpp
```diff
@@ -0,0 +1,462 @@
+/*! \file amc.cpp
```
mexanick: `amc.cpp` -> `daq_monitor.cpp`, brief description should be updated as well

mexanick (changes requested):
Couple of typos need fixing, otherwise fine.

include/amc.h
bdorney: whoops :D

src/calibration_routines.cpp
bdorney: yup.

src/daq_monitor.cpp
bdorney: oops.

mexanick (approved):

Issue #49, opened 2018-09-28
Improvement: multi link CTP7 modules should have an NOH key in the RPC request (Good for new comer)


bdorney:
## Brief summary of issue

Right now most of the multi link rpc modules have hardcoded the number of OH's to be 12 and rely on the `ohMask` to skip links (or to not consider additional links).

Due to the different link counts for the GE2/1 OH it may be that we have a different number of OH's on the CTP7.  As a result we should not hard code the number of OH's but instead pass this as a parameter. 

This would be a simple task for a new comer.

### Types of issue

- [ ] Bug report (report an issue with the code)
- [X] Feature request (request for change which adds functionality)

Should *not* be a breaking change, to do this you could simply do a check if the `NOH` key exists:

```C++
uint32_t NOH=12;
if (request->get_key_exists("NOH")){
   NOH = request->get_word("NOH");
}
```

If the key doesn't exist the number of optohybrids is defaulted to 12.

Note that while the number `12` doesn't need 32 bits to be expressed the `rpcsvc` API can only send/receive words in 32 bit representation.  So to prevent a type mismatch you should use `uint32_t`.  This is because the CTP7 itself executes each transaction between the Zynq processor and the FPGA on the AXI bus as a series of 32 bit words. 

## Expected Behavior

The RPC request for each of the following functions should have a key `NOH` which determines the number of optohybrids to be looped over in the OH loop:

https://github.com/cms-gem-daq-project/ctp7_modules/blob/a87d951df725e3477ae6f13fd8f1aeb7480c150d/src/calibration_routines.cpp#L1517
https://github.com/cms-gem-daq-project/ctp7_modules/blob/a87d951df725e3477ae6f13fd8f1aeb7480c150d/src/amc.cpp#L477
https://github.com/cms-gem-daq-project/ctp7_modules/blob/a87d951df725e3477ae6f13fd8f1aeb7480c150d/src/vfat3.cpp#L102
https://github.com/cms-gem-daq-project/ctp7_modules/blob/a87d951df725e3477ae6f13fd8f1aeb7480c150d/src/vfat3.cpp#L287

For an example of the implementation see how the key `NOH` in the RPC request is obtained in:

https://github.com/cms-gem-daq-project/ctp7_modules/blob/a87d951df725e3477ae6f13fd8f1aeb7480c150d/src/amc.cpp#L336-L349 

Then this should be used to control the OH loop.

After making your changes you need to try to compile, you can do this by executing:

[...]

PR #51, opened 2018-10-01
read the number of OHs from the AMC register and perform checks on the NOH rpc key

AndrewLevin:
Currently, the number of OHs is hardcoded as 12 in some places and it is read from the RPC request in other places. This pull request makes use of the NUM_OF_OH AMC register to avoid the hardcoding and perform a check on the number of OHs from in the RPC request.

## Description
The first 4 of the following functions (in amc.cpp, in calibration_routines.cpp, and vfat3.cpp) were previously using hardcoded number of OHs. The next 6 of the following functions (in daq_monitor.cpp) already used the number of OHs from the RPC request. I have changed all of these functions to: use the number of OHs from the RPC request if that exists and is less than or equal to the value of the NUM_OF_OH register, and otherwise to use the NUM_OF_OH register.

1) getOHVFATMaskMultiLink in amc.cpp
2) dacScanMultiLink in calibration_routines.cpp
3) configureVFAT3DacMonitorMultiLink in vfat3.cpp
4) readVFAT3ADCMultiLink in vfat3.cpp
5) getmonTRIGGERmain in daq_monitor.cpp
6) getmonTRIGGEROHmain in daq_monitor.cpp
7) getmonDAQOHmain in daq_monitor.cpp
8) getmonOHmain in daq_monitor.cpp
9) getmonOHSCAmain in daq_monitor.cpp
10) getmonOHSysmon in daq_monitor.cpp

### Types of changes

## Motivation and Context
There may be a different number of links on an OH, as explained in https://github.com/cms-gem-daq-project/ctp7_modules/issues/49. Instead of hardcoding the number or passing the number as a parameter to the rpc call, the number can be easily read from the NUM_OF_OH AMC register. In some cases the user may want to fewer than the total number of OHs, so passing the number as a parameter should be supported in that case.

## How Has This Been Tested?
I have compiled the changes and tested them on eagle64 as described in http://cmsonline.cern.ch/cms-elog/1066648

## Checklist:

### DIFF
--- a/src/amc.cpp
+++ b/src/amc.cpp
@@ -87,10 +87,16 @@ void getOHVFATMaskMultiLink(const RPCMsg *request, RPCMsg *response){
 
     struct localArgs la = {.rtxn = rtxn, .dbi = dbi, .response = response};
 
-    int NOH = readReg(&la, "GEM_AMC.GEM_SYSTEM.CONFIG.NUM_OF_OH");
+    unsigned int NOH = readReg(&la, "GEM_AMC.GEM_SYSTEM.CONFIG.NUM_OF_OH");
+    if (request->get_key_exists("NOH")){
+      if (request->get_word("NOH") <= NOH)
+        NOH = request->get_word("NOH");
+      else
+        LOGGER->log_message(LogManager::WARNING, stdsprintf("NOH requested (%i) > NUM_OF_OH (%i), NOH request will be disregarded",request->get_word("NOH"),NOH));
+    }
     
     uint32_t ohVfatMaskArray[12];
-    for(int ohN=0; ohN<NOH; ++ohN){
+    for(unsigned int ohN=0; ohN<NOH; ++ohN){
         // If this Optohybrid is masked skip it
         if(!((ohMask >> ohN) & 0x1)){
             ohVfatMaskArray[ohN] = 0xffffff;
--- a/src/amc.cpp
+++ b/src/amc.cpp
@@ -89,10 +89,10 @@ void getOHVFATMaskMultiLink(const RPCMsg *request, RPCMsg *response){
 
     unsigned int NOH = readReg(&la, "GEM_AMC.GEM_SYSTEM.CONFIG.NUM_OF_OH");
     if (request->get_key_exists("NOH")){
-      if (request->get_word("NOH") <= NOH)
-        NOH = request->get_word("NOH");
-      else
-        LOGGER->log_message(LogManager::WARNING, stdsprintf("NOH requested (%i) > NUM_OF_OH (%i), NOH request will be disregarded",request->get_word("NOH"),NOH));
+        if (request->get_word("NOH") <= NOH)
+            NOH = request->get_word("NOH");
+        else
+            LOGGER->log_message(LogManager::WARNING, stdsprintf("NOH requested (%i) > NUM_OF_OH (%i), NOH request will be disregarded",request->get_word("NOH"),NOH));
     }
     
     uint32_t ohVfatMaskArray[12];
--- a/src/amc.cpp
+++ b/src/amc.cpp
@@ -92,7 +92,7 @@ void getOHVFATMaskMultiLink(const RPCMsg *request, RPCMsg *response){
         if (NOH_requested <= NOH)
             NOH = NOH_requested;
         else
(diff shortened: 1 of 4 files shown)


AndrewLevin:
@bdorney, @sturdy, @mexanick, all tests were successful, so the pull request is ready for your review

bdorney (changes requested):
Seems like this PR does not provide a check if the key `NOH` exists or not.  This was mentioned in issue #49.  Additionally monitoring functions (now in `daq_monitor.cpp`) could benefit from a similar treatment.  Please add.

AndrewLevin:
@bdorney, so if the `NOH` key exists, I should use it, and if the key does not exist, then I should use the number in the NUM_OF_OH register?

bdorney:
> @dorney, so if the NOH key exists, I should use it, and if the key does not exist, then I should use the number in the NUM_OF_OH register?

Yes, I would also make a check that the value of the `NOH` key does not exceed `NUM_OF_OH` if it does there should be a `WARN` (or `WARNING`?) level message printed to the CTP7 log and then an automatic reset to `NUM_OF_OH` (this should be included in the log)

e.g. "NOH requested > NUM_OF_OH, reseting NOH to NUM_OF_OH"

AndrewLevin:
@bdorney, I made the requested change, rebuilt the .so files, tested them on eagle64, and found the test is successful

src/amc.cpp
```diff
@@ -87,8 +87,16 @@ void getOHVFATMaskMultiLink(const RPCMsg *request, RPCMsg *response){
 
     struct localArgs la = {.rtxn = rtxn, .dbi = dbi, .response = response};
 
+    unsigned int NOH = readReg(&la, "GEM_AMC.GEM_SYSTEM.CONFIG.NUM_OF_OH");
+    if (request->get_key_exists("NOH")){
+        if (request->get_word("NOH") <= NOH)
+            NOH = request->get_word("NOH");
```
bdorney: Why call this method twice?  Why not set to a temp variable and then assign to `NOH` if the `if` statement evaluates to `true`?

bdorney:
> @bdorney, I made the requested change, rebuilt the .so files, tested them on eagle64, and found the test is successful

You're missing the requested changes to `daq_monitor` functions. e.g.

> Additionally monitoring functions (now in daq_monitor.cpp) could benefit from a similar treatment. Please add.

src/amc.cpp
AndrewLevin: @bdorney, I just made an update in which I changed to using a temporary variable

AndrewLevin:
@bdorney, in daq_monitor.cc, NOH appears both as a function argument (e.g. in getmonTRIGGERmainLocal) and as a request key (e.g. in getmonOHmain). Please clarify how you would like me to treat the two cases. In both cases, should I reset NOH to NUM_OF_OH if the NOH request is larger?

bdorney:
> @bdorney, in daq_monitor.cc, NOH appears both as a function argument (e.g. in getmonTRIGGERmainLocal) and as a request key (e.g. in getmonOHmain). Please clarify how you would like me to treat the two cases. In both cases, should I reset NOH to NUM_OF_OH if the NOH request is larger?

I think there's no need to change the local function, since it's taken as an argument and if it's called from another function on the card register access is O(50us).

The non-local versions modify as you have done before.  Does that clarify things?

bdorney:
The real goal is that we want to steer away from hard coding the number of OH's per AMC since this is expected to change in the case of GE2/1; but may of these functions will use pieces of the FW that will be common to all GEM subsystems (e.g. SCA block, Sysmon block, VFAT address space, etc...).

This will reduce overall complexity and increase functionality.  The address space of the FPGA may change, but as long as the node names in remain the same in the xml this can be 100% handled by simply having a different xml and different contents of lmdb on the CTP7.

AndrewLevin:
@bdorney, I have now implemented the same logic for the NOH variables in daq_monitor that are read from the rpc request, please have a look

bdorney:
Please update the PR description to reflect the changes made (e.g. functions modified in daq_monitor are not mentioned in the description).

AndrewLevin:
@bdorney, it is updated now

bdorney (approved):

mexanick (approved):

PR #53, opened 2018-10-04
Making use of unused in bits in dataword of DAC Scan

bdorney:
## Description

Since each DAC of the VFAT has a different size, and in the case of the multilink scan not every OH may be scanned, I've modified the output dataword for the dac scan such that each word includes the VFAT position and the OH number.

This makes determining the origin of the word much easier.

### Types of changes

Change is non breaking, previously bits [32:18] in each word where unused.  Bits [32:27] are still left empty.

## Motivation and Context

See comments above.

## How Has This Been Tested?

See: https://github.com/cms-gem-daq-project/vfatqc-python-scripts/pull/203

## Checklist:

### DIFF
--- a/src/calibration_routines.cpp
+++ b/src/calibration_routines.cpp
@@ -1392,7 +1392,7 @@ std::vector<uint32_t> dacScanLocal(localArgs *la, uint32_t ohN, uint32_t dacSele
     map_dacSelect[36] = std::make_tuple("CFG_THR_ZCC_DAC", 0, 0xff);
     //map_dacSelect[37] = std::make_tuple("NOREG_VTSENSEINT", 0, 0); //Internal temperature sensor
     //map_dacSelect[38] = std::make_tuple("NOREG_VTSENSEEXT", 0, 0); //External temperature sensor (only on HV3b_V3(4) hybrids)
-    map_dacSelect[39] = std::make_tuple("CFG_ADC_VREF", 0, 0x3);
+    map_dacSelect[39] = std::make_tuple("CFG_VREF_ADC", 0, 0x3);
     //map_dacSelect[40] = std::make_tuple("CFG_", 0,);
     //map_dacSelect[41] = std::make_tuple("CFG_", 0,);
 
--- a/src/calibration_routines.cpp
+++ b/src/calibration_routines.cpp
@@ -1461,17 +1461,25 @@ std::vector<uint32_t> dacScanLocal(localArgs *la, uint32_t ohN, uint32_t dacSele
     for(uint32_t dacVal=dacMin; dacVal<=dacMax; dacVal += dacStep){ //Loop over DAC values
         for(int vfatN=0; vfatN<24; ++vfatN){ //Loop over VFATs
             //Skip masked VFATs
-            if ( !( (notmask >> vfatN) & 0x1)) continue;
+            if ( !( (notmask >> vfatN) & 0x1)){ //Case: VFAT is masked, skip
+                //Store word, but with adcVal = 0
+                int idx = vfatN*(dacMax-dacMin+1)/dacStep+(dacVal-dacMin)/dacStep;
+                vec_dacScanData[idx] = ((ohN & 0xf) << 23) + ((vfatN & 0x1f) << 18) + (dacVal & 0xff);
 
-            //Set DAC value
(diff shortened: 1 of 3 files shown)


src/calibration_routines.cpp
```diff
@@ -1471,7 +1471,7 @@ std::vector<uint32_t> dacScanLocal(localArgs *la, uint32_t ohN, uint32_t dacSele
 
             //Store value
             int idx = vfatN*(dacMax-dacMin+1)/dacStep+(dacVal-dacMin)/dacStep;
-            vec_dacScanData[idx] = (adcVal << 8) + dacVal;
+            vec_dacScanData[idx] = (ohN << 23) + (vfatN << 18) + (adcVal << 8) + dacVal;
```
jsturdy: may want to mask these values to protect against possible invalid values

src/calibration_routines.cpp
```diff
@@ -1349,7 +1349,7 @@ void checkSbitRateWithCalPulse(const RPCMsg *request, RPCMsg *response){
  *  \param dacStep step size to scan the dac in
  *  \param mask VFAT mask to use, a value of 1 in the N^th bit indicates the N^th VFAT is masked
  *  \param useExtRefADC if (true) false use the (externally) internally referenced ADC on the VFAT3 for monitoring
- *  \return Returns a std::vector<uint32_t> object of size 24*(dacMax-dacMin+1)/dacStep where dacMax and dacMin are described in the VFAT3 manual.  For each element bits [7:0] are the dacValue and bits [17:8] are the ADC readback value in either current or voltage units depending on dacSelect (again, see VFAT3 manual).
+ *  \return Returns a std::vector<uint32_t> object of size 24*(dacMax-dacMin+1)/dacStep where dacMax and dacMin are described in the VFAT3 manual.  For each element bits [7:0] are the dacValue, bits [17:8] are the ADC readback value in either current or voltage units depending on dacSelect (again, see VFAT3 manual), bits [22:18] are the VFAT position, and bits [26:23] are the optohybrid number.
```
jsturdy, line 1352: since you're not using all bits anyway, why do you reserve 4 for OH number?

src/calibration_routines.cpp
bdorney: I don’t understand what you mean, do you mean like:

((vfatN << 18) & 0x1f)

??

src/calibration_routines.cpp
bdorney: It has dacScanMultiLink() in mind. Here in dacScan() an std::vector<uint32_t> is returned. Then dacScanMultiLink performs a dac scan for unmasked OH’s and copies results for those in the vector that’s used to set the RPC response.  No array positions in the stored word array in the RPC response are reserved for masked OH’s. And since we have to work with 32 bit integers anyway this provides a mechanism for determining which (VFAT,OH) pair the data word came from.

bdorney: The vector solution addresses the problem that the DACs that can be monitored have different sizes

bdorney: Sorry responding from my phone

src/calibration_routines.cpp
jsturdy: > ((vfatN << 18) & 0x1f)

very definitely **not** this :-)
rather, something like `(vfatN &0x1f) << 18`, such that in the event that `vfatN` gets an invalid value, you don't modify bits corresponding to other information

src/calibration_routines.cpp
jsturdy: either i misunderstand your reply, or you misunderstood my point, i.e., the maximum number of OHs is 12, right? you only need 3 bits to represent this number

bdorney: I don't follow, a 3 bit number max's at 7.

jsturdy: fair point, nothing to see here, move along :-D

src/calibration_routines.cpp
bdorney: ahhh, okay.  Yes I see, okay I will update this.

bdorney:
bit mask added to bit positions in data words.  Ready for review.

jsturdy (approved):

mexanick (approved):

PR #60, opened 2018-10-17
Bug Fix: Removed IREF Scan and Addressing #59

### LINKED ISSUE #59: Bug Report: `applyMask(...)` is not used correctly when scanning DACs in higher bit positions in `dacScanLocal(...)`
## Brief summary of issue

There's an issue with the DAC scans when attempting to scan higher bits of global control registers due to writing address instead of node, mask applied incorrectly.

### Types of issue

- [X] Bug report (report an issue with the code)
- [ ] Feature request (request for change which adds functionality)

## Current Behavior

When performing a DAC scan on those DAC's that occupy higher bits of a global control register (e.g. `CFG_BIAS_PRE_I_BLCC` which is bits [13:8]` of `GBL_CFG_BIAS_2`) the value applied in `dacScanLocal(...)`:

https://github.com/cms-gem-daq-project/ctp7_modules/blob/e1d9d0c52a9bd5ffae96e99706ffd12b6ba2809b/src/calibration_routines.cpp#L1478-L1479

Is not correct, e.g. if logging is used to investigate what the value applied when using `applyMask`:

```
Oct 17 07:03:00 eagle64 local0.info rpcsvc[30814]: calibration_routines.dacScanMultiLink: dac Val 0x3f with regMask 0x3f00 and applied value (after Mask) 0x0
```
Instead of `0x3f` written to bits [13:8] a value of `0x0` is written which is incorrect.  Instead of [applyMask](https://github.com/cms-gem-daq-project/ctp7_modules/blob/e1d9d0c52a9bd5ffae96e99706ffd12b6ba2809b/src/utils.cpp#L236-L249) perhaps a bit shift needs to be used instead.

### Steps to Reproduce (for bugs)

1. `run_scans.py dacScanV3 eagle64 0x3 3`

## Possible Solution (for bugs)

The DAC scans are *very* fast as is (~1 minute max) so instead of trying to use the raw address when performing the DAC scan using the node name would be more reliable, e.g. instead of:

```C++
writeRawAddress(regAddr[vfatN], applyMask(dacVal, regMask[vfatN]), la->response);
```

We use:

```C++
writeReg(la, strDacReg, dacVal);
```

Where `strDacReg` is the full node name of the DAC of interest.

Prevents DAC scans on higher bits of global control registers.  Prevents configuring the chip properly.

## Your Environment

* Version used: `e5a6b0e87c8584b2466933b2af264cb3f320e7ed`

bdorney:
## Description

After discussion with VFAT3 team the `CFG_IREF` should never be scanned with the internal ADC's of the VFAT3.  It can only be scanned with an external ADC.  Therefore I've removed it from `dacScanLocal(...)`.

Also fixes #59.

### Types of changes

- [X] Bug fix (non-breaking change which fixes an issue)

Breaking change b/c `CFG_IREF` is no longer allowed.

## Motivation and Context

We where told we could damage the VFAT3 by writing `CFG_IREF` with values below `0x20` and that this cannot be scanned with the internal ADCs of the VFAT3 so it was removed.

Addresses #59.

## How Has This Been Tested?

Ran a dac scan myself:

```bash
run_scans.py dacScanV3 eagle64 0x3 3
```

screen shot 2018-10-17 at 09 35 33

Before this DAC scan was garbage.  Now it matches expectation.

## Checklist:


src/calibration_routines.cpp
```diff
@@ -1460,6 +1454,8 @@ std::vector<uint32_t> dacScanLocal(localArgs *la, uint32_t ohN, uint32_t dacSele
 
     //Set the VFATs into Run Mode
     broadcastWriteLocal(la, ohN, "CFG_RUN", 0x1, mask);
+    LOGGER->log_message(LogManager::INFO, stdsprintf("VFATs not in 0x%x where set to run mode", mask));
```
jsturdy: typo `where` -> `were`

src/calibration_routines.cpp
```diff
@@ -1460,6 +1454,8 @@ std::vector<uint32_t> dacScanLocal(localArgs *la, uint32_t ohN, uint32_t dacSele
...
+    std::this_thread::sleep_for(std::chrono::seconds(1)); //I noticed that DAC values behave weirdly immediately after VFAT is placed in run mode (probably voltage/current takes a moment to stabalize)
```
jsturdy, line 1458: i suppose this second is only at the beginning of the scan, so time is not an issue?
otherwise i'd suggest finding the minimum time required for stabilization

jsturdy (commented):
Review comment also linked to comment in #59

src/calibration_routines.cpp
bdorney: > i suppose this second is only at the beginning of the scan, so time is not an issue?

That is correct.  It happens only once, at the start of the scan, it poses no issue.

mexanick (approved):

jsturdy (approved):
looks good to me

BenjaminRS (approved):
Was going to have the same comment as Jared (re minimum sleep time) - but i see it is resolved

Issue #62, opened 2018-10-25
Maintenance: Change Local CTP7 Modules to Struct's as input arguments (Good for Newcomer)


bdorney:
<!--- Provide a general summary of the issue in the Title above -->
## Brief summary of issue
<!--- Provide a description of the issue, including any other issues or pull requests it references -->
Several of our `rpcmodule`s have *huge* argument lists that are cumbersome to use and also maintain; @mexanick and I have wanted to trim this down for some time to use several struct's to handle this instead. 

### Types of issue
<!--- Propsed labels (see CONTRIBUTING.md) to help maintainers label your issue: -->
- [ ] Bug report (report an issue with the code)
- [ ] Feature request (request for change which adds functionality)
- [x] Maintenance (improves long-term maintainability of the code base)

## Expected Behavior
<!--- If you're describing a bug, tell us what should happen -->
<!--- If you're suggesting a change/improvement, tell us how it should work -->
There are a set of common parameters that are typically passed to the local functions on the CTP7.  These could be stored in a set of `struct`'s and then instances of these `struct`'s could be passed as input.

Goes without saying but the non-local functions which call the local functions mentioned below should *also* be modified accordingly.

### Expected Struct's

```C++
    struct ParamCalPulse{
        bool enable; //true (false) turn on (off) calpulse
        bool isCurrent; //true (false) is current injection (voltage pulse)

        uint32_t duration; //duration in BX's (CFG_CAL_DUR)
        uint32_t extVoltStep; //External voltage step 0->disable; 1->enable (CFG_CAL_EXT)
        uint32_t height; //height of calpulse (CFG_CAL_DAC)
        uint32_t phase; //phase of calpulse (CFG_CAL_PHI)
        uint32_t polarity; //polarity of calpulse 0->pos; 1->neg (CFG_CAL_SEL_POL)
        uint32_t scaleFactor; //current pulse scale factor (CFG_CAL_FS)

        ParamCalPulse(){
            enable = false;
            isCurrent = false;

            duration = 0x1ff;
            extVoltStep = 0x0;
            height = 0x0;                                                                                                                                                                                                                
            phase = 0x0;
            polarity = 0x0;
            scaleFactor = 0x0;
        }
    }; //End ParamCalPulse

    struct ParamScan{
        //Hardware selection
        uint32_t ohN;
        uint32_t ohMask;
        uint32_t vfatN;
        uint32_t vfatMask;
        uint32_t chan;      //channel of interest

[...]

PR #63, opened 2018-10-28
[WIP] add three new structs to pass a large number of arguments cleanly

AndrewLevin:
Some of the functions in calibration_routines.cpp and optohybrid.cpp now take a large number of arguments. There is also large overlap between the arguments that the different functions take. We can use structs to more elegantly pass these arguments.

## Description
Three new structs have been defined in `include/utils.h` for calpulse-related arguments, scan-related arguments, and ttcgen-related arguments:

```
struct ParamCalPulse{
    bool enable; //true (false) turn on (off) calpulse
    bool isCurrent; //true (false) is current injection (voltage pulse)

    uint32_t duration; //duration in BX's (CFG_CAL_DUR)
    uint32_t extVoltStep; //External voltage step 0->disable; 1->enable (CFG_CAL_EXT)
    uint32_t height; //height of calpulse (CFG_CAL_DAC)
    uint32_t phase; //phase of calpulse (CFG_CAL_PHI)
    uint32_t polarity; //polarity of calpulse 0->pos; 1->neg (CFG_CAL_SEL_POL)
    uint32_t scaleFactor; //current pulse scale factor (CFG_CAL_FS)

    ParamCalPulse(){
        enable = false;
        isCurrent = false;

        duration = 0x1ff;
        extVoltStep = 0x0;
        height = 0x0;
        phase = 0x0;
        polarity = 0x0;
        scaleFactor = 0x0;
    }
}; //End ParamCalPulse

struct ParamScan{
    //Hardware selection
    uint32_t ohN;
    uint32_t ohMask;
    uint32_t vfatN;
    uint32_t vfatMask;
    uint32_t chan;      //channel of interest

    //Params
    bool useUltra;
    bool useExtTrig;

    uint32_t dacMax;    //Maximum dac value
    uint32_t dacMin;    //Minimum dac value
    uint32_t dacSelect; //DAC to use for Monitoring
    uint32_t dacStep;   //step size for dac
    uint32_t nevts;     //Number of events
    uint32_t waitTime;  //unit of time; uints depend on function

    std::string scanReg;     //Register to scan against

    ParamScan(){
        vfatMask = 0x0;
        ohMask = 0xfff;

        useUltra = true;
        useExtTrig = false;

        dacMax=254;
        dacMin=0;
        dacStep=1;
        nevts=100;
    }
}; //End ParamScan

struct ParamTtcGen{
    bool enable;

    uint32_t L1Ainterval;
    uint32_t mode;
    uint32_t nPulses;
    uint32_t pulseDelay;
    uint32_t pulseRate;
    uint32_t type;

    ParamTtcGen(){
        enable = false;

        L1Ainterval = 250;
        pulseDelay = 40;
        pulseRate = 40079000 / L1Ainterval;
    }

    uint32_t calcRate(){
        if(L1Ainterval > 0){
            pulseRate = 40079000 / L1Ainterval;
        }
        else{
            pulseRate = 0;
        }

        return pulseRate;
    }
}; //End ParamTtcGen
```
The following 9 functions have been modified to take these structs:

`confCalPulseLocal` in `src/calibration_routines.cpp`
`ttcGenConfLocal` in `src/calibration_routines.cpp`
`genScanLocal` in `src/calibration_routines.cpp`
`sbitRateScanLocal` in `src/calibration_routines.cpp`
`sbitRateScanParallelLocal` in `src/calibration_routines.cpp`
`checkSbitMappingWithCalPulseLocal` in `src/calibration_routines.cpp`
`checkSbitRateWithCalPulseLocal` in `src/calibration_routines.cpp`
`dacScanLocal` in `src/calibration_routines.cpp`
`configureScanModuleLocal` in `src/optohybrid.cpp`
`getUltraScanResultsLocal` in `src/optohybrid.cpp`

All of the places where these functions are called have been modified to pass these structs.

### Types of changes

## Motivation and Context
This change was requested in https://github.com/cms-gem-daq-project/ctp7_modules/issues/62

## How Has This Been Tested?
All of the scans that are affected by this update have been run on the coffin setup with and without this update, and no statistically significant changes are apparent: http://cmsonline.cern.ch/cms-elog/1077762

## Checklist:

### DIFF
--- a/include/calibration_routines.h
+++ b/include/calibration_routines.h
@@ -71,15 +71,15 @@ std::unordered_map<uint32_t, uint32_t> setSingleChanMask(int ohN, int vfatN, uns
  */
 void applyChanMask(std::unordered_map<uint32_t, uint32_t> map_chanOrigMask, localArgs *la);
 
-/*! \fn void confCalPulseLocal(localArgs *la, uint32_t ohN, uint32_t mask, uint32_t ch, bool toggleOn, bool currentPulse, uint32_t calScaleFactor)
+/*! \fn bool confCalPulseLocal(localArgs *la, ParamCalPulse *calParams, ParamScan *scanParams)
  *  \brief Configures the calibration pulse for channel ch on all VFATs of ohN that are not in mask to either be on (toggleOn==true) or off (toggleOn==false).  If ch == 128 and toggleOn == False will write the CALPULSE_ENABLE bit for all channels of all vfats that are not masked on ohN to 0x0.
  *  \param la Local arguments structure
  *  \param ohN (= scanParams->oh) Optical link number
  *  \param mask (= scanParams->vfat) VFAT mask 
  *  \param ch (= scanParams->ch) Channel of interest 
  *  \param toggleOn (= calParams->enable) if true (false) turns the calibration pulse on (off) for channel ch 
- *  \param currentPulse (= calParams->isCurrent) Selects whether to use current or volage pulse (passed through calParams)
- *  \param calScaleFactor (= calParams->scaleFactor) Scale factor for the calibration pulse height (00 = 25%, 01 = 50%, 10 = 75%, 11 = 100%) (passed through calParams)
+ *  \param currentPulse (= calParams->isCurrent) Selects whether to use current or volage pulse
+ *  \param calScaleFactor (= calParams->scaleFactor) Scale factor for the calibration pulse height (00 = 25%, 01 = 50%, 10 = 75%, 11 = 100%)
  */
 bool confCalPulseLocal(localArgs *la, ParamCalPulse *calParams, ParamScan *scanParams);
 
@@ -114,7 +114,7 @@ void ttcGenToggleLocal(localArgs * la, uint32_t ohN, bool enable);
  */
 void ttcGenToggle(const RPCMsg *request, RPCMsg *response);
 
-/*! \fn void ttcGenConfLocal(localArgs * la, uint32_t ohN, uint32_t mode, uint32_t type, uint32_t pulseDelay, uint32_t L1Ainterval, uint32_t nPulses, bool enable)
+/*! \fn void ttcGenConfLocal(localArgs * la, uint32_t ohN, ParamTTCGen *ttcParams)
  *  \brief Configures TTC generator. Local callable version of ttcGenConf
  *
  *  - **v3**  electronics behavior:
@@ -139,12 +139,12 @@ void ttcGenToggle(const RPCMsg *request, RPCMsg *response);
  *
  *  \param la Local arguments structure
  *  \param ohN Optical link
- *  \param mode T1 controller mode
- *  \param type Type of T1 signal to send
- *  \param pulseDelay Delay between CalPulse and L1A
- *  \param L1Ainterval How often to repeat signals (only for enable = true)
- *  \param nPulses Number of calibration pulses to generate
- *  \param enable If true (false) ignore (take) ttc commands from backplane for this AMC (affects all links)
+ *  \param mode (= ttcParams->mode) T1 controller mode
+ *  \param type (= ttcParams->type) Type of T1 signal to send
+ *  \param pulseDelay (= ttcParams->delay) Delay between CalPulse and L1A
+ *  \param L1Ainterval (= ttcParams->interval) How often to repeat signals (only for enable = true)
+ *  \param nPulses (= ttcParams->nPulses) Number of calibration pulses to generate
+ *  \param enable (= ttcParams->enable) true (false) ignore (take) ttc commands from backplane for this AMC (affects all links)
  */
 void ttcGenConfLocal(localArgs * la, uint32_t ohN, ParamTTCGen *ttcParams);
 
@@ -152,46 +152,46 @@ void ttcGenConfLocal(localArgs * la, uint32_t ohN, ParamTTCGen *ttcParams);
  *  \brief Configures TTC generator
  *
  *  - **v3**  electronics behavior:
- *    * pulseDelay (= ttcParams->delay) (only for enable = true), delay between CalPulse and L1A
- *    * L1Ainterval (= ttcParams->interval) (only for enable = true), how often to repeat signals
- *    * enable (= ttcParams->enable) = true (false) ignore (take) ttc commands from backplane for this AMC (affects all links)
+ *    * pulseDelay (only for enable = true), delay between CalPulse and L1A
+ *    * L1Ainterval (only for enable = true), how often to repeat signals
+ *    * enable = true (false) ignore (take) ttc commands from backplane for this AMC (affects all links)
  *  - **v2b** electronics behavior:
  *    * Configure the T1 controller
- *    * mode (= ttcParams->mode):
+ *    * mode:
  *      * 0 (Single T1 signal),
  *      * 1 (CalPulse followed by L1A),
  *      * 2 (pattern)
- *    * type (= ttcParams->type) (only for mode 0, type of T1 signal to send):
+ *    * type (only for mode 0, type of T1 signal to send):
  *      * 0 L1A
  *      * 1 CalPulse
  *      * 2 Resync
  *      * 3 BC0
- *    * pulseDelay (= ttcParams->delay) (only for mode 1), delay between CalPulse and L1A
- *    * L1Ainterval (= ttcParams->interval) (only for mode 0,1), how often to repeat signals
- *    * nPulses (ttcParams->nPulses) how many signals to send (0 is continuous)
- *    * enable (= ttcParams->enable)= true (false) start (stop) the T1Controller for link ohN
+ *    * pulseDelay (only for mode 1), delay between CalPulse and L1A
+ *    * L1Ainterval (only for mode 0,1), how often to repeat signals
+ *    * nPulses how many signals to send (0 is continuous)
+ *    * enable = true (false) start (stop) the T1Controller for link ohN
  *
  *  \param request RPC request message
  *  \param response RPC response message
  */
 void ttcGenConf(const RPCMsg *request, RPCMsg *response);
 
-/*! \fn void genScanLocal(localArgs *la, uint32_t *outData, uint32_t ohN, uint32_t mask, uint32_t ch, bool useCalPulse, bool currentPulse, uint32_t calScaleFactor, uint32_t nevts, uint32_t dacMin, uint32_t dacMax, uint32_t dacStep, std::string scanReg, bool useUltra, bool useExtTrig)
+/*! \fn void genScanLocal(localArgs *la, uint32_t *outData, ParamCalPulse *calParams, ParamScan *scanParams, bool useExtTrig)
  *  \brief Generic calibration routine. Local callable version of genScan
  *  \param la Local arguments structure
  *  \param outData pointer to the results of the scan
- *  \param ohN Optical link
- *  \param mask VFAT mask
- *  \param ch Channel of interest
- *  \param useCalPulse Use  calibration pulse if true
- *  \param currentPulse Selects whether to use current or volage pulse
- *  \param calScaleFactor
- *  \param nevts Number of events per calibration point
- *  \param dacMin Minimal value of scan variable
- *  \param dacMax Maximal value of scan variable
- *  \param dacStep Scan variable change step
- *  \param scanReg DAC register to scan over name
- *  \param useUltra Set to 1 in order to use the ultra scan
+ *  \param ohN (= scanParams->oh) Optical link
+ *  \param mask (= scanParams->vfat) VFAT mask
+ *  \param ch (= scanParams->chan) Channel of interest
+ *  \param useCalPulse (= calParams->enable) Use  calibration pulse if true
+ *  \param currentPulse (= calParams->isCurrent) Selects whether to use current or volage pulse
+ *  \param calScaleFactor (= calParams->scaleFactor)
+ *  \param nevts (= scanParams->nevts) Number of events per calibration point
+ *  \param dacMin (= scanParams->min) Minimal value of scan variable
+ *  \param dacMax (= scanParams->max) Maximal value of scan variable
+ *  \param dacStep (= scanParams->step) Scan variable change step
+ *  \param scanReg (= scanParams->scanReg) DAC register to scan over name
+ *  \param useUltra (= scanParams->useUltra) Set to 1 in order to use the ultra scan
  *  \param useExtTrig Set to 1 in order to use the backplane triggers
  */
 void genScanLocal(localArgs *la, uint32_t *outData, ParamCalPulse *calParams, ParamScan *scanParams, bool useExtTrig);
@@ -203,7 +203,7 @@ void genScanLocal(localArgs *la, uint32_t *outData, ParamCalPulse *calParams, Pa
  */
 void genScan(const RPCMsg *request, RPCMsg *response);
 
-/*! \fn void sbitRateScanLocal(localArgs *la, uint32_t *outDataDacVal, uint32_t *outDataTrigRate, uint32_t ohN, uint32_t maskOh, bool invertVFATPos, uint32_t ch, uint32_t dacMin, uint32_t dacMax, uint32_t dacStep, std::string scanReg, uint32_t waitTime)
+/*! \fn void sbitRateScanLocal(localArgs *la, uint32_t *outDataDacVal, uint32_t *outDataTrigRate, ParamScan *scanParams, bool invertVFATPos)
  *  \brief SBIT rate scan. Local version of sbitRateScan
  *
  *  * Measures the SBIT rate seen by OHv3 ohN for the non-masked VFAT found in maskOh as a function of scanReg
@@ -218,19 +218,19 @@ void genScan(const RPCMsg *request, RPCMsg *response);
  *  \param la Local arguments structure
  *  \param outDataDacVal
  *  \param outDataTrigRate
- *  \param ohN Optohybrid optical link number
- *  \param maskOh VFAT mask, should only have one unmasked chip
+ *  \param ohN (= scanParams.oh) Optohybrid optical link number
+ *  \param maskOh (= scanParams.vfat) VFAT mask, should only have one unmasked chip
  *  \param invertVFATPos is for FW backwards compatiblity; if true then the vfatN =  23 - map_maskOh2vfatN[maskOh]
- *  \param ch Channel of interest
- *  \param dacMin Minimal value of scan variable
- *  \param dacMax Maximal value of scan variable
- *  \param dacStep Scan variable change step
- *  \param scanReg DAC register to scan over name
- *  \waitTime Measurement duration per point in milliseconds
+ *  \param ch (= scanParams->chan) Channel of interest
+ *  \param dacMin (= scanParams->min) Minimal value of scan variable
+ *  \param dacMax (= scanParams->max) Maximal value of scan variable
+ *  \param dacStep (= scanParams->step) Scan variable change step
+ *  \param scanReg (= scanParams->scanReg) DAC register to scan over name
+ *  \param waitTime (= scanParams->waitTime) Measurement duration per point in milliseconds
  */
 void sbitRateScanLocal(localArgs *la, uint32_t *outDataDacVal, uint32_t *outDataTrigRate, ParamScan *scanParams, bool invertVFATPos);
 
-/*! \fn void sbitRateScanParallelLocal(localArgs *la, uint32_t *outDataDacVal, uint32_t *outDataTrigRatePerVFAT, uint32_t *outDataTrigRateOverall, uint32_t ohN, uint32_t vfatmask, uint32_t ch, uint32_t dacMin, uint32_t dacMax, uint32_t dacStep, std::string scanReg)
+/*! \fn void sbitRateScanParallelLocal(localArgs *la, uint32_t *outDataDacVal, uint32_t *outDataTrigRatePerVFAT, uint32_t *outDataTrigRateOverall,ParamScan *scanParams)
  *  \brief Parallel SBIT rate scan. Local version of sbitRateScan
  *
  *  * Measures the SBIT rate seen by OHv3 ohN for the non-masked VFATs defined in vfatmask as a function of scanReg
@@ -262,7 +262,7 @@ void sbitRateScanParallelLocal(localArgs *la, uint32_t *outDataDacVal, uint32_t 
  */
 void sbitRateScan(const RPCMsg *request, RPCMsg *response);
 
-/*! \fn void checkSbitMappingWithCalPulseLocal(localArgs *la, uint32_t *outData, uint32_t ohN, uint32_t mask, bool currentPulse, uint32_t calScaleFactor, uint32_t nevts, uint32_t L1Ainterval, uint32_t pulseDelay)
+/*! \fn void checkSbitMappingWithCalPulseLocal(localArgs *la, uint32_t *outData, ParamCalPulse *calParams, ParamScan *scanParams, ParamTTCGen *ttcParams, uint32_t mask)
  *  \brief With all but one channel masked, pulses a given channel, and then checks which sbits are seen by the CTP7, repeats for all channels on vfatN; reports the (vfat,chan) pulsed and (vfat,sbit) observed where sbit=chan*2; additionally reports if the cluster was valid.
  *  \details The SBIT Monitor stores the 8 SBITs that are sent from the OH (they are all sent at the same time and correspond to the same clock cycle). Each SBIT clusters readout from the SBIT Monitor is a 16 bit word with bits [0:10] being the sbit address and bits [12:14] being the sbit size, bits 11 and 15 are not used.
  *  \details The possible values of the SBIT Address are [0,1535].  Clusters with address less than 1536 are considered valid (e.g. there was an sbit); otherwise an invalid (no sbit) cluster is returned.  The SBIT address maps to a given trigger pad following the equation \f$sbit = addr % 64\f$.  There are 64 such trigger pads per VFAT.  Each trigger pad corresponds to two VFAT channels.  The SBIT to channel mapping follows \f$sbit=floor(chan/2)\f$.  You can determine the VFAT position of the sbit via the equation \f$vfatPos=7-int(addr/192)+int((addr%192)/64)*8\f$.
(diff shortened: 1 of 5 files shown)


src/calibration_routines.cpp
```diff
@@ -58,17 +58,22 @@ void applyChanMask(std::unordered_map<uint32_t, uint32_t> map_chanOrigMask, loca
     }
 }
 
-/*! \fn void confCalPulseLocal(localArgs *la, uint32_t ohN, uint32_t mask, uint32_t ch, bool toggleOn, bool currentPulse, uint32_t calScaleFactor)
+/*! \fn void confCalPulseLocal(localArgs *la, ParamCalPulse calParams, ParamScan scanParams)
```
bdorney: Pass these as pointers; you’re duplicating the memory needed.

src/calibration_routines.cpp
```diff
@@ -387,33 +396,49 @@ void ttcGenConf(const RPCMsg *request, RPCMsg *response)
...
- *  \param calScaleFactor
- *  \param nevts Number of events per calibration point
- *  \param dacMin Minimal value of scan variable
- *  \param dacMax Maximal value of scan variable
- *  \param dacStep Scan variable change step
- *  \param scanReg DAC register to scan over name
- *  \param useUltra Set to 1 in order to use the ultra scan
- *  \param useExtTrig Set to 1 in order to use the backplane triggers
+ *  \param calParams Arguments related to the calibration pulse (enable, isCurrent, scaleFactor)
+ *  \param scanParams Scan-related arguments (useExtTrig, useUltra, ohN, nevts, vfatMask, chan, dacMin, dacMax, dacStep, scanReg)
```
bdorney: This help is not descriptive, what is scanParams? E.g. it’s an instance of the ParamsScan struct.

src/calibration_routines.cpp
```diff
@@ -387,33 +396,49 @@ void ttcGenConf(const RPCMsg *request, RPCMsg *response)
...
+    ParamScan ps;
+
+    ps.ohN = ohN;
+    
+    ptg.mode = mode;
+    ptg.type = type;
+    ptg.pulseDelay = pulseDelay;
+    ptg.L1Ainterval = L1Ainterval;
+    ptg.nPulses = nPulses;
+    ptg.enable = enable;
```
bdorney: Variable names should be descriptive; easy to understand and maintain. Try to use a common convention. E.g. in local functions you’re using scanParams, and in non-local functions you’re using ps. Be consistent.

src/calibration_routines.cpp
```diff
@@ -658,13 +685,32 @@ void genScan(const RPCMsg *request, RPCMsg *response)
...
+
+    ps.ohN = ohN;
+    ps.vfatMask = mask;
+    ps.chan = ch;
+    ps.nevts = nevts;
+    ps.dacMin = dacMin;
+    ps.dacMax = dacMax;
+    ps.dacStep = dacStep;
+    ps.scanReg = scanReg;
+    ps.useUltra = useUltra;
```
bdorney: You’re duplicating memory here. E.g. you’re creating a `bool` `useUltra` to store the value from the rpc msg and then you’re only ever using this `bool` to set the corresponding bool in the `ParamsScan` struct.  This is sloppy and will be hard to maintain in the long run. Just set the struct when getting the value from the rpc message.

bdorney (commented):
Know this is WIP but direction should be corrected before you get to far in testing. Propagate comments through to all functions

src/calibration_routines.cpp
bdorney: Book = bool

src/calibration_routines.cpp
AndrewLevin: done

src/calibration_routines.cpp
AndrewLevin: changed

src/calibration_routines.cpp
AndrewLevin: fixed

src/calibration_routines.cpp
AndrewLevin: changed

AndrewLevin:
@bdorney, which tests would you look to see run for this PR?

bdorney:
> @bdorney, which tests would you look to see run for this PR?

Every function that you modified should be tested on a real system to ensure behavior hasn't changed.  You should first trace down which rpc modules are used by which SW tools and confirm that they are still usable and give expected results/outputs/behavior.  This might require an afternoon or two of testing on the coffin setup.

AndrewLevin:
@bdorney, where did you envision that the `calcRate` function in the `ParamTtcGen` `struct` would be used?

bdorney:
> @bdorney, where did you envision that the calcRate function in the ParamTtcGen struct would be used?

If a user provides an `L1AInterval` but not a rate this `L1AInterval` is then used to determine the pulse rate via `calcRate`.  I'm not sure how applicable that is to `ctp7_modules` it was originally an idea for a struct exposed to `python` via boost [here](https://github.com/cms-gem-daq-project/xhal/pull/89/files).

AndrewLevin:
ok, in ctp7_modules, it seems that the inverse case, the `pulseRate` is used to calculate the `L1AInterval`, is used:

https://github.com/cms-gem-daq-project/ctp7_modules/blob/ab9f57487ab79e3f37710da4db84835d80b1401f/src/calibration_routines.cpp#L1195-L1201

bdorney:
> ok, in ctp7_modules, it seems that the inverse case, the `pulseRate` is used to calculate the `L1AInterval`, is used:
> 
> [ctp7_modules/src/calibration_routines.cpp](https://github.com/cms-gem-daq-project/ctp7_modules/blob/ab9f57487ab79e3f37710da4db84835d80b1401f/src/calibration_routines.cpp#L1195-L1201)
> Lines 1195 to 1201 in [ab9f574](/cms-gem-daq-project/ctp7_modules/commit/ab9f57487ab79e3f37710da4db84835d80b1401f)
>  uint32_t L1Ainterval; 
>  if(pulseRate > 0){ 
>      L1Ainterval = int(40079000 / pulseRate); 
>  } 
>  else{ 
>      L1Ainterval = 0; 
>  }

That's only in this specific function which is specifically making a test of rate, the user does not supply `L1AInterval` here, in *other* functions `L1AInterval` is specified.

bdorney:
You'll need to update your local copy of `develop` and rebase this PR off of that since https://github.com/cms-gem-daq-project/ctp7_modules/pull/71 was merged.

include/utils.h
```diff
@@ -23,6 +23,104 @@
...
+        useExtTrig = false;
+
+        dacMax=254;
+        dacMin=0;
+        dacStep=1;
+        nevts=100;
+    }
+}; //End ParamScan
+
+/*! \struct ParamTtcGen
```
jsturdy: Re comments in other threads, style should be to use "almost" camel case: for types (`struct`s and `class`es, use `PascalCase`, variables use `CamelCase`, the distinction is that the first letter is capitalized for `PascalCase`)
Then, within the name, any acronym, e.g., `TTC`, `DAC`, should *always* be capitalized completely, or completely uncapitalized, based on how the style rule would otherwise apply, e.g., 
```c++
ParamTTCGen ttcConfig;
```

include/utils.h
```diff
@@ -23,6 +23,104 @@
...
+ *  Contains arguments related to ttc generation
+ */
+struct ParamTtcGen{
+    bool enable; //If true (false) ignore (take) ttc commands from backplane for this AMC (affects all links)
+
+    uint32_t L1Ainterval; //How often to repeat signals
```
jsturdy: Here, as in the previous comment, the style rule is not followed, so I would suggest `L1Ainterval` -> `interval` **unless**, the TTC module can be configured to have different intervals, e.g., interval between L1A and some other command (which I see you have already defined below) and "interval" would become ambiguous.

How else can this module be configured?
* Is `nPulses`  just the number of pulses?  Or is it the number of `type` signals? Does the module keep sending L1As after sending `nPulses` pulses?
* I'd suggest dropping `pulse` from the variable names, as it seems that this `struct` could(should?) be rather more generic

src/calibration_routines.cpp
```diff
@@ -1311,9 +1493,16 @@ void dacScanMultiLink(const RPCMsg *request, RPCMsg *response){
...
+
+        scanParams.ohN = ohN;
+        scanParams.dacSelect = dacSelect;
+        scanParams.dacStep = dacStep;
+        scanParams.vfatMask = vfatMask;
+        
         //Get dac scan results for this optohybrid
         LOGGER->log_message(LogManager::INFO, stdsprintf("Performing DAC Scan for OH%i", ohN));
-        dacScanResults = dacScanLocal(&la, ohN, dacSelect, dacStep, vfatMask, useExtRefADC);
+        dacScanResults = dacScanLocal(&la, &scanParams, useExtRefADC);
```
jsturdy: Got some extra whitespace here.

include/utils.h
```diff
@@ -23,6 +23,104 @@
...
+ *  Contains arguments related to scans
+ */
+struct ParamScan{
+    //Hardware selection
+    uint32_t ohN;       //optical link number
+    uint32_t ohMask;    //OH mask
+    uint32_t vfatN;     //VFAT number
+    uint32_t vfatMask;  //VFAT mask
+    uint32_t chan;      //channel of interest
+
```
jsturdy, line 61: Ignorance here, but are `ohN` and `ohMask` ever used concurrently? Does the function using this `ParamScan` `struct` know whether it's operating on a single OH or on multiples? (same comment for `vfatN` and `vfatMask`)
If they are used simultaneously, the structure makes sense, if not, you can get by with just `oh` and `vfat` and the code that uses this `struct` would appropriately fill the parameter with either the specific device, or the masked list of devices (on the other end, the code receiving the `struct` should also not get confused, so the first question should be answered also for that)

include/utils.h
```diff
@@ -23,6 +23,104 @@
...
+    //Params
+    bool useUltra;   //Set to 1 in order to use the ultra scan
+    bool useExtTrig; //Set to 1 in order to use the backplane triggers
+
+    uint32_t dacMax;    //Maximum dac value
+    uint32_t dacMin;    //Minimum dac value
+    uint32_t dacSelect; //DAC to use for Monitoring
+    uint32_t dacStep;   //step size for dac
```
jsturdy: Is `dac` really necessary here? Why not just `min/max/select/step`?

jsturdy (changes requested):
Couple of tweaks, couple of suggested improvements

include/utils.h
AndrewLevin: done

src/calibration_routines.cpp
AndrewLevin: fixed

include/utils.h
AndrewLevin: I would say they are needed, since this struct is used for many types of scans, and just having `min` and `max` could lead to confusion. For example, in an scurve scan `min` and `max` could potentially refer to the minimum and maximum channel number to be scanned.

jsturdy: Can they?  Is it not understood (documented) that these are for the thing being scanned (channel is very rarely (never?) *the* thing that is being scanned).  The `struct` already has a channel field, no?
Since this `struct` was created with `dac` in the field names, it's clearly **not** been used for this purpose yet.
I suppose this is the trouble with trying to implement a generic object...
Think through all the current use cases (and if there are future ones that are envisioned, but not yet implemented), and then make sure that the `struct` is well documented for the various use cases.  If that is done, then the potential confusion becomes a problem with someone not reading the documentation...

AndrewLevin: OK, I have changed `dacMin`, `dacMax`, and `dacStep` to just `min`, `max`, and `step`. I left `dacSelect` as `dacSelect`, since that is a special setting for the DAC scans

include/utils.h
AndrewLevin: `vfatN` and `vfatMask` are used simultaneously here:

https://github.com/cms-gem-daq-project/ctp7_modules/blob/d565057cf54791a568cb2f922963c0291cb153e6/src/calibration_routines.cpp#L833-L878

I guess it is because the current pulse needs to be disabled in all channels in all vfats that are present other than the one of interest, and we cannot assume that all vfats 24 vfats are present or alive.

`ohN` and `ohMask` are used simultaneously in a case where `ohMask` actually represents a mask of vfats (see the discussion in https://github.com/cms-gem-daq-project/ctp7_modules/issues/62#issuecomment-433712988):

https://github.com/cms-gem-daq-project/ctp7_modules/blob/d565057cf54791a568cb2f922963c0291cb153e6/src/calibration_routines.cpp#L655-L658

include/utils.h
AndrewLevin: I have changed `L1AInterval` to `interval` and `pulseDelay` to `delay` and left the other members of `ParamTTCGen` the same. It is not clear to me either whether`nPulses` only applies to the pulses or other types of signals, but it seems that it is only relevant for v2b electronics

https://github.com/cms-gem-daq-project/ctp7_modules/blob/d565057cf54791a568cb2f922963c0291cb153e6/include/calibration_routines.h#L154-L172

which I thought were already deprecated?

include/utils.h
jsturdy: * In the first example `ohN` is an index and `vfatN` is as well. `vfatMask` is the mask indicating which VFATs are "alive" but should not "participate".  Is this routine designed to only ever run on a *single* VFAT on a *single* at any given time?  If yes, then any `mask` can be inferred/derived based on the provided `ohN` and `vfatN` fields and the `mask` fields become redundant.

* In the second example I see `ohN` is used as an index, and "`ohMask`" is a bitmask of VFATs.  Again, here one is `OH` specific (`ohN`) and one is `VFAT` specific ("`ohMask`"), and they could equally well be served by a single `ohN` and `vfatN` (or more appropriately named) field

In general, I don't feel that this particular `struct` has been optimally designed as is, but as it is so tightly coupled to this PR, I can't realistically ask you to decouple the developments related to it.

I can't tell you right now what the "best" design would be (as it *strongly* depends on the types of scans that will be serviced by this `struct`), but what I can tell you is that if you have the bare minimum required parameters that *all* scans share, you can work towards which of them can have different meanings, depending on the type of the scan.  Additionally, I would strongly caution that this `struct` should **not** be a catchall for *any* parameter that *any* scan could need, which is what it feels like currently and leads to my aforementioned feeling about the design (apologies, I should have paid closer attention in #62), but from the original function signatures of:
```
confCalPulseLocal in src/calibration_routines.cpp
ttcGenConfLocal in src/calibration_routines.cpp
genScanLocal in src/calibration_routines.cpp
sbitRateScanLocal in src/calibration_routines.cpp
sbitRateScanParallelLocal in src/calibration_routines.cpp
checkSbitMappingWithCalPulseLocal in src/calibration_routines.cpp
checkSbitRateWithCalPulseLocal in src/calibration_routines.cpp
dacScanLocal in src/calibration_routines.cpp
configureScanModuleLocal in src/optohybrid.cpp
getUltraScanResultsLocal in src/optohybrid.cpp
```

I could probably give a more useful set of guidelines, so I'll take a look at each of those individually and reply with some thoughts, but since you've been working closely with them, you probably already have a good idea of where these optimizations can happen.

AndrewLevin: @jsturdy, I replaced `vfatN`, `vfatMask`, `ohN`, and `ohMask` with just `vfat` and `oh`, and in `checkSbitMappingWithCalPulseLocal` and `checkSbitRateWithCalPulseLocal` the mask returned from `vfatSyncCheckLocal` is now used directly as the notmask, and the VFAT mask passed as an rpc parameter is now ignored.

bdorney: > @jsturdy, I replaced `vfatN`, `vfatMask`, `ohN`, and `ohMask` with just `vfat` and `oh`, and in `checkSbitMappingWithCalPulseLocal` and `checkSbitRateWithCalPulseLocal` the mask returned from `vfatSyncCheckLocal` is now used directly as the notmask, and the VFAT mask passed as an rpc parameter is now ignored.

This should specifically *not* be done.  If you use the `vfatmask` that is obtained from `vfatSyncCheckLocal()` and ignore the provided `vfatmask` by the user then you're not doing what the user is requesting.  Moreover you're forcing it to "work" in a situation where an error *should* be thrown: i.e. the user wants VFAT15 to be checked but VFAT15 is not sync'd so it is ignored...then the user wonders why all data for VFAT15 is returned as zero and starts to wonder why.  This causes the user/expert to chase a problem that is different from the real problem.  In this case it *should* have thrown an error but did not.

I need some more time to digest this and will report back.  But for sure what you've described above should *not* be done.  Revert it.

include/calibration_routines.h
```diff
@@ -101,7 +101,7 @@ void dacMonConfLocal(localArgs * la, uint32_t ohN, uint32_t ch);
  *  \param ohN Optical link
  *  \param enable See detailed mehod description
  */
-void ttcGenToggleLocal(localArgs * la, uint32_t ohN, bool enable);
+void ttcGenToggleLocal(localArgs * la, ParamScan *scanParams, ParamTTCGen *ttcParams);
```
bdorney: You're guilty of the catch-all problem @jsturdy and I are trying to ward against ([discussion](
https://github.com/cms-gem-daq-project/ctp7_modules/pull/63/files#r242623163).  This function has no need of the other parameters in `ParamScan` nor `ParamTTCGen` and three simple arguments are al that is needed.  Revert.

include/calibration_routines.h
```diff
@@ -81,15 +81,15 @@ void applyChanMask(std::unordered_map<uint32_t, uint32_t> map_chanOrigMask, loca
...
+bool confCalPulseLocal(localArgs *la, ParamCalPulse *calParams, ParamScan *scanParams);
 
 /*! \fn void dacMonConfLocal(localArgs * la, uint32_t ohN, uint32_t ch)
  *  \brief Configures DAQ monitor. Local version only
  *  \param la Local arguments structure
  *  \param ohN Optical link number
  *  \param ch Channel of interest
  */
-void dacMonConfLocal(localArgs * la, uint32_t ohN, uint32_t ch);
+void dacMonConfLocal(localArgs * la, ParamScan *scanParams);
```
bdorney: [same comment](
https://github.com/cms-gem-daq-project/ctp7_modules/pull/63/files#r246382903).

include/calibration_routines.h
```diff
@@ -81,15 +81,15 @@ void applyChanMask(std::unordered_map<uint32_t, uint32_t> map_chanOrigMask, loca
  *  \param currentPulse Selects whether to use current or volage pulse
  *  \param calScaleFactor Scale factor for the calibration pulse height (00 = 25%, 01 = 50%, 10 = 75%, 11 = 100%)
  */
-bool confCalPulseLocal(localArgs *la, uint32_t ohN, uint32_t mask, uint32_t ch, bool toggleOn, bool currentPulse, uint32_t calScaleFactor);
+bool confCalPulseLocal(localArgs *la, ParamCalPulse *calParams, ParamScan *scanParams);
```
bdorney, line 84: This really only needs `ParamCalPulse` as many parameters of `ParamScan` are not used here.  Here `ohN` could simply be `oh` if needed; and `mask` could remain as an argument.

include/calibration_routines.h
```diff
@@ -146,7 +146,7 @@ void ttcGenToggle(const RPCMsg *request, RPCMsg *response);
  *  \param nPulses Number of calibration pulses to generate
  *  \param enable If true (false) ignore (take) ttc commands from backplane for this AMC (affects all links)
  */
-void ttcGenConfLocal(localArgs * la, uint32_t ohN, uint32_t mode, uint32_t type, uint32_t pulseDelay, uint32_t L1Ainterval, uint32_t nPulses, bool enable);
+void ttcGenConfLocal(localArgs * la, ParamScan *scanParams, ParamTTCGen *ttcParams);
```
bdorney: Again you're trading a single argument `ohN` for a struct where you're only using *one* argument.  The use of `ParamTTCGen` is correct, however the use of `ParamScan` I think is overkill.  Recall don't be catch-all to our functions.

include/utils.h
bdorney: To follow-up on the point raised by @jsturdy we can indeed have further simplification.  Having one or more of these struct's serve as a "catch-all" for parameters defeats the purpose of implementing them.  The primary offender in this case appears to be (based on my own bad design) the `ParamScan` `struct`.  Indeed the terms:

 - `ohN` (`vfatN`), and
 - `ohMask` (`vfatmask`)

Can be compressed into two terms `oh` and `vfat` but then the function documentation needs to clearly state what is expected, e.g. a single integer, or an N-bit mask.  Furthermore we can remove from `ParamScan`:

 - `dacSelect` (only used by `dacScanLocal()` and friends), and
 - `useExtTrig` (only used by `genScanLocal()` and friends).

These two parameters can just be regular function arguments and removed from the `ParamScan` struct.

For the case of:

- `checkSbitMappingWithCalPulseLocal()`, and
- `checkSbitRateWithCalPulseLocal()`

Their functionality should not be changed.  And this can be easily accommodated by instead of taking `mask` from `vfatSyncCheckLocal()` taking it as an additional input argument.  This preserves the intended use case.  

Finally, don't trade "up" on complexity, for example there's no reason to replace one single integer with a struct that carries that integer, see discussion [here](https://github.com/cms-gem-daq-project/ctp7_modules/pull/63/files#r246382903).

bdorney (changes requested):
Don't try to use the struct's as "catch-all's" additionally don't trade up in complexity by passing a struct argument when a simple object would do.

include/utils.h
AndrewLevin: done

include/calibration_routines.h
AndrewLevin: done

include/calibration_routines.h
AndrewLevin: done

include/calibration_routines.h
AndrewLevin: it is not clear if you saw the `ch` argument, which is also part of ParamScan, can you confirm?

include/calibration_routines.h
AndrewLevin: done

bdorney:
Not sure if the comments @jsturdy and I left have been addressed.  But this would need to be `rebase`'d off the current `HEAD` of `develop`.

AndrewLevin:
Yes, I have addressed the comments from you and @jsturdy. I just rebased off the HEAD of develop.

bdorney:
Can you mark this as WIP until you’ve sufficiently tested all the new changes; or provide a git diff between the current HEAD and the commit where this was tested last.

AndrewLevin:
@bdorney, here is the diff between the HEAD (7341bbe5ed6798a53c9a653726297873f7db7a7b) and the commit I last tested on the ctp7 (fb8b2a4a38e5fb25d60e15f0bf9e8997b54af633):

https://github.com/cms-gem-daq-project/ctp7_modules/pull/63/files/7dfedd414a70ae0abb54da8f92d03479f7621f8c..7341bbe5ed6798a53c9a653726297873f7db7a7b

I have added [WIP]. Please let me know when I should retest it.

bdorney:
> @bdorney, here is the diff between the HEAD ([7341bbe](https://github.com/cms-gem-daq-project/ctp7_modules/commit/7341bbe5ed6798a53c9a653726297873f7db7a7b)) and the commit I last tested on the ctp7 ([fb8b2a4](https://github.com/cms-gem-daq-project/ctp7_modules/commit/fb8b2a4a38e5fb25d60e15f0bf9e8997b54af633)):
> ...
> ...
> I have added [WIP]. Please let me know when I should retest it.

In the future something more like [this](https://github.com/cms-gem-daq-project/ctp7_modules/pull/63/files/7dfedd414a70ae0abb54da8f92d03479f7621f8c..7341bbe5ed6798a53c9a653726297873f7db7a7b) is more useful.  Keep in mind the web GUI can do this.

include/calibration_routines.h
bdorney: It seems this was not addressed or the commit has not been pushed.

include/calibration_routines.h
bdorney: same

include/utils.h
```diff
@@ -55,71 +55,66 @@ struct ParamCalPulse{
  */
 struct ParamScan{
     //Hardware selection
-    uint32_t ohN;       //optical link number
-    uint32_t ohMask;    //OH mask
-    uint32_t vfatN;     //VFAT number
-    uint32_t vfatMask;  //VFAT mask
-    uint32_t chan;      //channel of interest
+    uint32_t oh;       //optical link number or optical link mask
```
bdorney: optical link is a misnomer.  One OH has 3 bidirectional optical links and 2 unidirectional links.  Update the comment to reflect what this actually is:  OH index or OH Mask

bdorney: Additionally the documentation written in the `doxygen` comment for this struct is woefully inadequate.

include/utils.h
```diff
@@ -55,71 +55,66 @@ struct ParamCalPulse{
...
-        dacStep=1;
+        max=254;
+        min=0;
+        step=1;
         nevts=100;
     }
 }; //End ParamScan
 
-/*! \struct ParamTtcGen
+/*! \struct ParamTTCGen
```
bdorney, line 86: Insufficient `doxygen` documentation.

src/calibration_routines.cpp
```diff
@@ -692,12 +681,12 @@ void sbitRateScanLocal(localArgs *la, uint32_t *outDataDacVal, uint32_t *outData
 
 void sbitRateScanParallelLocal(localArgs *la, uint32_t *outDataDacVal, uint32_t *outDataTrigRatePerVFAT, uint32_t *outDataTrigRateOverall,ParamScan *scanParams)
 {
-    uint32_t ohN = scanParams->ohN;
-    uint32_t vfatmask = scanParams->ohN;
+    uint32_t ohN = scanParams->oh;
+    uint32_t vfatmask = scanParams->vfat;
```
bdorney: unnecessary white space.

src/calibration_routines.cpp
```diff
@@ -712,7 +701,7 @@ void sbitRateScanParallelLocal(localArgs *la, uint32_t *outDataDacVal, uint32_t
                 la->response->set_string("error",regBuf);
                 return;
             }
-
+
```
bdorney: unnecessary tab characters.  You should setup your editor to strip this automatically.

src/calibration_routines.cpp
```diff
@@ -970,12 +957,11 @@ void checkSbitMappingWithCalPulseLocal(localArgs *la, uint32_t *outData, ParamCa
             } //End Loop over clusters
         } //End Pulses for this channel
 
-        ParamCalPulse calParams_modified(*calParams);
-        
-        calParams_modified.enable = false;
+        ParamCalPulse calParams_disabled(*calParams);
+        calParams_disabled.enable = false;
```
bdorney: By calling the constructor in this manner you're supplying a pointer.  So now `calParams_disabled` points to `calParams`.  Are you sure it's making a [deep copy and not a shallow copy](https://www.learncpp.com/cpp-tutorial/915-shallow-vs-deep-copying/)?  I suspect not.

For example, in L961 when you set `calParams_disabled.enable` to `false` I suspect this also changes  `calParams.enable`.

bdorney: Same comment at line 910.

src/calibration_routines.cpp
AndrewLevin: fixed

src/calibration_routines.cpp
AndrewLevin: fixed

include/calibration_routines.h
AndrewLevin: Was only modified in the .cpp files, sorry.

include/calibration_routines.h
AndrewLevin: Was only modified in the .cpp file, sorry.

include/utils.h
AndrewLevin: I changed the `//` comments to `/*!<`...`*/` comments.

include/utils.h
AndrewLevin: I changed the `//` comments to `/*!<`...`*/` comments.

AndrewLevin:
I am using a dereferenced pointer, and the structs do not have any member variables that are pointers, so I am doing a full copy of the data.
________________________________

AndrewLevin:
Same comment.
________________________________

include/utils.h
bdorney: Can you build the documentation and show a screen shot of how this will look?  I'm not sure this is the intended way `doxygen` should be used.

AndrewLevin: Yes:

screenshot1
screenshot2
screenshot3

jsturdy:
Will be revisited

Issue #65, opened 2018-11-09
[feature request] Phase shifting module


jsturdy:
## Brief summary of issue
To remove the phase shifting from the `cmsgemos` implementation, and away from a script living on the CTP7, this should be incorporated into a module (until it is fully implemented in the FW)

### Types of issue
- [x] Feature request (request for change which adds functionality)

## Possible Solution (for bugs)
Look at (`python`):
https://github.com/jsturdy/GEM_AMC/blob/fe3780048191b6a23ef04d1d430d3b53c5e8e8b8/scripts/test_phase_shifting.py#L99-L451 

for the current incarnation (and associated helper functions), which *should* be functionally identical to the `c++` implementation done here (with `uhal`):
https://github.com/jsturdy/cmsgemos/blob/1e8e6a90181eb7c4b4bf4cf495fa4f3437b81d4f/gemhardware/src/common/HwGenericAMC.cc#L549-L1019

bdorney:
I'll take this one on.  This is a purely AMC function correct?

jsturdy:
> This is a purely AMC function correct?

Yes, N.B., it **only** works with V2 FW (unless @evka85 has ported the updated TTC module to the V3)

bdorney:
Addressed by #69.

PR #69, opened 2018-11-13
New Feature: Phase Align Function

bdorney:
## Description

Addresses #65.  Migrated from `uhal` based methods in `cmsgemos`.

Provided local and nonlocal functions for `uhal` based equivalent functions in `cmsgemos`:

- `checkPllLock(...)`, and
- `ttcMMCMPhaseShift(...)`

The keys for the RPC request match input arguments.  No rpc response is returned by `ttcMMCMPhaseShift(...)` but for `checkPllLock(...)` the key `lockCnt` is inserted into the rpc response.

Note that `checkPllLock(...)` is provided for standalone functionality but `ttcMMCMPhaseShiftLocal(...)` calls `checkPllLockLocal(...)` preserving the functionality.  I am not sure if the uhal based `checkPllLock(...)` function is being used elsewhere in `cmsgemos` so I included a remote function just in case.

Also implemented the requested readback check in `ttcMMCMPhaseShiftLocal(...)`.  

Finally in the cases **where** exceptions **were** suggested in the comments I've set the `error` string in the RPC response.  The calling function from the DAQ machine should check for the presence of this key (e.g. `response->get_key_exists("error")`) and raise the appropriate exception or take the appropriate follow-up action.  Note that when the `"error"` key is set it definitely means the procedure *didn't* work or was not executed.

Not sure if additional keys should be inserted into the rpc response, e.g. in the case where `CMSGEMOS_WARN` logger was used one could thing that the `"warning"` key could be inserted into the rpc response but I'm not sure how `cmsgemos` is handling it...from the code snippet presented in #65 it seemed there was no exception being raised so if it's not necessary I'll just leave it as is.  

In all cases logging is now managed by logger in RPC land and will show up in the local/remote log file for the CTP7 (not the xdaq output which I suspect was previously being done). 

### Types of changes

## Motivation and Context

#65 

## How Has This Been Tested?

Not tested yet, would need `eagle34` in 904 to have a "good linux image" or would need some time during interfill period to try to test on `eagle61` in P5. @jsturdy 

## Checklist:

### DIFF
--- a/include/amc.h
+++ b/include/amc.h
@@ -11,20 +11,20 @@
 #include "utils.h"
 #include <unistd.h>
 
-/*! \fn uint32_t checkPllLockLocal(localArgs * la, uint32_t readAttempts)
+/*! \fn uint32_t checkPLLLockLocal(localArgs * la, uint32_t readAttempts)
  *  \brief Resets the PLL and checks if it relocks
  *  \param la Local arguments structure
  *  \param readAttempts Specifies the number of times to reset the PLL and check for a relock
  *  \return Returns the number of times the PLL relocked
  */
-uint32_t checkPllLockLocal(localArgs * la, int readAttempts);
+uint32_t checkPLLLockLocal(localArgs * la, int readAttempts);
 
-/*! \fn void checkPllLock(const RPCMsg *request, RPCMsg *response)
+/*! \fn void checkPLLLock(const RPCMsg *request, RPCMsg *response)
  *  \brief Resets the PLL and checks if it relocks, see local method for details
  *  \param request RPC request message
  *  \param response RPC response message
  */
-void checkPllLock(const RPCMsg *request, RPCMsg *response);
+void checkPLLLock(const RPCMsg *request, RPCMsg *response);
 
 /*! \fn unsigned int fw_version_check(const char* caller_name, localArgs *la)
  *  \brief Returns AMC FW version
--- a/include/amc.h
+++ b/include/amc.h
@@ -11,6 +11,21 @@
 #include "utils.h"
 #include <unistd.h>
 
+/*! \fn uint32_t checkPllLockLocal(localArgs * la, uint32_t readAttempts)
+ *  \brief Resets the PLL and checks if it relocks
+ *  \param la Local arguments structure
+ *  \param readAttempts Specifies the number of times to reset the PLL and check for a relock
+ *  \return Returns the number of times the PLL relocked
+ */
+uint32_t checkPllLockLocal(localArgs * la, int readAttempts);
+
+/*! \fn void checkPllLock(const RPCMsg *request, RPCMsg *response)
+ *  \brief Resets the PLL and checks if it relocks, see local method for details
+ *  \param request RPC request message
+ *  \param response RPC response message
+ */
+void checkPllLock(const RPCMsg *request, RPCMsg *response);
+
 /*! \fn unsigned int fw_version_check(const char* caller_name, localArgs *la)
  *  \brief Returns AMC FW version
  *  in case FW version is not 1.X or 3.X sets an error string in response
@@ -56,11 +71,29 @@ void getOHVFATMaskMultiLink(const RPCMsg *request, RPCMsg *response);
  */
 std::vector<uint32_t> sbitReadOutLocal(localArgs *la, uint32_t ohN, uint32_t acquireTime, bool *maxNetworkSizeReached);
 
-/*! \fn sbitReadOut(const RPCMsg *request, RPCMsg *response)
+/*! \fn void sbitReadOut(const RPCMsg *request, RPCMsg *response)
  *  \brief readout sbits using the SBIT Monitor.  See the local callable methods documentation for details.
  *  \param request RPC response message
  *  \param response RPC response message
  */
 void sbitReadOut(const RPCMsg *request, RPCMsg *response);
+
+/*! \fn void ttcMMCMPhaseShiftLocal(localArgs *la, bool shiftOutOfLockFirst, bool useBC0Locked, bool doScan)
+ *  \brief Performs the locking procedure performs up to 3840 shifts (full width of one good + bad region).
+ *  \details If shiftOutOfLockFirst is true, then the procedure will first shift into a bad region (not an edge), find the next good lock status, shift halfway through the region.  1920 for BC0_LOCKED, 1000 for PLL_LOCKED.
+ *  \details If shiftOutOfLockFirst is false, then the procedure will find X consecutive "good" locks for X = 200 (50) for BC0_LOCKED (PLL_LOCKED).  It will then reverse direction and shift backwards halfway.  If a bad lock is encountered it will reset and try again.  Otherwise it will take the phase at the back half point
+ *  \param la Local arguments structure
+ *  \param shiftOutOfLockFirst controls whether the procedure will force a relock
+ *  \useBC0Locked controls whether the procedure will use the BC0_LOCKED or the PLL_LOCKED register.  Note for GEM_AMC FW > 1.13.0 BC0_LOCKED doesn't work.
+ *  \param doScan tells the procedure to run through the full possibility of phases several times, and just logs the place where it found a lock
+ */
+void ttcMMCMPhaseShiftLocal(localArgs *la, bool shiftOutOfLockFirst, bool useBC0Locked, bool doScan);
+
+/*! \fn void ttcMMCMPhaseShift(const RPCMsg *request, RPCMsg *response)
+ *  \brief phase alignment algorithm for ttc phase on uTCA backplane.  See the local callable methods documentation for details.
+ *  \param request RPC response message
+ *  \param response RPC response message
+ */
+void ttcMMCMPhaseShift(const RPCMsg *request, RPCMsg *response);
 
 #endif
--- a/src/amc.cpp
+++ b/src/amc.cpp
@@ -267,18 +267,18 @@ void ttcMMCMPhaseShiftLocal(localArgs *la, bool shiftOutOfLockFirst, bool useBC0
     std::string strTTCCtrlBaseNode = "GEM_AMC.TTC.CTRL.";
     std::vector<std::pair<std::string, uint32_t> > vec_ttcCtrlRegs;
 
-    vec_ttcCtrlRegs["DISABLE_PHASE_ALIGNMENT"]       = 0x1;
-    vec_ttcCtrlRegs["PA_DISABLE_GTH_PHASE_TRACKING"] = 0x1;
-    vec_ttcCtrlRegs["PA_MANUAL_OVERRIDE"]            = 0x1;
-    vec_ttcCtrlRegs["PA_MANUAL_SHIFT_DIR"]           = 0x1;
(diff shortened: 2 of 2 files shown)


include/amc.h
```diff
@@ -11,6 +11,21 @@
 #include "utils.h"
 #include <unistd.h>
 
+/*! \fn uint32_t checkPllLockLocal(localArgs * la, uint32_t readAttempts)
+ *  \brief Resets the PLL and checks if it relocks
+ *  \param la Local arguments structure
+ *  \param readAttempts Specifies the number of times to reset the PLL and check for a relock
+ *  \return Returns the number of times the PLL relocked
+ */
+uint32_t checkPllLockLocal(localArgs * la, int readAttempts);
```
jsturdy: Don't know if this was a direct copy (and in `cmsgemos` it might have been a direct copy from @evka85's original script), but I strongly prefer capitalization that correlates with actuality, e.g., `Pll` -> `PLL`, `Ttc` -> `TTC` (it's already done for `MMCM`, so this is more of a consistency request).
If these types of indicators occur in the middle of variables/function names, they should be all caps, if they are at the beginning, then they should be all lower case.

src/amc.cpp
```diff
@@ -222,6 +258,512 @@ void sbitReadOut(const RPCMsg *request, RPCMsg *response){
     return;
 } //End sbitReadOut()
 
+void ttcMMCMPhaseShiftLocal(localArgs *la, bool shiftOutOfLockFirst, bool useBC0Locked, bool doScan){
+    const int PLL_LOCK_READ_ATTEMPTS  = 10;
+
+    LOGGER->log_message(LogManager::INFO,"starting phase shifting procedure");
+
+    std::string strTtcCtrlBaseNode = "GEM_AMC.TTC.CTRL.";
+    std::unordered_map<std::string, uint32_t> map_ttcCtrlRegs;
```
jsturdy: If `unordered` means that things don't happen in a given order, this is not the intention, the sequence of `write`s  should be understood to be as given

bdorney: Got it; I'll update...for sure this won't preserve the sequence illustrated.

include/amc.h
bdorney: I was following camel case; I'll update.

src/amc.cpp
bdorney: It's now changed to a `std::vector<std::pair<std::string, uint32_t> >` object.  This will preserve the ordering.

src/amc.cpp
```diff
@@ -222,6 +259,512 @@ void sbitReadOut(const RPCMsg *request, RPCMsg *response){
...
+        if (reversingForLock){
+            totalShiftCount -= 1;
+        }
+        else{
+            totalShiftCount += 1;
+        }
+    }
+
+    if (bestLockFound) {
+        writeReg(la,"GEM_AMC.TTC.CTRL.MMCM_RESET", 0x1);
```
jsturdy, line 733: This was a bug in the `cmsgemos` implementation, it should be a `MODULE_RESET`

src/amc.cpp
```diff
@@ -222,6 +259,512 @@ void sbitReadOut(const RPCMsg *request, RPCMsg *response){
...
+    vec_ttcCtrlRegs.push_back(std::make_pair("PA_DISABLE_GTH_PHASE_TRACKING" , 0x1));
+    vec_ttcCtrlRegs.push_back(std::make_pair("PA_MANUAL_OVERRIDE"            , 0x1));
+    vec_ttcCtrlRegs.push_back(std::make_pair("PA_MANUAL_SHIFT_DIR"           , 0x1));
+    vec_ttcCtrlRegs.push_back(std::make_pair("PA_GTH_MANUAL_OVERRIDE"        , 0x1));
+    vec_ttcCtrlRegs.push_back(std::make_pair("PA_GTH_MANUAL_SHIFT_DIR"       , 0x0));
+    vec_ttcCtrlRegs.push_back(std::make_pair("PA_GTH_MANUAL_SHIFT_STEP"      , 0x1));
+    vec_ttcCtrlRegs.push_back(std::make_pair("PA_GTH_MANUAL_SEL_OVERRIDE"    , 0x1));
+    vec_ttcCtrlRegs.push_back(std::make_pair("PA_GTH_MANUAL_COMBINED"        , 0x1));
+    vec_ttcCtrlRegs.push_back(std::make_pair("GTH_TXDLYBYPASS"               , 0x1));
+    vec_ttcCtrlRegs.push_back(std::make_pair("PA_MANUAL_PLL_RESET"           , 0x1));
```
jsturdy, line 280: These resets are strobed, so they shouldn't be read back (the readback below could probably be optional altogether)

src/amc.cpp
```diff
@@ -222,6 +259,512 @@ void sbitReadOut(const RPCMsg *request, RPCMsg *response){
...
+    vec_ttcCtrlRegs.push_back(std::make_pair("CNT_RESET"                     , 0x1));
+
+    // write & readback of aforementioned registers
+    uint32_t readback;
+    for(auto ttcRegIter = vec_ttcCtrlRegs.begin(); ttcRegIter != vec_ttcCtrlRegs.end(); ++ttcRegIter){
```
jsturdy, line 285: Can simplify this block a bit with range based `for` loop syntax, e.g.,
```c++
for (auto &ttcRegIter: vec_ttcCtrlRegs) {
  writeReg(la, strTTCCtrlBaseNode + ttcRegIter.first, ttcRegIter.second);
  std::this_thread::sleep_for(std::chrono::microseconds(250));
  if ((ttcRegIter.first).rfind("RESET") != std::string::npos)
    continue;
...
}
```

src/amc.cpp
```diff
@@ -222,6 +259,512 @@ void sbitReadOut(const RPCMsg *request, RPCMsg *response){
...
+                             false, false, false, false, true, false, false,
+                             false, false, false, true, false, false};
+
+    int nGoodLocks       = 0;
+    int nShiftsSinceLock = 0;
+    int nBadLocks        = 0;
+    int totalShiftCount  = 0;
+
+    for (int i = 0; i < maxShift; ++i) {
+        writeReg(la, strTTCCtrlBaseNode + "CNT_RESET", 0x1);
```
jsturdy, line 347: This should also be removed (was an artifact in the `cmsgemos` implementation from an older version, it has some unintended consequences in the logic)

src/amc.cpp
```diff
@@ -222,6 +259,512 @@ void sbitReadOut(const RPCMsg *request, RPCMsg *response){
...
+        writeReg(la, strTTCCtrlBaseNode + "PA_GTH_MANUAL_SHIFT_EN", 0x1);
+
+        if (!reversingForLock && (gthShiftCnt == 39)) {
```
jsturdy, line 350: before this block (right after the current L348) should have the following (again, wasn't in the `cmsgemos` version, but was in the `python` version
```c++
uint32_t tmpGthShiftCnt  = readReg(la,"GEM_AMC.TTC.STATUS.CLK.PA_MANUAL_GTH_SHIFT_CNT");
uint32_t tmpMmcmShiftCnt = readReg(la,"GEM_AMC.TTC.STATUS.CLK.PA_MANUAL_SHIFT_CNT");

bool mmcmShiftRequired = false;
if (reversingForLock)
  mmcmShiftRequired = mmcmShiftTable[gthShiftCnt];
else
  mmcmShiftRequired = mmcmShiftTable[gthShiftCnt+1];
```

src/amc.cpp
```diff
@@ -222,6 +259,512 @@ void sbitReadOut(const RPCMsg *request, RPCMsg *response){
...
+                    " match the expected value."
+                    " Expected shift cnt = " + std::to_string(gthShiftCnt) +
+                    ", ctp7 returned "       + std::to_string(tmpGthShiftCnt));
+
+            writeReg(la, "GEM_AMC.TTC.CTRL.PA_GTH_MANUAL_SHIFT_EN", 0x1);
+            tmpGthShiftCnt = readReg(la,"GEM_AMC.TTC.STATUS.CLK.PA_MANUAL_GTH_SHIFT_CNT");
+            //FIX ME should this continue indefinitely...?
+        }
+
+        if (mmcmShiftTable[gthShiftCnt+1]) {
```
jsturdy, line 382: With the change suggested at L350, this would become:
```c++
if (mmcmShiftRequired) {
```

src/amc.cpp
```diff
@@ -222,6 +259,512 @@ void sbitReadOut(const RPCMsg *request, RPCMsg *response){
...
+            }
+            else {
+                if (reversingForLock) {
+                    mmcmShiftCnt -= 1;
+                } else {
+                    mmcmShiftCnt += 1;
+                }
+            }
+
+            tmpMmcmShiftCnt = readReg(la,"TTC.STATUS.CLK.PA_MANUAL_SHIFT_CNT");
```
jsturdy, line 397: This (and a couple of other places) are missing the `GEM_AMC` base

jsturdy:
Currently, even with the changes mentioned thus far, the procedure is working for a while, and then getting stuck in a funky state (seems it finds 1 good lock, and then just gets "bad" lock status for 1000 straight shifts).  The `python` CTP7 local version is working, so I hope tracking down the logic difference is the reason (so far, a line-by-line comparison hasn't yielded any fruit...)

src/amc.cpp
```diff
@@ -222,6 +259,512 @@ void sbitReadOut(const RPCMsg *request, RPCMsg *response){
...
+                    if (nBadLocks > 500) {
+                      firstUnlockFound = true;
+                      LOGGER->log_message(LogManager::DEBUG,
+                              "500 unlocks found after " + std::to_string(i+1) + " shifts:"
+                              " bad locks "              + std::to_string(nBadLocks) +
+                              ", good locks "            + std::to_string(nGoodLocks) +
+                              ", mmcm phase count = "    + std::to_string(phase) +
+                              ", mmcm phase ns = "       + std::to_string(phaseNs) + "ns");
+                    }
+                    else {
```
jsturdy, line 552: Found it (another bug in the `cmsgemos` implementation, but not in the `python` version)!
This `else` block (L552 up to L592/3) belongs to `if (shiftOutOfLockFirst)`.
After closing it, the `if (!firstUnlockFound)` should be closed on L594 leading into the `else { // shift to first good PLL locked`

jsturdy (changes requested):
I'll create a PR to your branch with all the things I modified, this latest one is the biggest, the rest were little things here and there (and some formatting/whitespace as I was tracking backwards to the source of the issue)
If you don't have time/don't feel like merging that PR in, let me know, and I'll supersede this PR instead

bdorney:
@jsturdy I merged your PR into my repo...is any new testing required or where the commits you pushed to my repo sufficient (and tested)?

jsturdy (dismissed):
I don't think there was anything else outstanding, but I didn't rigorously test it

jsturdy:
Though, if you could `rebase` off of `cms-gem-daq-project/develop`, I think that would be better than pressing the "Update branch" button

bdorney:
> Though, if you could `rebase` off of `cms-gem-daq-project/develop`, I think that would be better than pressing the "Update branch" button

Rebased.

jsturdy (approved):

PR #71, opened 2018-11-15
Adding CFG_CAL_DAC to DAC Scans & Resolving Array Indexing Error

bdorney:
## Description

Added `CFG_CAL_DAC` to DAC Scans.

Created header file for `calibration_routines.h` to be in line with other modules. 

Finally fixed a bug in `dacScanMultiLink(...)` where if an OH was masked the RPC word array that would be stored in the RPC response would "jump" a link and it would not be possible to figure out from index arguments which OH was stored where.

### Types of changes

## Motivation and Context

`CFG_CAL_DAC` was not enabled in list of possible DAC scans. Corrected.

Additionally output word array wouldn't have correct indexing if an OH was skipped. 

## How Has This Been Tested?

http://cmsonline.cern.ch/cms-elog/1073862

## Checklist:


jsturdy (dismissed):
It's a bit hard to find the part of this PR that corresponds to the title, within the restructuring

jsturdy:
Actually, looking deeper I find a commit that needs a fix.
You removed default arguments from the function declaration, and left them in the function definition.
Best practice is to put default values in the function declaration, not the function definition.
Then anyone reading the API (header) would not know about the default arguments if it is placed in the definition.

jsturdy (approved):
Thanks!

Issue #72, opened 2018-11-26
Move the GBT tools from reg_utils python scripts to RPC module


lpetre-ulb:
## Brief summary of issue

As described in [this issue](https://github.com/cms-gem-daq-project/reg_utils/issues/41), the `gbt.py` tool cannot be called on the CTP7 with the 12 OH's address table, because of out-of-memory error.

The best solution would be to implement the features of the [`gbt.py` tool](https://github.com/cms-gem-daq-project/reg_utils/blob/master/python/reg_interface/scripts/gbt.py) within RPC functions.

The required features are 

* Program the GBTX from config file.
* Perform a phase on the GBTX.
* Manually set the phase of one e-link.

### Types of issue

- [ ] Bug report (report an issue with the code)
- [x] Feature request (request for change which adds functionality)

## Expected Behavior

The functions currently supported by the `gbt.py` python tool must be ported to RPC modules.

## Current Behavior

The GBT tool is currently a python script and not a RPC module.

## Context (for feature requests)

The `gbt.py` tool cannot be called on the CTP7 with the 12 OH's address table because of out-of-memory errors.

bdorney:
So this should be compartmentalised into a series of functions.

I see the following functions being necessary:

- `writeGBTRegLocal(LocalArgs *la, uint32_t ohN, uint32_t gbtN, ...)`: For OHX this should write a GBTX register with value Y.
- `writeGBTConfigLocal(....)` this should apply a given GBT config to the GBTx on OHX
    - This should have a local and a non-local function
    - I think the non-local function should have an RPC key for binary data that is the GBT config, this can be passed to this function at the DAQ machine side either as an input file or a DB read.
   - The local version of this function obviously calls `writeGBTRegLocal(...)`
- `scanGBTPhaseLocal(...)` this should ask to scan the phase on OHX GBTY and it should take as input a parameter which specifies the number of phase scans that should be performed
   - The local version of this function obviously calls `writeGBTRegLocal(...)`
   - There should be a local and non-local version of this function
   - This could also have a multi link non-local function (see for example `dacScanMultiLink(...)`.
   - For each GBT e-link (VFAT Pos) this should return in the RPC response a key which stores a word array where the array position corresponds to e-link phase and the array contents is the number of times that phase was "good."
- `writeGBTVFATPhaseLocal(...)`: this should write the phase of a given VFAT position to be X
   - The local version of this function obviously calls `writeGBTRegLocal(...)`
[...]

PR #73, opened 2018-11-28
Implement RPC methods for configuring the GBTX's

lpetre-ulb:
## Description

Addresses #72 

This PR adds a new RPC module which implements 6 new methods related to the GBTX configuration :

* `writeGBTConfig` updates the configuration of one chosen GBTX.
* `writeGBTConfigMultiLink`, multi-link version of the previous method.
* `writeGBTPhase` updates the RX phases of one OH.
* `writeGBTPhaseMultiLink`, multi-link version of the previous method.
* `scanGBTPhase` scans the phases on one OH.
* `scanGBTPhaseMultiLink`, multi-link version of the previous method.

### Types of changes

## Motivation and Context

This PR fixes issue #72.

## How Has This Been Tested?

These tests were performed to see if the functionalities implemented in the `gbt.py` were still working with the RPC modules and so the new CTP7 releases :

* Writing the GBTX configurations. It has been tested for different OH, multiple GBTX's at a time.
* Setting the GBTX RX phases for 1 VFAT at a time.
* Setting the GBTX RX phases for multiples VFAT's at a time.
* Scanning the GBTX RX phases for one OH.

More details are available in the `reg_utils` PR : https://github.com/cms-gem-daq-project/reg_utils/pull/42

The environment was the following :
* CTP7 : eagle63
* CTP7 firmware : 3.7.0
* CTP7 build-id : CTP7-GENERIC-20180529T153916-0500-4935611
* `xHAL` commit https://github.com/cms-gem-daq-project/xhal/pull/99/commits/59d788dea1cf0dfb1888c39eb9744f3e905abdcb
* `cmsgemos` commit https://github.com/cms-gem-daq-project/cmsgemos/pull/226/commits/66639b6672b2a9fdcdec2a02c67b93e9904a759f
* The new `gbt.py` used for these tests comes from commit https://github.com/cms-gem-daq-project/reg_utils/pull/42/commits/c4ad5f3afe2ba3236538c1336c37d081f015af68

## Checklist:


jsturdy (commented):
My first suggestion would be to put these `GBT` related functions into a separate module, e.g., `gbt.cpp/h`.
Sure, the GBTs physically live only on an OptoHybrid, but these modules don't do any sort of object oriented encapsulation so I don't think there's any real reason to group them with "OptoHybrid" functions, and "scoping" the functionality at the file level is probably easier to organize

mexanick:
Actually, the modules themselves provide a sort of encapsulation as in order to perform a call the the caller should load an appropriate module and the call its method. I agree with @jsturdy that functional module separation should take place here for the better code organization.

lpetre-ulb:
Ok, thank you for the suggestion, I've moved the methods to a new RPC module and new files. I've also added all the "multi-link" variants (see the updated first post).

bdorney:
What's the status of this PR? This is a mission critical action and I was hoping that we would have made more progress on it.

lpetre-ulb:
I've performed the tests yesterday afternoon. Writing the configuration, setting and scanning the phases work as expected. I'm committing last changes before review. I'm also porting the `gbt.py` tool in `reg_utils` for using the new RPC methods.

lpetre-ulb:
I've updated the first post to keep it up-to-date with the status of the PR. The only missing part is the new `gbt.py` tool of which its PR should be ready before the end of the day.

lpetre-ulb:
With the last [PR for the `gbt.py` tool](https://github.com/cms-gem-daq-project/reg_utils/pull/42), the PR's are made in all four repositories. The GBT tool functionalities are working from the DAQ machine with the version 3.7.0 of the CTP7 firmware.

include/gbt.h
```diff
@@ -0,0 +1,90 @@
...
+ *  \param la Local arguments structure
+ *  \param ohN Optohybrid optical link number
+ *  \param phaseMin Lowest phase to scan (min = 0)
+ *  \param phaseMax Highest phase to scan (max = 15)
+ *  \param phaseStep Step to scan the phases
+ */
+std::vector<uint32_t> scanGBTPhaseLocal(localArgs *la, const uint32_t ohN, const uint8_t phaseMin, const uint8_t phaseMax, const uint8_t phaseStep);
+
+/*! \fn void scanGBTPhase(const RPCMsg *request, RPCMsg *response)
+ *  \brief Scan the GBT phases of one optical link
```
bdorney, line 21: What is `optical link` here?  e.g. does this only scan `GBT2` of `OHX`?  Or does it scan for *all* `GBT`'s on `OHX`?  Please update documentation to improve clarity.

include/gbt.h
```diff
@@ -0,0 +1,90 @@
...
+ *  \brief Local callable version of writeGBTXPhase
+ *  \param la Local arguments structure
+ *  \param ohN Optohybrid optical link number
+ *  \param vfatN VFAT index of which the phase is changed
+ *  \param config Phase to write
+ */
+void writeGBTPhaseLocal(localArgs *la, const uint32_t ohN, const uint32_t vfatN, const uint8_t phase);
+
+/*! \fn void writeGBTPhase(const RPCMsg *request, RPCMsg *response)
+ *  \brief Write the configuration of one GBTX
```
bdorney, line 67: Does this write the *full* configuration? e.g. all e-link phases on this `GBTx`? Or does it write just a single phase? From the functional arguments I assume the single phase case.  Please update the documentation to resolve this ambiguity.

src/gbt.cpp
```diff
@@ -0,0 +1,381 @@
...
+    static constexpr std::array<uint32_t, 24> vfatToGBTX{
+        { 1, 1, 1, 1, 1, 1, 1, 0, 1, 2, 2, 2, 0, 0, 0, 0, 1, 2, 2, 2, 2, 2, 2, 0 }
+    };
+    static constexpr std::array<uint8_t, 24> vfatToElink{
+        { 5, 9, 2, 3, 1, 8, 6, 6, 4, 1, 5, 4, 3, 2, 1, 0, 7, 8, 6, 7, 2, 3, 9, 8 }
+    };
+    static constexpr std::array<std::array<uint16_t, 3>, 10> elinkToRegisters{
+        { {69, 73, 77}, {67, 71, 75}, {93, 97, 101}, {91, 95, 99}, {117, 121, 125}, {115, 119, 123}, {141, 145, 149}, {139, 143, 147}, {165, 169, 173}, {163, 167, 171} }
+    };
+}
```
bdorney, line 18: These are specific to GE1/1 but we will have soon (Q1 2019) GE2/1 OH's that will have different hardware link assignments.  How do we intend to handle this?

One approach would be to use different namespaces for the different hardware.  Or a different naming convention.  I think this should be finalized *before* this PR is merged.

Beyond the scope of this PR I would like to see a header file that has all hard-coded hardware specific parameters in a single location; I suspect some refactoring would be needed to achieve this.

src/gbt.cpp
```diff
@@ -0,0 +1,381 @@
...
+
+void writeGBTRegLocal(localArgs *la, const uint32_t ohN, const uint32_t gbtN, const uint16_t address, const uint8_t value){
```
bdorney, line 20: In the header file this is the last function you declare but in this file it's the first definition you provide.  Perhaps for maintainability this should follow the order in which it's declared.

src/gbt.cpp
bdorney: @jsturdy @mexanick

src/gbt.cpp
```diff
@@ -0,0 +1,381 @@
...
+    writeReg(la, "GEM_AMC.SLOW_CONTROL.IC.EXECUTE_WRITE", 1);
+
+    return;
+} //End writeGBTRegLocal(...)
+
+void writeGBTConfigLocal(localArgs *la, const uint32_t ohN, const uint32_t gbtN, const std::vector<uint8_t> &config){
+    LOGGER->log_message(LogManager::INFO, stdsprintf("Writing the configuration of OH #%u - GBTX #%u.", ohN, gbtN));
+
+    for (size_t address = 0; address < config.size(); address++){
+        writeGBTRegLocal(la, ohN, gbtN, static_cast<uint16_t>(address), config[address]);
```
bdorney, line 47: You're assuming that `config` matches the expected size and order (e.g. index 0 is address 0) without:

1. actually checking that the size meets expectation, or
2. including a statement in the function documentation stating how the elements of `config` should be ordered and what it's size is expected to be (e.g. less than 366).

src/gbt.cpp
```diff
@@ -0,0 +1,381 @@
...
+        LOGGER->log_message(LogManager::ERROR, stdsprintf("The gbtN parameter supplied (%u) exceeds the number of GBTX's per OH (%u).", gbtN, gbtMax));
+        response->set_string("error", stdsprintf("The gbtN parameter supplied (%u) exceeds the number of GBTX's per OH (%u).", gbtN, gbtMax));
+        return;
+    }
+
+    // config key
+    if (request->get_binarydata_size("config") != 366){
+        LOGGER->log_message(LogManager::ERROR, "The configuration has not the correct size. It must be 366 8-bits registers long.");
+        response->set_string("error", "The configuration has not the correct size. It must be 366 8-bits registers long.");
+        return;
```
bdorney, line 86: This information is *not* contained in the function documentation in the header file.  This should be addressed.

src/gbt.cpp
```diff
@@ -0,0 +1,381 @@
...
+    }
+
+    // gbtN key
+    const uint32_t gbtN = request->get_word("gbtN");
+    const uint32_t gbtMax = 3;
+    if (gbtN > gbtMax){
+        LOGGER->log_message(LogManager::ERROR, stdsprintf("The gbtN parameter supplied (%u) exceeds the number of GBTX's per OH (%u).", gbtN, gbtMax));
+        response->set_string("error", stdsprintf("The gbtN parameter supplied (%u) exceeds the number of GBTX's per OH (%u).", gbtN, gbtMax));
+        return;
+    }
```
bdorney, line 80: This information is *not* contained in the function documentation in the header file. This should be addressed.

src/gbt.cpp
```diff
@@ -0,0 +1,381 @@
...
+    const std::bitset<32> ohMask(request->get_word("ohMask"));
+    const size_t ohUnmaskedNumber = ohMask.count();
+    const uint32_t ohMax = readReg(&la, "GEM_AMC.GEM_SYSTEM.CONFIG.NUM_OF_OH");
+
+    // gbtMasks key
+    if (request->get_word_array_size("gbtMasks") != ohUnmaskedNumber){
+        LOGGER->log_message(LogManager::ERROR, stdsprintf("The provided number of gbtMask's does not fullfill the expectations (%u).", ohUnmaskedNumber));
+        response->set_string("error", stdsprintf("The provided number of gbtMask's does not fullfill the expectations (%u).", ohUnmaskedNumber));
+        return;
+    }
```
bdorney, line 118: If you look at the structure of all functions the rpc request and response have keys that match the local callable functions; and in that case their documentation is provided by the local callable function.  But here you're using the keys `ohMask` and `gbtMasks` that are neither documented in this function documentation nor used by the local callable function.

Consider the `\details` field in `doxygen` to provide this information.  A developer/expert user should be able to understand what is needed for successful execution from looking at the function API and right now this is not possible.

src/gbt.cpp
```diff
@@ -0,0 +1,381 @@
...
+    std::vector<uint32_t> gbtMasks = request->get_word_array("gbtMasks");
+
+    // config key
+    const size_t expectedConfigSize = ohUnmaskedNumber*3*366;
+    if (request->get_binarydata_size("config") != expectedConfigSize){
+        LOGGER->log_message(LogManager::ERROR, stdsprintf("The configuration has not the correct size. This method expects %i 8-bits registers.", expectedConfigSize));
+        response->set_string("error", stdsprintf("The configuration has not the correct size. This method expects %i 8-bits registers.", expectedConfigSize));
+        return;
+    }
```
bdorney, line 127: Not included in function documentation.

src/gbt.cpp
```diff
@@ -0,0 +1,381 @@
...
+
+    // ohMask key
+    const std::bitset<32> ohMask(request->get_word("ohMask"));
+    const size_t ohUnmaskedNumber = ohMask.count();
+    const uint32_t ohMax = readReg(&la, "GEM_AMC.GEM_SYSTEM.CONFIG.NUM_OF_OH");
+
+    // gbtMasks key
+    if (request->get_word_array_size("gbtMasks") != ohUnmaskedNumber){
+        LOGGER->log_message(LogManager::ERROR, stdsprintf("The provided number of gbtMask's does not fullfill the expectations (%u).", ohUnmaskedNumber));
+        response->set_string("error", stdsprintf("The provided number of gbtMask's does not fullfill the expectations (%u).", ohUnmaskedNumber));
```
bdorney, line 116: add the provided `gbtMasks` in these error statements.

src/gbt.cpp
```diff
@@ -0,0 +1,381 @@
...
+                // If GBT is not masked
+                if ((gbtMasks[ohConfigIdx] >> gbtN) & 0x1){
+                    gbtConfig.assign(config.begin()+(ohConfigIdx*3+gbtN)*366, 
+                                        config.begin()+(ohConfigIdx*3+gbtN+1)*366);
+                    writeGBTConfigLocal(&la, ohN, gbtN, gbtConfig);
+                }
+            }
+            ohConfigIdx++;
+        }
+    }
```
bdorney, line 147: It seems like you're relying on someone to provide both an `ohMask` (12 bit number) and then a vector that contains 3-bit numbers, coupled with a check on the number of unmasked bits in the `ohMask`.  With this information you're looping over using this to define which GBT's are programmed.  But this is overly complicated in my view.  There's a better way.

All you need is the `gbtMasks` key.  Force the `i^th` element of `gbtMasks` to correspond to the `i^th` optohybrid.  Then a value of `0x0` in the `i^th` element implies the OH is masked; any nonzero value for that element implies the corresponding OH is *not* masked and provides the GBT's to be considered.

This requires minimal modification to the loop in Lines 134-147 while simplifying the rpc input.

src/gbt.cpp
```diff
@@ -0,0 +1,381 @@
...
+    for(unsigned char regN = 0; regN < 3; regN++){
+        const uint16_t regAddress = elinkMappings::elinkToRegisters[elinkMappings::vfatToElink[vfatN]][regN];
+
+        writeGBTRegLocal(la, ohN, gbtN, regAddress, phase);
+    }
+
+    return;
+} //End writeGBTPhaseLocal
+
+void writeGBTPhase(const RPCMsg *request, RPCMsg *response){
```
bdorney, line 167: How you have made this function is more like `writeAllGBTPhases(...)` and while this *is* potentially useful it was not what was requested in #72.  We are looking for a function that takes a single VFAT position and a phase value and sets that phase for the corresponding elink.

Keep this function, but rename it to  `writeAllGBTPhases(...)`, update the documentation to reflect that this is for *all* VFATs on an optohybrid and then make a new function `writeGBTPhase(...)` that is the remote callable method for writing just one VFAT worth of info (as requested in #72).

src/gbt.cpp
```diff
@@ -0,0 +1,381 @@
...
+    {
+        if(vfatMask.test(vfatN)){
+            writeGBTPhaseLocal(&la, ohN, vfatN, phases[vfatN]);
+        }
+    }
+
+    return;
+} //End writeGBTPhase
+
+void writeGBTPhaseMultiLink(const RPCMsg *request, RPCMsg *response){
```
bdorney, line 210: Rename to `writeAllGBTPhaseMultiLink` and update function documentation accordingly.

src/gbt.cpp
```diff
@@ -0,0 +1,381 @@
...
+    auto dbi = lmdb::dbi::open(rtxn, nullptr);
+
+    struct localArgs la = {.rtxn = rtxn, .dbi = dbi, .response = response};
+
+    // ohMask key
+    const std::bitset<32> ohMask(request->get_word("ohMask"));
+    const uint32_t ohMax = readReg(&la, "GEM_AMC.GEM_SYSTEM.CONFIG.NUM_OF_OH");
+
+    // vfatMask key
+    const std::bitset<32> vfatMask(request->get_word("vfatMask"));
```
bdorney, line 226: Multi link functions should *not* take a `vfatMask` from the rpc message but determine the *optohybrid specific* `vfatMask` at runtime, see for example:

https://github.com/cms-gem-daq-project/ctp7_modules/blob/314cec21c7b565f898e461b487e1504907b78d6c/src/calibration_routines.cpp#L1282-L1284

bdorney: (otherwise you *force* all OH's to use *the same* `vfatMask` which is not correct).

src/gbt.cpp
```diff
@@ -0,0 +1,381 @@
...
+
+    // phases key
+    const size_t expectedSize = ohMask.count()*24;
+    if (request->get_binarydata_size("phases") != expectedSize){
+        LOGGER->log_message(LogManager::ERROR, "There must be 24 VFAT phases by OH.");
+        response->set_string("error", "There must be 24 VFAT phases by OH.");
+        return;
+    }
```
bdorney, line 234: Information not in documentation at function definition.

bdorney: Moreover you have not described the expected ordering.

bdorney: Change error message to:

`There must be 24 VFAT phases per OH`

e.g. `by OH` does not make sense.

src/gbt.cpp
```diff
@@ -0,0 +1,381 @@
...
+
+    return;
+} //End writeGBTPhaseMultiLink
+
+std::vector<uint32_t> scanGBTPhaseLocal(localArgs *la, const uint32_t ohN, const uint8_t phaseMin, const uint8_t phaseMax, const uint8_t phaseStep){
+    LOGGER->log_message(LogManager::INFO, stdsprintf("Scannng the phases for OH #%u.", ohN));
+
+    std::vector<uint32_t> results;
+
+    for(uint8_t phase = phaseMin; phase <= phaseMax; phase += phaseStep){
```
bdorney, line 258: At minimum you should check that `phaseMax` makes sense; e.g. an 8-bit number goes to 255 while the max phase that can be set is 15 (16? I don't remember if indexing starts at 1 or 0).

src/gbt.cpp
```diff
@@ -0,0 +1,381 @@
...
+            phaseResult |= (0x1 << vfatN);
+        }
+
+        results.push_back(phaseResult);
+    }
+
+    return results;
+} //End scanGBTPhaseLocal
+
+void scanGBTPhase(const RPCMsg *request, RPCMsg *response){
```
bdorney, line 295: I was expecting that this would *also* take a key that would determine the number of scans that would be performed as described [here](https://github.com/cms-gem-daq-project/ctp7_modules/issues/72#issuecomment-441643357).  The goal is to have a tool that makes `N` scans in a single RPC call (otherwise you lose a lot of time to the RPC message) and that this can be used to potentially determine the GBT phase settings.

We saw that between long & short boards the phase values differ and that even performing multiple scans on the *same* OH+GEB pair one after another give different results (edge effects).

src/gbt.cpp
```diff
@@ -0,0 +1,381 @@
...
+
+        // Set the new phases
+        for(uint32_t vfatN = 0; vfatN < 24; vfatN++){
+            writeGBTPhaseLocal(la, ohN, vfatN, phase);
+        }
+
+        // Try to synchronise the VFAT's
+        std::this_thread::sleep_for(std::chrono::milliseconds(100));
+        writeReg(la, "GEM_AMC.GEM_SYSTEM.CTRL.LINK_RESET", 1);
+        std::this_thread::sleep_for(std::chrono::milliseconds(300));
```
bdorney, line 275: Why are these wait times *so* long?  Especially if this scan application is run multiple times this would take *forever*.

src/gbt.cpp
```diff
@@ -0,0 +1,381 @@
...
+        }
+    }
+
+    return;
+} //End writeGBTPhaseMultiLink
+
+std::vector<uint32_t> scanGBTPhaseLocal(localArgs *la, const uint32_t ohN, const uint8_t phaseMin, const uint8_t phaseMax, const uint8_t phaseStep){
+    LOGGER->log_message(LogManager::INFO, stdsprintf("Scannng the phases for OH #%u.", ohN));
+
+    std::vector<uint32_t> results;
```
bdorney, line 256: In the function documentation you have not described the contents, nor the order, of this container.  Use the `\returns` field in `doxygen`.

src/gbt.cpp
```diff
@@ -0,0 +1,381 @@
...
+    if (ohN > ohMax){
+        LOGGER->log_message(LogManager::ERROR, stdsprintf("The ohN parameter supplied (%u) exceeds the number on OH's supported by the CTP7 (%u).", ohN, ohMax));
+        response->set_string("error", stdsprintf("The ohN parameter supplied (%u) exceeds the number on OH's supported by the CTP7 (%u).", ohN, ohMax));
+        return;
+    }
+
+    // phases config keys
+    const uint8_t phaseMin = request->get_word("phaseMin");
+    const uint8_t phaseMax = request->get_word("phaseMax");
+    const uint8_t phaseStep = request->get_word("phaseStep");
```
bdorney, line 318: Should *also* check here that `phaseMax` makes sense.  Keep the check in both places since the local callable function can be called elsewhere.

src/gbt.cpp
```diff
@@ -0,0 +1,381 @@
...
+
+    // Perform the scan
+    std::vector<uint32_t> results = scanGBTPhaseLocal(&la, ohN, phaseMin, phaseMax, phaseStep);
+
+    response->set_word_array("results", results);
```
bdorney, line 323: Needs to be explained in function documentation.

src/gbt.cpp
```diff
@@ -0,0 +1,381 @@
...
+    struct localArgs la = {.rtxn = rtxn, .dbi = dbi, .response = response};
+
+    // ohMask key
+    const std::bitset<32> ohMask(request->get_word("ohMask"));
+    const uint32_t ohMax = readReg(&la, "GEM_AMC.GEM_SYSTEM.CONFIG.NUM_OF_OH");
+
+    // phase keys
+    const uint8_t phaseMin = request->get_word("phaseMin");
+    const uint8_t phaseMax = request->get_word("phaseMax");
+    const uint8_t phaseStep = request->get_word("phaseStep");
```
bdorney, line 346: check that `phaseMax` makes sense.

src/gbt.cpp
```diff
@@ -0,0 +1,381 @@
...
+
+    // Perform the scan
+    std::vector<uint32_t> results = scanGBTPhaseLocal(&la, ohN, phaseMin, phaseMax, phaseStep);
+
+    response->set_word_array("results", results);
+
+    return;
+} //Enc scanGBTPhase
+
+void scanGBTPhaseMultiLink(const RPCMsg *request, RPCMsg *response){
```
bdorney, line 328: Again was expecting this to take a parameter that would define the number of scans to be performed in *this* rpc call.

src/gbt.cpp
```diff
@@ -0,0 +1,381 @@
...
+    for(uint32_t ohN = 0; ohN < ohMax; ohN++)
+    {
+        if(ohMask.test(ohN)){
+            auto ohResults = scanGBTPhaseLocal(&la, ohN, phaseMin, phaseMax, phaseStep);
+
+            results.insert(results.end(), ohResults.begin(), ohResults.end());
+        }
+    }    
+
+    response->set_word_array("results", results);
```
bdorney, line 360: documentation needed

bdorney (changes requested):
Lacking appropriate function documentation.  One key function is missing (e.g. `writeGBTPhase` what is provided is actually a `writeAllGBTPhases`). Additionally architectural changes needed in several functions to reduce complexity.

include/gbt.h
lpetre-ulb: It scans _all_ the `GBT`'s on `ohN`. I used `optical link` because it was used somewhere else where it seems to means the OH index. I'll clarify that point.

include/gbt.h
lpetre-ulb: It write just a single phase. I forgot to update the `\brief line`...

src/gbt.cpp
lpetre-ulb: What could be done is to use `#ifdef` when the values are different between GE1/1 and GE2/1. It is then possible to chose at compile time if modules are for GE1/1 or GE2/1 (with the `-D` option of the compiler). In fact, this solution does not requires to centralize all the constants in a single location. It might however be cleaner, particularly if the constants are sorted within namespaces.

src/gbt.cpp
lpetre-ulb: Yes, right, it should be in documentation. The size is already checked in the non-local versions and using it with a incorrect size vector will just fail (silently if too short, with errors from `writeGBTRegLocal`).

src/gbt.cpp
lpetre-ulb: Ok, I will add these informations.

src/gbt.cpp
lpetre-ulb: Yes, that's a better possibility, even if the keys will loose in uniformity. However, I would still need to keep track of the number of OH or GBT's which have been configured. Except, it the user writes padding in the `config` blob at the place of masked OH's/GBT's, but the data will be less compressed.

src/gbt.cpp
lpetre-ulb: Ok, I'll add the function.

src/gbt.cpp
lpetre-ulb: Indeed, forcing all the OH's to use the same `vfatMask` is maybe not the best solution. But using `getOHVFATMaskLocal` won't work. For example, after a phase scan, all the phases are left at value 15 which never works. I'll instead provide a array of `vfatMask` (like the gbtMasks).

src/gbt.cpp
lpetre-ulb: Of course, I missed that.

src/gbt.cpp
lpetre-ulb: I misunderstood that request; I understood it as the `phaseMin`, `phaseMax` and `phaseStep` parameters. I'll add a `N` scans parameters. As the number of phases is very small, does it worth keeping the `phaseMin`, `phaseMax` and `phaseStep` parameters ?

src/gbt.cpp
lpetre-ulb: These are the wait times from the original `gbt.py` tool :

https://github.com/cms-gem-daq-project/reg_utils/blob/7b9cf050c63dae5f3a9b5803389130cff2f039fa/python/reg_interface/arm/gbt_utils.py#L68-L72

I don't know if they can reliably reduced.

src/gbt.cpp
bdorney: Right, but the local version could be called by some other function so it too should work as a standalone object with proper handling.

src/gbt.cpp
bdorney: I’m not sure I understand this issue. Could you elaborate?

src/gbt.cpp
bdorney: Yes you’re right. On further reflection why do we need the VFAT mask at all here?  Can it be dropped? The only thing that seems to actually matter is the attempt to read CFG_RUN and this could be handled by catching in the bus error that would be raised by a failed read.  Or did I miss something?

bdorney: I would propose dropping VFAT mask, accepting scanning over dead/empty VFATs, and catching bus errors.  I don’t think this would be a significant issue.

src/gbt.cpp
bdorney: You’ve already coded them so if it’s easier to keep them why not; just assign default values to them in the function declaration in the header file, e.g. min 0, max 15, step size 1.

src/gbt.cpp
bdorney: Ah I see. Mmhmm, can you contact @evka85 to see if they can be reliably reduced or what the rational is behind their length?

src/gbt.cpp
lpetre-ulb: I don't understand exactly what you mean; this is the `writeAllGBTPhaseMultiLink` function. In the scan phase function, the scan indeed runs over the dead/empty VFAT's.

My idea behind an array of `vfatMask` is to be able to write some phases spread over all links in a single RPC call while leaving untouched the phases we do not want to change. Isn't it useful ?

src/gbt.cpp
lpetre-ulb: Ok, I'll send him an email. In the meantime, I'll leae the wait times untouched.

src/gbt.cpp
lpetre-ulb: My idea for the `writeGBTConfigMultiLink` function was to be able to use a different GBTX configuration file for each GBT to program, in a single RPC call. It allows to easily mix short and long chambers. These configurations must be concatenated in order to fit in key system, but there are multiple packing possibilities :
1. Concatenate all configurations. If the corresponding GBT is masked, you can add 366 zeroes, but the data is not compressed in any way.
2. Concatenate the 3 configurations only for the unmasked OH's. (This is what I did by imitation with the single link function call.)
3. Concatenate only the configurations of the unmasked GBT's. (This is what I plan to do.)

The the last two possibilities require to keep track of the position in the packed configuration data. But there are ways to keep the tracking clean.

src/gbt.cpp
lpetre-ulb: Ok, I'll do as you suggest.

But that raise a question for me. Where do you expect the checks to be done ? For example, in the `writeGBTRegLocal` function, I should check that `ohN` is inferior to the number of OH's supported by the CTP7. However, the performances will suffer from a systematic access to the `GEM_AMC.GEM_SYSTEM.CONFIG.NUM_OF_OH` register. This is the reason why I removed some checks before committing.

src/gbt.cpp
bdorney: For this PR don't worry about it because it's outside the scope.  I just continue to raise this issue.

src/gbt.cpp
bdorney: Yes it makes sense to have a register check only when necessary since this causes a transaction on the bus which is O(us) but the CTP7's Zynq processor is 667 MHz [[1]](http://iopscience.iop.org/article/10.1088/1748-0221/11/02/C02011/pdf) so checks that do not require register access are "cheap."

Does this clarify things?

src/gbt.cpp
bdorney: Okay; I look forward to seeing the implementation.  But I think you got my primary concern: namely we don't need to provide two inputs that could potentially contradict.  This can be done just with `gbtMasks` and the `ohMasks` input is not needed.

src/gbt.cpp
bdorney: Can you provide a use case in which you would *want* to do that?  This function in my mind is called at configure.  Then everything is configured.  In which case would you want to write only some phases?  I can't think of a case for that.

src/gbt.cpp
lpetre-ulb: Not really... Indeed, checks that require access to the registers are "expensive", that was the point of my question. Maybe it will be clearer if I rephrase it.

When is a check absolutely necessary ? It is possible to check every input parameter at each function call, but there is a (potentially massive) performance hit. So I decided to check only parameters coming from the RPC keys (even if I forgot `phase...`). The internal interface has then minimal checks and is supposed to be used correctly by the developer (of course, I should have written more documentation).

It is a matter of development choice. Can I conclude that you prefer that all input parameters are checked at each function call, except if the check requires register access ?

src/gbt.cpp
lpetre-ulb: My use case was the correction of multiple VFAT phases in test stands while leaving the others at their default value. I could also argue that if the only use case is at configure, this `writeAllPhasesMultiLink` method is well performed with an on-the-fly modification of the configuration blob. But, OK, I'll remove the `vfatMasks`.

bdorney:
Status update on this PR?

lpetre-ulb:
Yes, I tested the modifications on the hardware this afternoon. The methods still work as expected. As I better understand what you expect, I preferred to start a new clean PR which addresses the points you raised : https://github.com/cms-gem-daq-project/ctp7_modules/pull/75.

PR #74, opened 2018-11-30
Use the new cached ADC registers introduced in release 3.7.1 of the ctp7 FW

AndrewLevin:
@evka85 released a new version of the ctp7 FW which uses caching for the registers `ADC0` and `ADC1` (https://github.com/evka85/GEM_AMC/releases/tag/v3.7.1). The corresponding changes to the ctp7 software are in the pull request.

## Description
The `dacScanLocal` function in `calibration_routines.cpp` has been changed to use the cache registers `ADC0_CACHED` and `ADC1_CACHED` if they exist, while also maintaining backward compatibility. The cache registers need to be updated before every read by reading or writing to the registers `ADC0_UPDATE` and `ADC1_UPDATE`, and then a 20 microsecond delay is needed to allow the cache registers to be updated.

### Types of changes

## Motivation and Context
This change was required for the ctp7 software to be consistent with the ctp7 firmware release v3.7.1, and to resolve an issue we found with reads of ADC0 and ADC1 timing out: http://cmsonline.cern.ch/cms-elog/1074808

## How Has This Been Tested?
The change has been tested by running 11 DAC scans on eagle64. Details, including the summary plot for one of the DAC scans, are in this elog:

http://cmsonline.cern.ch/cms-elog/1076528

## Checklist:


src/calibration_routines.cpp
```diff
@@ -1201,6 +1222,12 @@ std::vector<uint32_t> dacScanLocal(localArgs *la, uint32_t ohN, uint32_t dacSele
                 writeReg(la, strDacReg, dacVal);
                 for(int i=0; i<100; ++i){ //Read 100 times and take avg value
                     //Read the ADC
+                    if (foundAdcCached){
+                        //either reading or writing this register will trigger a cache update
+                        readRawAddress(adcCacheUpdateAddr[vfatN], la->response);
+                        //updating the cache takes 20 us
+                        std::this_thread::sleep_for(std::chrono::microseconds(20));
```
bdorney, line 1231: I would increase this to 25us (increases time to scan a single DAC by 0.12s) instead of using the 20us statement from Evaldas.  Without knowing if there is a spread in this update time you might just be sampling the same value and not a new value and then this doesn't accurately reflect the system.

bdorney (changes requested):
Increase slightly the cached register update time wait.

src/calibration_routines.cpp
```diff
@@ -1201,6 +1222,12 @@ std::vector<uint32_t> dacScanLocal(localArgs *la, uint32_t ohN, uint32_t dacSele
                 writeReg(la, strDacReg, dacVal);
                 for(int i=0; i<100; ++i){ //Read 100 times and take avg value
                     //Read the ADC
+                    if (foundAdcCached){
+                        //either reading or writing this register will trigger a cache update
+                        readRawAddress(adcCacheUpdateAddr[vfatN], la->response);
+                        //updating the cache takes 20 us
+                        std::this_thread::sleep_for(std::chrono::microseconds(25));
+                    }
                     Temp = readRawAddress(adcAddr[vfatN], la->response);
```
jsturdy: This isn't related to this PR per se, but this is a rather poorly named variable.
On top of that, it's not even clear that it's necessary as it is declared outside the loop, assigned to the read value, and then immediately added to the value that matters.

jsturdy: Related, in [L1223](https://github.com/cms-gem-daq-project/ctp7_modules/blob/1fdc445546755ecdf388cf2f4b582227f707e9c3/src/calibration_routines.cpp#L1223) and [L1234](https://github.com/cms-gem-daq-project/ctp7_modules/blob/1fdc445546755ecdf388cf2f4b582227f707e9c3/src/calibration_routines.cpp#L1234), things would be better served by a `nReads` variable declared elsewhere

Additionally, it doesn't look like `adcVal` is ever reset after finishing the loop on a given `DAC` for a given `VFAT` at a given `DAC` set point; I can't imagine that this is the desired behaviour.
```pseudocode
init adcVal to 0
for minVal to maxVal
  for vfat
    for read
      adcVal += readVal
    adcVal /= 100 # this is now the initial value for the next VFAT
    storeRetVal(index,adcVal)
```

jsturdy (dismissed):
I think this PR is probably fine, but reading a few surrounding bits and found a few bugs in the module, you could fix them in this PR, but it isn't strictly mandatory

src/calibration_routines.cpp
evka85: The 20us I quoted already has about 50% margin built in, so that should be safe. But of course I'm not opposed to increasing to 25us -- this won't increase the total run time by any noticeable amount, unless this is read at least a million times.

bdorney: Ah okay; thanks for pointing this out! If the 25us is already implemented then I would leave it as is; unless additional changes are pending.

I expect this to be a “short term” solution anyway; with final solution coming from an upgrade of the Linux core by UWISC

bdorney (dismissed):

jsturdy (changes requested):
https://github.com/cms-gem-daq-project/ctp7_modules/blob/6c8d018d5a87f96fc6ec63b51672a22491be656a/src/calibration_routines.cpp#L1224 
should move to:
https://github.com/cms-gem-daq-project/ctp7_modules/blob/6c8d018d5a87f96fc6ec63b51672a22491be656a/src/calibration_routines.cpp#L1206 (it doesn't need to be inside two loops, and could conceivably be passed in as a function parameter)

Then, https://github.com/cms-gem-daq-project/ctp7_modules/blob/6c8d018d5a87f96fc6ec63b51672a22491be656a/src/calibration_routines.cpp#L1235 should reference this value, instead of the old hardcoded value

jsturdy (approved):

PR #75, opened 2018-12-10
 Implement RPC methods for configuring the GBTX's v2

lpetre-ulb:
## Description

[This PR is the continuation of the [previous PR about GBTX's](https://github.com/cms-gem-daq-project/ctp7_modules/pull/73). I started the new PR in order to keep the git log clean, but also the GitHub comments readable.]

Addresses https://github.com/cms-gem-daq-project/ctp7_modules/issues/72.

This PR adds a new RPC module which implements 3 new methods related to the GBTX configuration :

* `writeGBTConfig` updates the configuration of one chosen GBTX.
* `writeGBTPhase` updates one VFAT RX phase of one OH.
* `scanGBTPhases` scans the phases on one OH.

It lacks the MultiLink versions, but all the important functionalities are working. If the code and the documentation meet the expectation, I will add these methods.

### Types of changes

## Motivation and Context

This PR fixes issue https://github.com/cms-gem-daq-project/ctp7_modules/issues/72.

## How Has This Been Tested?

These tests were performed to see if the functionalities implemented in the `gbt.py` were still working with the RPC modules and so the new CTP7 releases :

* Writing the GBTX configurations. It has been tested for different OH and multiple GBTX's.
* Setting the GBTX RX phases for 1 VFAT at a time.
* Scanning the GBTX RX phases for one OH.

The environment was the following :
* CTP7 : eagle63
* CTP7 firmware : 3.7.0
* CTP7 build-id : CTP7-GENERIC-20180529T153916-0500-4935611
* `xHAL` commit https://github.com/cms-gem-daq-project/xhal/pull/99/commits/cfa38ed635ca078c0d3a2c24d02db9b9da17c327
* `cmsgemos` commit https://github.com/cms-gem-daq-project/cmsgemos/pull/226/commits/fd0747ed87a4f087bb00d1857ac6feae46a9d5fe
* The new `gbt.py` used for these tests comes from commit (in progress).

## Checklist:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -88,8 +88,8 @@ lib/optohybrid.so: src/optohybrid.cpp
 lib/calibration_routines.so: src/calibration_routines.cpp
 	$(CXX) $(CFLAGS) -std=c++1y -O3 -pthread $(INC) $(LDFLAGS) -fPIC -shared -Wl,-soname,calibration_routines.so -o $@ $< -lwisci2c -lxhal -llmdb -l:utils.so -l:extras.so -l:optohybrid.so -l:vfat3.so -l:amc.so
 
-lib/gbt.so: src/gbt.cpp include/gbt.h include/moduleapi.h include/memhub.h include/utils.h
-	$(CXX) $(CFLAGS) -std=c++1y -O3 -pthread $(INC) $(LDFLAGS) -fPIC -shared -Wl,-soname,gbt.so -o $@ $< -l:memhub.so
+lib/gbt.so: src/gbt.cpp include/gbt.h include/gbt_constants.h include/moduleapi.h include/memhub.h include/utils.h lib/utils.so
+	$(CXX) $(CFLAGS) -std=c++1y -O3 -pthread $(INC) $(LDFLAGS) -fPIC -shared -Wl,-soname,gbt.so -o $@ $< -l:memhub.so -l:utils.so
 
 clean: cleanrpm
 	-rm -rf lib/*.so
--- a/Makefile
+++ b/Makefile
@@ -40,6 +40,7 @@ TARGET_LIBS += lib/daq_monitor.so
 TARGET_LIBS += lib/vfat3.so
 TARGET_LIBS += lib/optohybrid.so
 TARGET_LIBS += lib/calibration_routines.so
+TARGET_LIBS += lib/gbt.so
 
 .PHONY: clean rpc prerpm
 
@@ -86,6 +87,9 @@ lib/optohybrid.so: src/optohybrid.cpp
 
 lib/calibration_routines.so: src/calibration_routines.cpp
 	$(CXX) $(CFLAGS) -std=c++1y -O3 -pthread $(INC) $(LDFLAGS) -fPIC -shared -Wl,-soname,calibration_routines.so -o $@ $< -lwisci2c -lxhal -llmdb -l:utils.so -l:extras.so -l:optohybrid.so -l:vfat3.so -l:amc.so
+
+lib/gbt.so: src/gbt.cpp include/gbt.h include/moduleapi.h include/memhub.h include/utils.h
+	$(CXX) $(CFLAGS) -std=c++1y -O3 -pthread $(INC) $(LDFLAGS) -fPIC -shared -Wl,-soname,gbt.so -o $@ $< -l:memhub.so
 
 clean: cleanrpm
 	-rm -rf lib/*.so
--- a/Makefile
+++ b/Makefile
@@ -90,7 +90,7 @@ lib/optohybrid.so: src/optohybrid.cpp
 lib/calibration_routines.so: src/calibration_routines.cpp
 	$(CXX) $(CFLAGS) -std=c++1y -O3 -pthread $(INC) $(LDFLAGS) -fPIC -shared -Wl,-soname,calibration_routines.so -o $@ $< -lwisci2c -lxhal -llmdb -l:utils.so -l:extras.so -l:optohybrid.so -l:vfat3.so -l:amc.so
 
-lib/gbt.so: src/gbt.cpp include/gbt.h include/hw_constants.h include/moduleapi.h include/memhub.h include/utils.h lib/utils.so
+lib/gbt.so: src/gbt.cpp include/gbt.h include/hw_constants.h include/hw_constants_checks.h include/moduleapi.h include/memhub.h include/utils.h lib/utils.so
 	$(CXX) $(CFLAGS) -std=c++1y -O3 -pthread $(INC) $(LDFLAGS) -fPIC -shared -Wl,-soname,gbt.so -o $@ $< -l:memhub.so -l:utils.so
 
 clean: cleanrpm
--- a/Makefile
+++ b/Makefile
@@ -26,6 +26,8 @@ IncludeDirs += ${BUILD_HOME}/xhal/xhalcore/include
 #IncludeDirs += /opt/cactus/include
 INC=$(IncludeDirs:%=-I%)
 
+CFLAGS += -DGEM_VARIANT="ge11"
+
 LDFLAGS+= -L${BUILD_HOME}/xhal/xhalarm/lib
 LDFLAGS+= -L${BUILD_HOME}/$(Package)/lib
 
--- a/Makefile
+++ b/Makefile
@@ -26,7 +26,11 @@ IncludeDirs += ${BUILD_HOME}/xhal/xhalcore/include
 #IncludeDirs += /opt/cactus/include
 INC=$(IncludeDirs:%=-I%)
 
-CFLAGS += -DGEM_VARIANT="ge11"
+ifndef GEM_VARIANT
+	GEM_VARIANT = ge11
+endif
+
+CFLAGS += -DGEM_VARIANT="${GEM_VARIANT}"
 
 LDFLAGS+= -L${BUILD_HOME}/xhal/xhalarm/lib
 LDFLAGS+= -L${BUILD_HOME}/$(Package)/lib
--- a/Makefile
+++ b/Makefile
@@ -88,7 +88,7 @@ lib/optohybrid.so: src/optohybrid.cpp
 lib/calibration_routines.so: src/calibration_routines.cpp
 	$(CXX) $(CFLAGS) -std=c++1y -O3 -pthread $(INC) $(LDFLAGS) -fPIC -shared -Wl,-soname,calibration_routines.so -o $@ $< -lwisci2c -lxhal -llmdb -l:utils.so -l:extras.so -l:optohybrid.so -l:vfat3.so -l:amc.so
 
-lib/gbt.so: src/gbt.cpp include/gbt.h include/gbt_constants.h include/moduleapi.h include/memhub.h include/utils.h lib/utils.so
+lib/gbt.so: src/gbt.cpp include/gbt.h include/hw_constants.h include/moduleapi.h include/memhub.h include/utils.h lib/utils.so
 	$(CXX) $(CFLAGS) -std=c++1y -O3 -pthread $(INC) $(LDFLAGS) -fPIC -shared -Wl,-soname,gbt.so -o $@ $< -l:memhub.so -l:utils.so
 
 clean: cleanrpm
--- a/include/gbt.h
+++ b/include/gbt.h
@@ -1,0 +1,10 @@
+/*! \file
+ *  \brief RPC module for GBT methods
+ *  \author Laurent Pétré <lpetre@ulb.ac.be>
+ */
+
+#ifndef GBT_H
+#define GBT_H
+
+#endif
+
--- a/include/gbt.h
+++ b/include/gbt.h
@@ -10,6 +10,47 @@
 
 #include "utils.h"
 
+/*! \brief Scan the GBT phases of one OptoHybrid.
+ *  \param[in] request RPC response message
+ *  \param[out] response RPC response message
+ *
+ *  The method expects the following RPC keys :
+ *  - `word ohN` : OptoHybrid index number.
+ *  - `word N` : The number of times the scan must performed.
+ *  - `word phaseMin` : Lowest phase to scan (min = 0).
+ *  - `word phaseMax` : Highest phase to scan (max = 15).
+ *  - `word phaseStep` : Step to scan the phases.
+ *
+ *  The method returns the following RPC keys :
+ *  - `word_array results` : Array of word containing the results of the scans. See details of the `scanGBTPhase` function for a description of the words.
+ *  - `string error` : If an error occurs, this keys exists and contains the error message.
+ */
+void scanGBTPhases(const RPCMsg *request, RPCMsg *response);
+
+/*! \brief Local callable version of `scanGBTPhase`.
+ *  \param[in, out] la Local arguments structure.
+ *  \param[out] results A `std::vector` containing the results of the scan. See details for the items description.
+ *  \param[in] ohN OptoHybrid index number.
+ *  \param[in] N The number of times the scan must performed.
+ *  \param[in] phaseMin Lowest phase to scan (min = 0).
+ *  \param[in] phaseMax Highest phase to scan (max = 15).
+ *  \param[in] phaseStep Step to scan the phases.
+ *  \return Returns `false` in case of success; `true` in case of error. The precise error is logged and written to the `error` RPC key.
+ *
+ *  \detail
+ *
+ *  The scan seeks for valid RX phases for the VFAT's of one OptoHybrid. A phase is considered valid when `LINK_GOOD = 1`, `SYNC_ERR_CNT = 0` and `CFG_RUN != 0xdeaddead`. In order to improve the reliability of the scan, it is repeated `N` times.
+ *
+ *  Each item of the results vector is composed as follow :
+ *  - [31:28] : ohN
+ *  - [27:24] : phase
+ *  - [23:0] : result of the N scans (and) for the i^th VFAT. A `1` indicates a good phase while a `0` indicates a bad phase.
+ *
+ *  The length of the vector is determined by `phaseMin`, `phaseMax` and `phaseStep` : 
+ *  (`phaseMax` - `phaseMin` + 1)/`phaseStep`
+ */
+bool scanGBTPhasesLocal(localArgs *la, std::vector<uint32_t> &results, const uint32_t ohN, const uint32_t N = 1, const uint8_t phaseMin = gbt::phaseMin, const uint8_t phaseMax = gbt::phaseMax, const uint8_t phaseStep = 1);
+
 /*! \brief Write the GBT configuration of one OptoHybrid.
  *  \param[in] request RPC response message.
  *  \param[out] response RPC response message.
--- a/include/gbt.h
+++ b/include/gbt.h
@@ -6,7 +6,32 @@
 #ifndef GBT_H
 #define GBT_H
 
+#include "gbt_constants.h"
+
 #include "utils.h"
+
+/*! \brief Write the GBT configuration of one OptoHybrid.
+ *  \param[in] request RPC response message.
+ *  \param[out] response RPC response message.
+ *
+ *  The method expects the following RPC keys :
+ *  - `word ohN` : OptoHybrid index number.
+ *  - `word gbtN` : Index of the GBT to configure. There 3 GBT's per OptoHybrid in the GE1/1 chambers.
+ *  - `binarydata config` : Configuration blob of the GBT. This is a 366 element long vector of the registers to write. The registers are sorted from address 0 to address 365.
+ *
+ *  The method returns the following RPC keys :
+ *  - `string error` : If an error occurs, this keys exists and contains the error message.
+ */
+void writeGBTConfig(const RPCMsg *request, RPCMsg *response);
+
+/*! \brief Local callable version of `writeGBTConfig`
+ *  \param[in, out] la Local arguments structure.
+ *  \param[in] ohN OptoHybrid index number.
+ *  \param[in] gbtN Index of the GBT to write. There 3 GBT's per OptoHybrid in the GE1/1 chambers.
+ *  \param[in] config Configuration blob of the GBT. This is a 366 elements long array whose each element is the value of one register sorted from address 0 to address 365.
+ *  \return Returns `false` in case of success; `true` in case of error. The precise error is logged and written to the `error` RPC key.
+ */
+bool writeGBTConfigLocal(localArgs *la, const uint32_t ohN, const uint32_t gbtN, const gbt::config_t &config);
 
 /*! \brief This function writes a single register in the given GBT of the given OptoHybrid.
  *  \param[in, out] la Local arguments structure.
--- a/include/gbt.h
+++ b/include/gbt.h
@@ -49,7 +49,7 @@ void scanGBTPhases(const RPCMsg *request, RPCMsg *response);
  *  The length of the vector is determined by `phaseMin`, `phaseMax` and `phaseStep` : 
  *  (`phaseMax` - `phaseMin` + 1)/`phaseStep`
  */
-bool scanGBTPhasesLocal(localArgs *la, std::vector<uint32_t> &results, const uint32_t ohN, const uint32_t N = 1, const uint8_t phaseMin = gbt::phaseMin, const uint8_t phaseMax = gbt::phaseMax, const uint8_t phaseStep = 1);
+bool scanGBTPhasesLocal(localArgs *la, std::vector<uint32_t> &results, const uint32_t ohN, const uint32_t N = 1, const uint8_t phaseMin = gbt::PHASE_MIN, const uint8_t phaseMax = gbt::PHASE_MAX, const uint8_t phaseStep = 1);
 
 /*! \brief Write the GBT configuration of one OptoHybrid.
  *  \param[in] request RPC response message.
--- a/include/gbt.h
+++ b/include/gbt.h
@@ -22,14 +22,13 @@
  *  - `word phaseStep` : Step to scan the phases.
  *
  *  The method returns the following RPC keys :
- *  - `word_array results` : Array of word containing the results of the scans. See details of the `scanGBTPhase` function for a description of the words.
+ *  - `word_array OHX.VFATY` : Array of word containing the results of the scans. See details of the `scanGBTPhase` function for a description of the words.
  *  - `string error` : If an error occurs, this keys exists and contains the error message.
  */
 void scanGBTPhases(const RPCMsg *request, RPCMsg *response);
 
 /*! \brief Local callable version of `scanGBTPhase`.
  *  \param[in, out] la Local arguments structure.
- *  \param[out] results A `std::vector` containing the results of the scan. See details for the items description.
  *  \param[in] ohN OptoHybrid index number.
  *  \param[in] N The number of times the scan must performed.
  *  \param[in] phaseMin Lowest phase to scan (min = 0).
@@ -41,15 +40,12 @@ void scanGBTPhases(const RPCMsg *request, RPCMsg *response);
  *
  *  The scan seeks for valid RX phases for the VFAT's of one OptoHybrid. A phase is considered valid when `LINK_GOOD = 1`, `SYNC_ERR_CNT = 0` and `CFG_RUN != 0xdeaddead`. In order to improve the reliability of the scan, it is repeated `N` times.
  *
- *  Each item of the results vector is composed as follow :
- *  - [31:28] : ohN
- *  - [27:24] : phase
- *  - [23:0] : result of the N scans (and) for the i^th VFAT. A `1` indicates a good phase while a `0` indicates a bad phase.
+ *  The results are returned as RPC keys named `OHX.VFATY` where X is the index of the OptoHyrid and Y the index of the VFAT within this OptoHybrid.
+ *  Each key is a word array of 16 elements for the 16 possible phases ordered from phase 0 to phase 15.
+ *  Each word is the number of time the scan was "good" out of the total number of scan requested, N.
  *
- *  The length of the vector is determined by `phaseMin`, `phaseMax` and `phaseStep` : 
- *  (`phaseMax` - `phaseMin` + 1)/`phaseStep`
  */
-bool scanGBTPhasesLocal(localArgs *la, std::vector<uint32_t> &results, const uint32_t ohN, const uint32_t N = 1, const uint8_t phaseMin = gbt::PHASE_MIN, const uint8_t phaseMax = gbt::PHASE_MAX, const uint8_t phaseStep = 1);
+bool scanGBTPhasesLocal(localArgs *la, const uint32_t ohN, const uint32_t N = 1, const uint8_t phaseMin = gbt::PHASE_MIN, const uint8_t phaseMax = gbt::PHASE_MAX, const uint8_t phaseStep = 1);
 
 /*! \brief Write the GBT configuration of one OptoHybrid.
  *  \param[in] request RPC response message.
--- a/include/gbt.h
+++ b/include/gbt.h
@@ -33,6 +33,29 @@ void writeGBTConfig(const RPCMsg *request, RPCMsg *response);
  */
 bool writeGBTConfigLocal(localArgs *la, const uint32_t ohN, const uint32_t gbtN, const gbt::config_t &config);
 
+/*! \brief Write the phase of a single VFAT.
+ *  \param[in] request RPC response message.
+ *  \param[out] response RPC response message.
+ *
+ *  The method expects the following RPC keys :
+ *  - `word ohN` : OptoHybrid index number.
+ *  - `word vfatN` : Index of the VFAT to configure. There 24 VFAT's per OptoHybrid in the GE1/1 chambers.
+ *  - `word phase` : Phase value to write. The values span from 0 to 15.
+ *
+ *  The method returns the following RPC keys :
+ *  - `string error` : If an error occurs, this keys exists and contains the error message.
+ */
+void writeGBTPhase(const RPCMsg *request, RPCMsg *response);
+
+/*! \brief Local callable version of `writeGBTPhase`
+ *  \param[in, out] la Local arguments structure.
+ *  \param[in] ohN OptoHybrid index number.
+ *  \param[in] vfatN VFAT index of which the phase is changed.
+ *  \param[in] phase Phase value to write. Minimal phase is 0 and maximal phase is 15.
+ *  \return Returns `false` in case of success; `true` in case of error. The precise error is logged and written to the `error` RPC key.
+ */
+bool writeGBTPhaseLocal(localArgs *la, const uint32_t ohN, const uint32_t vfatN, const uint8_t phase);
+
 /*! \brief This function writes a single register in the given GBT of the given OptoHybrid.
  *  \param[in, out] la Local arguments structure.
  *  \param[in] ohN OptoHybrid index number. Warning : the value of this parameter is not checked because of the cost of a register access.
--- a/include/gbt.h
+++ b/include/gbt.h
@@ -6,5 +6,17 @@
 #ifndef GBT_H
 #define GBT_H
 
+#include "utils.h"
+
+/*! \brief This function writes a single register in the given GBT of the given OptoHybrid.
+ *  \param[in, out] la Local arguments structure.
+ *  \param[in] ohN OptoHybrid index number. Warning : the value of this parameter is not checked because of the cost of a register access.
+ *  \param[in] gbtN Index of the GBT to write. There 3 GBT's per OptoHybrid in the GE1/1 chambers.
+ *  \param[in] address GBT register address to write. The highest writable address is 365.
+ *  \param[in] value Value to write to the GBT register.
+ *  \return Returns `false` in case of success; `true` in case of error. The precise error is logged and written to the `error` RPC key.
+ */
+bool writeGBTRegLocal(localArgs *la, const uint32_t ohN, const uint32_t gbtN, const uint16_t address, const uint8_t value);
+
 #endif
 
--- a/include/gbt.h
+++ b/include/gbt.h
@@ -6,7 +6,7 @@
 #ifndef GBT_H
 #define GBT_H
 
-#include "gbt_constants.h"
+#include "hw_constants.h"
 
 #include "utils.h"
 
--- a/include/gbt_constants.h
+++ b/include/gbt_constants.h
@@ -1,0 +1,22 @@
+/*! \file
+ *  \brief Header containing the constants for the GBT module.
+ *  \author Laurent Pétré <lpetre@ulb.ac.be>
+ */
+
+#ifndef GBT_CONSTANTS_H
+#define GBT_CONSTANTS_H
+
+#include <stdint.h>
+
+namespace gbt{
+    /*! \brief The number of GBT's per OptoHybrid.
+     */
+    constexpr uint32_t gbtsPerOH = 3;
+
+    /*! \brief The size of the GBT configuration address space. The corresponding addresses span from 0 to 365.
+     */
+    constexpr uint16_t configSize = 366;
+}
+
+#endif
+
--- a/include/gbt_constants.h
+++ b/include/gbt_constants.h
@@ -7,6 +7,7 @@
 #define GBT_CONSTANTS_H
 
 #include <stdint.h>
+#include <array>
 
 namespace gbt{
     /*! \brief The number of GBT's per OptoHybrid.
@@ -16,6 +17,10 @@ namespace gbt{
     /*! \brief The size of the GBT configuration address space. The corresponding addresses span from 0 to 365.
      */
     constexpr uint16_t configSize = 366;
+
+    /*! \brief This type defines a GBT configuration blob.
+     */
+    typedef std::array<uint8_t, configSize> config_t;
 }
 
 #endif
--- a/include/gbt_constants.h
+++ b/include/gbt_constants.h
@@ -14,6 +14,10 @@ namespace gbt{
      */
     constexpr uint32_t gbtsPerOH = 3;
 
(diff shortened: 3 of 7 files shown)


Makefile
```diff
@@ -87,6 +88,9 @@ lib/optohybrid.so: src/optohybrid.cpp
 lib/calibration_routines.so: src/calibration_routines.cpp
 	$(CXX) $(CFLAGS) -std=c++1y -O3 -pthread $(INC) $(LDFLAGS) -fPIC -shared -Wl,-soname,calibration_routines.so -o $@ $< -lwisci2c -lxhal -llmdb -l:utils.so -l:extras.so -l:optohybrid.so -l:vfat3.so -l:amc.so
 
+lib/gbt.so: src/gbt.cpp include/gbt.h include/gbt_constants.h include/moduleapi.h include/memhub.h include/utils.h lib/utils.so
```
bdorney: I'm not sure I understand this make file line? Why is the target list here so long? In some cases these are already built.  The following line should be sufficient to ensure they are all linked correctly and they are found by the compiler.

include/gbt.h
```diff
@@ -0,0 +1,111 @@
...
+ *  The method expects the following RPC keys :
+ *  - `word ohN` : OptoHybrid index number.
+ *  - `word N` : The number of times the scan must performed.
+ *  - `word phaseMin` : Lowest phase to scan (min = 0).
+ *  - `word phaseMax` : Highest phase to scan (max = 15).
+ *  - `word phaseStep` : Step to scan the phases.
+ *
+ *  The method returns the following RPC keys :
+ *  - `word_array results` : Array of word containing the results of the scans. See details of the `scanGBTPhase` function for a description of the words.
+ *  - `string error` : If an error occurs, this keys exists and contains the error message.
```
bdorney, line 26: Are you lines 17 through 26 will be included in the `doxygen` build? Why aren't you using the `\details` and `\returns` fields?

include/gbt.h
```diff
@@ -0,0 +1,111 @@
...
+
+/*! \brief Local callable version of `scanGBTPhase`.
+ *  \param[in, out] la Local arguments structure.
+ *  \param[out] results A `std::vector` containing the results of the scan. See details for the items description.
+ *  \param[in] ohN OptoHybrid index number.
+ *  \param[in] N The number of times the scan must performed.
+ *  \param[in] phaseMin Lowest phase to scan (min = 0).
+ *  \param[in] phaseMax Highest phase to scan (max = 15).
+ *  \param[in] phaseStep Step to scan the phases.
+ *  \return Returns `false` in case of success; `true` in case of error. The precise error is logged and written to the `error` RPC key.
```
bdorney: If this is declared as `void` how can it return something?  This documentation doesn't make sense.

include/gbt.h
```diff
@@ -0,0 +1,111 @@
...
+ *
+ *  \detail
+ *
+ *  The scan seeks for valid RX phases for the VFAT's of one OptoHybrid. A phase is considered valid when `LINK_GOOD = 1`, `SYNC_ERR_CNT = 0` and `CFG_RUN != 0xdeaddead`. In order to improve the reliability of the scan, it is repeated `N` times.
+ *
+ *  Each item of the results vector is composed as follow :
+ *  - [31:28] : ohN
+ *  - [27:24] : phase
+ *  - [23:0] : result of the N scans (and) for the i^th VFAT. A `1` indicates a good phase while a `0` indicates a bad phase.
```
bdorney: From this documentation I do not understand how bits `[23:0]` should be interpreted.  Are you trying to say this should be interpreted as a 24 bit number where the i^th bit corresponds to the i^th VFAT?  

This was not really what was requested [here](https://github.com/cms-gem-daq-project/ctp7_modules/issues/72#issuecomment-441643357):

>For each GBT e-link (VFAT Pos) this should return in the RPC response a key which stores a word array where the array position corresponds to e-link phase and the array contents is the number of times that phase was "good."

I apologize if this was not clear but I was hoping more for something like the following:

- RPC Response contains for the X^th OH and Y^th GBT on this OH a key `OHX.GBTY`,
- This key returns a word array of 16 elements, and
- For the i^th element in this array this stores the number of time `M` out of scans `N` where the i^th phase was "good".

This is a bit easier to understand and unpack than what you have done here (and also follows very nicely the examples in daq_monitor).

include/gbt.h
bdorney: Apply this comment to all function documentation below.

include/gbt.h
bdorney: Apply this comment to all documentation below.

include/gbt.h
```diff
@@ -0,0 +1,111 @@
...
+ *
+ *  The length of the vector is determined by `phaseMin`, `phaseMax` and `phaseStep` : 
+ *  (`phaseMax` - `phaseMin` + 1)/`phaseStep`
+ */
+bool scanGBTPhasesLocal(localArgs *la, std::vector<uint32_t> &results, const uint32_t ohN, const uint32_t N = 1, const uint8_t phaseMin = gbt::phaseMin, const uint8_t phaseMax = gbt::phaseMax, const uint8_t phaseStep = 1);
```
bdorney: Why are we passing a vector by reference here?  Why not just have this return a vector and store this in the RPC response?  The `set_word_array` function can take a `std::vector<uint32_t>` input.

include/gbt.h
```diff
@@ -0,0 +1,111 @@
...
+ *  - `string error` : If an error occurs, this keys exists and contains the error message.
+ */
+void writeGBTConfig(const RPCMsg *request, RPCMsg *response);
+
+/*! \brief Local callable version of `writeGBTConfig`
+ *  \param[in, out] la Local arguments structure.
+ *  \param[in] ohN OptoHybrid index number.
+ *  \param[in] gbtN Index of the GBT to write. There 3 GBT's per OptoHybrid in the GE1/1 chambers.
+ *  \param[in] config Configuration blob of the GBT. This is a 366 elements long array whose each element is the value of one register sorted from address 0 to address 365.
+ *  \return Returns `false` in case of success; `true` in case of error. The precise error is logged and written to the `error` RPC key.
```
bdorney: If this is defined as `void` how can it return something? This documentation doesn't make sense.

include/gbt.h
```diff
@@ -0,0 +1,111 @@
...
+ */
+
+#ifndef GBT_H
+#define GBT_H
+
+#include "gbt_constants.h"
+
+#include "utils.h"
+
+/*! \brief Scan the GBT phases of one OptoHybrid.
```
bdorney, line 13: There is some confusion over this documentation; it's hard to follow which function this documentation block corresponds to.  You need to be using `doxygen` syntax and for each function the `\fn` field is expected at the *start* of the documentation block.  If you notice all other functions in this repo are organized like:

```C++
/*! \fn function declaration
     \brief text
     \param param1
     ...
     \returns text if returns something
     */
     function without any white space in between documentation and function
```

Inserting other `doxygen` fields where appropriate (e.g. `\details`).

Makefile
jsturdy: It looks like it is designed to determine what changes will trigger a rebuild of the `gbt.so` shared object; whether **all** of the targets are *necessary*, I can't say...

include/gbt.h
jsturdy: FYI, and I'll probably go through an overhaul and cleanup at some point in the (not near) future, the `\fn` declaration is *only* required if the documentation block does **not** immediately precede the function it's documenting (it doesn't hurt, but it adds unnecessary lines)

include/gbt_constants.h
```diff
@@ -0,0 +1,61 @@
...
+ *  \author Laurent Pétré <lpetre@ulb.ac.be>
+ */
+
+#ifndef GBT_CONSTANTS_H
+#define GBT_CONSTANTS_H
+
+#include <stdint.h>
+#include <array>
+
+namespace gbt{
```
bdorney: This namespace is great; but I was hoping this would be one generic hardware constant file as a central location. e.g. not `gbt_constants.h` but `hw_constants.h`.  Could you rename the file?

include/gbt_constants.h
```diff
@@ -0,0 +1,61 @@
...
+    /*! \brief The number of GBT's per OptoHybrid.
+     */
+    constexpr uint32_t gbtsPerOH = 3;
```
bdorney: This is specific to GE1/1; but we need a way to distinguish from GE2/1 and ME0.  I suggest an outer namespace, e.g. `GE11`.  The our namespaces will be:

```C++
GE11
GE11::gbt
GE11::gbt::elinkMappings
GE21
GE21::gbt
GE21::gbt::elinkMappings
```

Then we can reference these in an appropriate way later on.

src/gbt.cpp
```diff
@@ -0,0 +1,296 @@
...
+
+/*! \brief This macro is used to terminate a function if an error occurs. It logs the message, write it to the `error` RPC key and returns the `error_code` value.
+ *  \param response A pointer to the RPC response object.
+ *  \param message The `std::string` error message.
+ *  \param error_code Value which is passed to the `return` statement.
+ */
+#define EMIT_RPC_ERROR(response, message, error_code){ \
+    LOGGER->log_message(LogManager::ERROR, message); \
+    response->set_string("error", message); \
+    return error_code; }
```
bdorney: Seems like this would be a multi-use function and would be more suited to `utils.h`.

src/gbt.cpp
```diff
@@ -0,0 +1,296 @@
...
+        //  - [23:0] : VFAT i status
+        uint32_t result = ((ohN << 28) & 0xf0000000) |
+                          ((phase << 24) & 0x0f000000) |
+                          0x00ffffff;
+
+        // Set the new phases
+        for(uint32_t vfatN = 0; vfatN < gbt::vfatsPerOH; vfatN++){
+            if (writeGBTPhaseLocal(la, ohN, vfatN, phase))
+                return true;
+        }
```
bdorney, line 68: What is the purpose of lines 102 to 105?  Won't this potentially skip some vfats?

src/gbt.cpp
```diff
@@ -0,0 +1,296 @@
...
+
+    // Write the triplcated phase registers
+    const uint32_t gbtN = gbt::elinkMappings::vfatToGBT[vfatN];
+
+    for(unsigned char regN = 0; regN < 3; regN++){
+        const uint16_t regAddress = gbt::elinkMappings::elinkToRegisters[gbt::elinkMappings::vfatToElink[vfatN]][regN];
+
+        if (writeGBTRegLocal(la, ohN, gbtN, regAddress, phase))
+            return true;
+    }
```
bdorney, line 195: What's the purpose of lines 251 & 252?  Why return here?  Doesn't this risk skipping `regN = 1` or `regN = 2`?

bdorney (changes requested):
Phase scan tool doesn't meet output expectations.

include/gbt_constants.h
jsturdy: Naively I would actually suggest the opposite; there will certainly be constants which are subproject specific, but there would also likely be some which are the same across all subprojects.
In that case, the `gbt` namespace could house all the `gbt` related constants/`typedef`s, and then within that namespace, a more scoped set could be defined for each subproject separately.

I wouldn't go too far down this road without sketching out a class/namespace diagram to try to understand the various interconnected pieces related to this design though

I would, however, suggest an all-caps-with-underscores naming convention for any constants/`enums` being defined

include/gbt_constants.h
jsturdy: Depending on how many of these eventually get put in, having them in separate files is a nice clean way to keep them organized.
I'd apply the same comment I made later about sketching out a design for this before developing too much further

src/gbt.cpp
```diff
@@ -0,0 +1,296 @@
...
+        EMIT_RPC_ERROR(la->response, stdsprintf("The phaseMin parameter supplied (%hhu) is smaller than the minimal phase (%hhu).", phaseMin, gbt::phaseMin), true)
+    if (phaseMin > gbt::phaseMax)
+        EMIT_RPC_ERROR(la->response, stdsprintf("The phaseMin parameter supplied (%hhu) is bigger than the maximal phase (%hhu).", phaseMin, gbt::phaseMax), true)
+
+    // phaseMax check
+    if (phaseMax < gbt::phaseMin)
+        EMIT_RPC_ERROR(la->response, stdsprintf("The phaseMax parameter supplied (%hhu) is smaller than the minimal phase (%hhu).", phaseMax, gbt::phaseMin), true)
+    if (phaseMax > gbt::phaseMax)
+        EMIT_RPC_ERROR(la->response, stdsprintf("The phaseMax parameter supplied (%hhu) is bigger than the maximal phase (%hhu).", phaseMax, gbt::phaseMax), true)
+
```
jsturdy, line 61: It seems that lines 75-89 replicate lines 48-57 (ish), and would probably be better served by a generic GBT phase parameter checker function (or only do the check in the local function)

src/gbt.cpp
```diff
@@ -0,0 +1,296 @@
...
+    // ohN check
+    const uint32_t ohMax = readReg(la, "GEM_AMC.GEM_SYSTEM.CONFIG.NUM_OF_OH");
+    if (ohN >= ohMax)
+        EMIT_RPC_ERROR(la->response, stdsprintf("The ohN parameter supplied (%u) exceeds the number of OH's supported by the CTP7 (%u).", ohN, ohMax), true)
+
+    // gbtN check
+    if (gbtN >= gbt::gbtsPerOH)
+        EMIT_RPC_ERROR(la->response, stdsprintf("The gbtN parameter supplied (%u) exceeds the number of GBT's per OH (%u).", gbtN, gbt::gbtsPerOH), true)
+
```
jsturdy, line 139: Same comment as in `scanGBTPhase` viz, checking the arguments

src/gbt.cpp
```diff
@@ -0,0 +1,296 @@
...
+    // vfatN check
+    if (vfatN >= gbt::vfatsPerOH)
+        EMIT_RPC_ERROR(la->response, stdsprintf("The vfatN parameter supplied (%u) exceeds the number of VFAT's per OH (%u).", vfatN, gbt::vfatsPerOH), true)
+
+    // phase check
+    if (phase < gbt::phaseMin)
+        EMIT_RPC_ERROR(la->response, stdsprintf("The phase parameter supplied (%hhu) is smaller than the minimal phase (%hhu).", phase, gbt::phaseMin), true)
+    if (phase > gbt::phaseMax)
+        EMIT_RPC_ERROR(la->response, stdsprintf("The phase parameter supplied (%hhu) is bigger than the maximal phase (%hhu).", phase, gbt::phaseMax), true)
+
```
jsturdy, line 186: Again, parameter checking comment

jsturdy (changes requested):

Makefile
lpetre-ulb: Indeed, it is designed to trigger a rebuilt of `gbt.so` if one of the listed files change. Each of these files is directly required for a proper compilation of `gbt.so`. For example, if one brings modifications to `gbt_constants.h`, `make build` will trigger a compilation of `gbt.so`. It avoids to force a `make clean`.

include/gbt.h
lpetre-ulb: Yes, it is included in the `doxygen` build. The `\details` fields does not seem to allow the \`...\` syntax and the `\returns` field is not consistent with a `void` function. It simply appears as details (but not `\details`) under the parameters list.

include/gbt.h
lpetre-ulb: I pass the vector by reference because I want to be able to return the status of the scan (success/failure) and so the return value is already used. Ideally, I would have used exceptions and catch them in the RPC method, but it is not used anywhere else.

include/gbt_constants.h
lpetre-ulb: Yes, I will. See next comment for more details.

include/gbt_constants.h
lpetre-ulb: Here is a quick suggestion that I could immediately apply for this PR before walking through the code searching for a better solution and sketching some diagrams.

* Merge all the constants in a single file : `hw_constants.h`. If ever the file grows too much, it could be divided later.
* Separate the constants/`enum`'s/`typedef`'s in namespaces.
* If the values are different for each subproject, create a sub namespace, then inject the sub namespace values in the parent namespace at compile time.

The file would look like :

``` c++
// Could also be set as compiler parameter
#define GEM_VARIANT GE11

namespace gbt{
    constexpr uint16_t configSize = 366;
    ...

    namespace GE11{
        constexpr uint32_t gbtsPerOH = 3;
        ....
    }

    using namespace GEM_VARIANT;
}
```

Also, yes, I can capitalize the constants/`enum`'s, but wouldn't that be a too heavy name convention ?

src/gbt.cpp
lpetre-ulb: Indeed, I just wanted to be as little intrusive as possible.

src/gbt.cpp
lpetre-ulb: Indeed, that would be much better.

src/gbt.cpp
lpetre-ulb: Indeed some VFAT's could be skipped. But this happens only when the phase cannot be set properly. Therefore, the scan would be irrelevant anyway. So, I think, it is best to stop it as soon as possible and return the error.

src/gbt.cpp
lpetre-ulb: Yes, it will. But if the write of one register fails, the GBTX configuration is corrupted anyway. So, just stop there and report the error as soon as possible.

src/gbt.cpp
lpetre-ulb: Indeed, checking the arguments in a separate function would be more elegant. However, to keep some generality the gain in LOC would be minimal compared to the phase case. Also, that does not respect the coding style of the other modules. I can change that if really required (maybe for initiating a new coding style).

The checks of the arguments in both the "RPC" method and the local function follows some comments from [previous PR](https://github.com/cms-gem-daq-project/ctp7_modules/pull/73). See for example [this one](https://github.com/cms-gem-daq-project/ctp7_modules/pull/73#discussion_r238980114) where @bdorney told me that the local functions could be used from else where and so the arguments need to be checked.

include/gbt.h
lpetre-ulb: This is indeed what I'm trying to say. The `(and)` is a "typo". It should say (with some improvements) :
> [23:0] : results of the scan. This 24 bits vector is the _and_ of the N scans which were performed. The i^th bith correspond to the i^th VFAT with the following convention : a `1` indicates a good phase while a `0` indicates a bad phase. 

Ok, I did not understand it like that. I'm sorry. But would it be easier to unpack ? In my testing script, the unpacking is extremely simple; more than playing with strings concatenation I think. Each word of the array is self contained.

include/gbt.h
lpetre-ulb: Yes, this is why I did not write the `\fn` declaration. Also, it is really easier to maintain because we can change the function name, parameter ordering and types without manually adapting the documentation; `doxygen` in smart enough to find them.

@jsturdy, I'm not sure what your cleanup will be. Will you remove or add the `\fn` everywhere it is possible ? I think it is best if I follow what you will do.

jsturdy: > @jsturdy, I'm not sure what your cleanup will be. Will you remove or add the `\fn` everywhere it is possible?

I would remove `\fn` everywhere that it is not used as `doxygen` intends

include/gbt_constants.h
jsturdy: > If the values are different for each subproject, create a sub namespace, then inject the sub namespace values in the parent namespace at compile time.

I don't think you want to do this injection at compile time, since the same system will eventually be used to control, e.g., `GE11` *and* `GE21` at the same time

> Also, yes, I can capitalize the constants/enum's, but wouldn't that be a too heavy name convention ?

I'm not sure I understand the concern here, capitalized `consts` (`_` delimited, if that wasn't clear) is a fairly standard practice (probably coming from `C` land where this was done with `define`s), and we have adopted it in other GEM code, notably in `cmsgemos`.  The primary goal is to make clear that when used, the intent the expression is a constant

src/gbt.cpp
jsturdy: Are you writing an `error`/`warning` message to the logger when this happens?
I didn't trace things back, but here all I see is the `return true`

src/gbt.cpp
jsturdy: > Also, that does not respect the coding style of the other modules.

I'm not sure I know what the "coding style" is here (I'm not sure it's rigorously defined), but I'll chat with @mexanick and @bdorney and see if we can't all agree on some style standards that would be clear to all

> See for example this one where @bdorney told me that the local functions could be used from else where and so the arguments need to be checked.

I see, this was why I suggested that the check could be done *only* in the local function, as the RPC callback is just a passthrough, and the "local" function will be called in both cases

include/gbt.h
bdorney: > Yes, it is included in the doxygen build. 

Can you provide a url for this build in the `ctp7_modules` channel on mattermost and give the mattermost link here?  You can host it on the ULB machine if need be.

include/gbt.h
bdorney: > Ok, I did not understand it like that. I'm sorry. But would it be easier to unpack ? In my testing script, the unpacking is extremely simple; more than playing with strings concatenation I think. Each word of the array is self contained.

If I understand your function output the analysis tool has to loop over a vector with N element and then track the status of each (elink, ohN, GBT) triplet from a complex container.  And in this case I imagine N is large (i.e. N >> 100).

In my design the output looks atm for each (ohN, GBT) pair, a vector of 16 elements and immediately reads off the number of successes at each phase without further complication.  This second method is easier to process and understand (even by eye).

Please implement the requested change.

src/gbt.cpp
bdorney: Right but if someone wants to use this definition they now have to include `gbt.cpp` which would make the dependencies complicated.  The above comment should have been understood as:

> move this definition to `utils.h`.

src/gbt.cpp
bdorney: > Are you writing an `error`/`warning` message to the logger when this happens?
> I didn't trace things back, but here all I see is the `return true`

After some digging `scanGBTPhasesLocal(...)` calls `writeGBTPhaseLocal(...)` which will call `writeGBTRegLocal(...)` which will call `writeReg(...)`.  At each stage if an input parameter is not understood it looks like the `error` key is assigned to the RPC response and written to the CTP7 Log.  Then finally if an actual register transaction fails from `writeReg(...)` in `writeGBTRegLocal(...)` the corresponding bus error will be written to the log/rpc repsonse `error` key.

I am not sure I like this return true/false methodology you've adopted here as it makes things a bit harder to track; but off the top of my head I cannot think of a better approach except for exception handling.  I'll let it slide for now baring further comment from @jsturdy or @mexanick.

include/gbt.h
lpetre-ulb: Thank you for the clarification. I'll keep the documentation as it is then.

src/gbt.cpp
lpetre-ulb: Yes, of course, this was implicit in my previous answer.

src/gbt.cpp
bdorney: I see; after some discussion with @jsturdy we will open a separate issue to see if there's a way to have a cleaner implementation at a lower level to catch failed writes (and not just here).

So the current method here is good.  Thanks for the clarification.

src/gbt.cpp
bdorney: > I'm not sure I know what the "coding style" is here (I'm not sure it's rigorously defined), but I'll chat with @mexanick and @bdorney and see if we can't all agree on some style standards that would be clear to all

Yes we should have a defined standard.  I agree it is not present as overtime we improved things as we (I) gained experience.  In most cases @jsturdy usually has a better architecture design idea than I do.

src/gbt.cpp
lpetre-ulb: Yes, this is how it works.

I don't like much this methodology either (very C-style while the project is mostly C++), but exceptions are not used anywhere else. If I could use exceptions, the result would be nicer.

Makefile
bdorney: Got it.  Thanks for clarification.

src/gbt.cpp
lpetre-ulb: > I see, this was why I suggested that the check could be done _only_ in the local function, as the RPC callback is just a passthrough, and the "local" function will be called in both cases

Yes, I'll do that where possible.

There is also the check for `ohMax` which is special because it requires a "slow" register access. Maybe, should I implement some easy cache system ? If ever the value of the register changes, it means that the CTP7 firmware has changed and so the RPC process is reloaded anyway.

include/gbt_constants.h
lpetre-ulb: > > If the values are different for each subproject, create a sub namespace, then inject the sub namespace values in the parent namespace at compile time.
> 
> I don't think you want to do this injection at compile time, since the same system will eventually be used to control, e.g., `GE11` _and_ `GE21` at the same time

So, how should I refer to the constants in the code ? If it explicitly refers to the `GE11` or `GE21` namespace, the purpose of clearly separating the constants is partly defeated. Or maybe that some mechanism will load the "constants" at run-time ?

Also, will `GE11` and `GE21` be mixed within the same CTP7/ATCA board ? If so, the RPC modules will be more complex in order to determine the variant on-the-fly for each link.

> > Also, yes, I can capitalize the constants/enum's, but wouldn't that be a too heavy name convention ?
> 
> I'm not sure I understand the concern here, capitalized `consts` (`_` delimited, if that wasn't clear) is a fairly standard practice (probably coming from `C` land where this was done with `define`s), and we have adopted it in other GEM code, notably in `cmsgemos`. The primary goal is to make clear that when used, the intent the expression is a constant

Yes, I'll modify the constant names (with `_` delimiters).

This is however a controversial subject. Personally, I prefer (in C/C++) to reserve the capitalized names to preprocessor define's/macro's where the C/C++ syntax may not be respected. The name then shouts for a good reason. As for the constants, there are in a namespace, so the `::` already raises an alert and I think about what I'm going to do.

include/gbt.h
lpetre-ulb: > If I understand your function output the analysis tool has to loop over a vector with N element and then track the status of each (elink, ohN, GBT) triplet from a complex container. And in this case I imagine N is large (i.e. N >> 100).

Hmm, no, this is not what in my implementation returns. The results are already merged in the RPC module, so you only receive the status of the phase (GOOD/BAD) after the N scans; if it fails once, the phase is considered BAD. So whatever N is, you receive 16 words (1 per phase) each of them being self-contained; there is no need to keep track of anything, just save the result.

> In my design the output looks atm for each (ohN, GBT) pair, a vector of 16 elements and immediately reads off the number of successes at each phase without further complication. This second method is easier to process and understand (even by eye).
> 
> Please implement the requested change.

I'll do.

include/gbt.h
lpetre-ulb: Here it is : https://mattermost.web.cern.ch/cms-gem-daq/pl/jqm787toyfbpfjghsyded13bbw

include/gbt_constants.h
jsturdy: > So, how should I refer to the constants in the code ? If it explicitly refers to the `GE11` or `GE21` `namespace`, the purpose of clearly separating the constants is partly defeated. Or maybe that some mechanism will load the "constants" at run-time?

Actually I glossed over this earlier, but you're right: either there needs to be a run time determination of this (easy to separate `GE11`/`GE21`/`ME0` within a given back-end card, less so if the code is running on a development platform that doesn't have access to expensive back-end cards and the methods are residing on a PC), **OR** the modules are compiled for their target system (and your current implementation is mostly sufficient, passing in the `GEM_VARIANT` as a compile-time option, rather than as a `define` in some header)

As I mentioned earlier, I don't think that fleshing out this complete design should be a part of this PR, as it will require quite some careful thought about the implementation, so I would say just keep what you have now, and add the compiler flag `-DGEM_VARIANT=\"${GEM_VARIANT}\"` that should be `GE11` by default now (well `ge11` if I'm going to nitpick, as I would make all namespaces lower case)

> This is however a controversial subject

**All** style based subjects are :-D

src/gbt.cpp
jsturdy: Reading that one register is an inexpensive transaction at this level, I wouldn't change things just to avoid that.

include/gbt_constants.h
bdorney: > > So, how should I refer to the constants in the code ? If it explicitly refers to the `GE11` or `GE21` `namespace`, the purpose of clearly separating the constants is partly defeated. Or maybe that some mechanism will load the "constants" at run-time?
> 
> Actually I glossed over this earlier, but you're right: either there needs to be a run time determination of this (easy to separate `GE11`/`GE21`/`ME0` within a given back-end card, less so if the code is running on a development platform that doesn't have access to expensive back-end cards and the methods are residing on a PC), **OR** the modules are compiled for their target system (and your current implementation is mostly sufficient, passing in the `GEM_VARIANT` as a compile-time option, rather than as a `define` in some header)

If we want a runtime variant we could again rely on the uTCA AMC/aTCA blade firmware version.  From @evka85 we know GE11 and GE21 will have different CTP7 FW versions so this naturally could be encoded into the software like it was presently done for v2b vs. v3 functionality.  This could then be used to determine the functionality at runtime.  Just my two cents.

src/gbt.cpp
bdorney: Is there a reason why we can't implement exception handling in `ctp7_modules` repo? @mexanick

src/gbt.cpp
lpetre-ulb: It is inexpensive, but not negligible. For example, if the `ohMax` check is done in `writeGBTRegLocal`, writing a GBT configuration would require 366 accesses to `GEM_AMC.GEM_SYSTEM.CONFIG.NUM_OF_OH`, 365 of them being unnecessary. Considering a register access taking ~10µs, configuring one GBT is ~3.6ms longer for nothing. If one wants to write all the GBT's of one CTP7, that makes more than useless 100ms.

The cache mechanism could be very simple and integrated within the new check functions :
```c++
inline bool checkOhN(localArgs *la, const uint32_t ohN){
    static uint32_t ohMax = -1;
    if (ohMax == -1)
        ohMax = readReg(&la, "GEM_AMC.GEM_SYSTEM.CONFIG.NUM_OF_OH");
        // Maybe a check on readReg ?

    if (ohN >= ohMax)
        EMIT_RPC_ERROR(la->response, stdsprintf("The ohN parameter supplied (%u) exceeds the number of OH's supported by the CTP7 (%u).", ohN, ohMax), true) 
}
```
It is a way of "defining a constant at run-time with information from the CTP7/back-end board".

src/gbt.cpp
jsturdy: It *can* be done, but it has to be done *right*, and this involves deciding what `throw` semantics each class of function should employ, i.e, you probably *don't* want the RPC callbacks to `throw`, unless the `rpcsvc` can `catch` and handle those exceptions.
It would make sense that "local" functions can `throw`, and the RPC callback (or external code) would be responsible for handling and returning data in a good state
This is *definitely* not within the scope of this PR, so should be moved to an issue and refined there

src/gbt.cpp
jsturdy: OK, in that case your proposed solution makes sense (though I'd think long and hard about whether `writeGBTRegLocal` should be a public interface function in this case, or should only be used internally by other functions that would do the sanity checks themselves...)

One could also consider what other FW version specific constants could benefit being exposed this way

lpetre-ulb:
I brought the requested changes.

Regarding the [discussion about `ohN` parameter check](https://github.com/cms-gem-daq-project/ctp7_modules/pull/75#discussion_r241608923), I haven't added the method yet and kept the warning in `writeRegLocal`. A more careful reflection should be made for parameter checking and the utility of a cache. Also, the gain in code is rather small (except for phases) if exceptions are not used; indeed, the return value of each check function call must must checked itself in order to terminate the function if needed.

More reflection should be made on constants, parameter checking and exceptions. These can (should ?) however be addressed in other issues and PR's.

src/gbt.cpp
```diff
@@ -34,77 +26,48 @@ void scanGBTPhases(const RPCMsg *request, RPCMsg *response){
...
     if (ohN >= ohMax)
         EMIT_RPC_ERROR(la->response, stdsprintf("The ohN parameter supplied (%u) exceeds the number of OH's supported by the CTP7 (%u).", ohN, ohMax), true)
 
     // phaseMin check
-    if (phaseMin < gbt::phaseMin)
-        EMIT_RPC_ERROR(la->response, stdsprintf("The phaseMin parameter supplied (%hhu) is smaller than the minimal phase (%hhu).", phaseMin, gbt::phaseMin), true)
-    if (phaseMin > gbt::phaseMax)
-        EMIT_RPC_ERROR(la->response, stdsprintf("The phaseMin parameter supplied (%hhu) is bigger than the maximal phase (%hhu).", phaseMin, gbt::phaseMax), true)
+    if (gbt::checkPhase(la->response, phaseMin))
+        return true;
```
jsturdy, line 53: I think you either want a check and action on `false` (definitely not `return true` here) (also L57), don/t you?

src/gbt.cpp
```diff
@@ -143,21 +107,14 @@ void writeGBTConfig(const RPCMsg *request, RPCMsg *response){
...
-    if (gbtN >= gbt::gbtsPerOH)
-        EMIT_RPC_ERROR(response, stdsprintf("The gbtN parameter supplied (%u) exceeds the number of GBT's per OH (%u).", gbtN, gbt::gbtsPerOH),)
 
-    // config key
+    // We must check the size of the config key
     const uint32_t configSize = request->get_binarydata_size("config");
-    if (configSize != gbt::configSize)
-        EMIT_RPC_ERROR(response, stdsprintf("The provided configuration has not the correct size. It is %u registers long while this methods expects %hu 8-bits registers.", configSize, gbt::configSize), );
+    if (configSize != gbt::CONFIG_SIZE)
+        EMIT_RPC_ERROR(response, stdsprintf("The provided configuration has not the correct size. It is %u registers long while this methods expects %hu 8-bits registers.", configSize, gbt::CONFIG_SIZE), );
```
jsturdy: There's a trailing comma, and potentially missing final parameter, though I may have glossed over the usage of `EMIT_RPC_ERROR`...
If you're returning `void`, I'd make it explicit.

Makefile
```diff
@@ -26,6 +26,8 @@ IncludeDirs += ${BUILD_HOME}/xhal/xhalcore/include
 #IncludeDirs += /opt/cactus/include
 INC=$(IncludeDirs:%=-I%)
 
+CFLAGS += -DGEM_VARIANT="ge11"
```
jsturdy: This is fine for now, but it should pick "ge11" (or other) up from the `make` environment, not from the `Makefile` (unless there will be a separate `Makefile` for all variants

jsturdy (changes requested):
Couple of tweaks/clarifications

src/gbt.cpp
lpetre-ulb: No, I definitely want to return `true`. The contract for the local functions is simple : return `true` in case of error and write enough debugging informations (i.e. `LOGGER` and RPC key `error`). The convention may not be very well chosen in this case (it comes from the historical C world with the `exit()` function), but I think it is best to use a uniform convention.

src/gbt.cpp
lpetre-ulb: Indeed, it returns `void`; the last parameter of the `EMIT_RPC_ERROR` macro can be empty (this is why I like to use capitals for macros). In the same idea, the trailing semi-colon is not necessary but does not hurt. I could pass `(void)""` as the last parameter to make it explicit.

Makefile
lpetre-ulb: Yes that would be better, but as the variable does not exist yet I chose this simpler solution. I'll change it to do a check on `GEM_VARIANT` existence and default back to `ge11`.

jsturdy:
@lpetre-ulb, @bdorney, coming back to this, was there anything outstanding in this PR that we *didn't* decide to postpone to a later PR?
In either case, @lpetre-ulb, can you `rebase` this onto the current `develop` branch?

jsturdy (dismissed):

src/gbt.cpp
mexanick: I stand with @jsturdy. For the sake of readability, I would add a void last parameter, or, even better, define the error code table and return an actual error code

mexanick (dismissed):

src/gbt.cpp
lpetre-ulb: Sure, I forgot with the Christmas break... Done ! Yes, an actual error code would be better (or maybe exceptions). However, in this precise case, this is the RPC function which must return `void`.

lpetre-ulb:
> @lpetre-ulb, @bdorney, coming back to this, was there anything outstanding in this PR that we _didn't_ decide to postpone to a later PR?

I don't think there is. Registers cache (could also be useful for VFATs chipID with error correcting code), exceptions and multi-link versions are postponed to later PR's.

> In either case, @lpetre-ulb, can you `rebase` this onto the current `develop` branch?

Done !

jsturdy (dismissed):

jsturdy:
@bdorney any outstanding issues from you? I see this is blocked by "changes requested"

bdorney:
Let me do a cursory review this afternoon/evening.  I see most of my comments are marked with `outdated` so I assume they have been addressed...but still, I'll have a look this afternoon.

Makefile
```diff
@@ -87,6 +94,9 @@ lib/optohybrid.so: src/optohybrid.cpp
 lib/calibration_routines.so: src/calibration_routines.cpp
 	$(CXX) $(CFLAGS) -std=c++1y -O3 -pthread $(INC) $(LDFLAGS) -fPIC -shared -Wl,-soname,calibration_routines.so -o $@ $< -lwisci2c -lxhal -llmdb -l:utils.so -l:extras.so -l:optohybrid.so -l:vfat3.so -l:amc.so
 
+lib/gbt.so: src/gbt.cpp include/gbt.h include/hw_constants.h include/hw_constants_checks.h include/moduleapi.h include/memhub.h include/utils.h lib/utils.so
+	$(CXX) $(CFLAGS) -std=c++1y -O3 -pthread $(INC) $(LDFLAGS) -fPIC -shared -Wl,-soname,gbt.so -o $@ $< -l:memhub.so -l:utils.so
+
```
bdorney, line 99: One comment is that "lower level" so's should be built before higher level so's.  But in this case since presently `gbt.so` is not used by any of the others it doesn't matter.  Just keep in mind for future (e.g. if you built `amc.so` last you'd have a bad day).

include/gbt.h
```diff
@@ -0,0 +1,107 @@
...
+void scanGBTPhases(const RPCMsg *request, RPCMsg *response);
+
+/*! \brief Local callable version of `scanGBTPhase`.
+ *  \param[in, out] la Local arguments structure.
+ *  \param[in] ohN OptoHybrid index number.
+ *  \param[in] N The number of times the scan must performed.
+ *  \param[in] phaseMin Lowest phase to scan (min = 0).
+ *  \param[in] phaseMax Highest phase to scan (max = 15).
+ *  \param[in] phaseStep Step to scan the phases.
+ *  \return Returns `false` in case of success; `true` in case of error. The precise error is logged and written to the `error` RPC key.
```
bdorney, line 37: General comment is that I think in addition to setting the `error` key it might be good to one day develop exception handling and such for these RPC functions.  That is beyond the scope of this PR though.

Makefile
jsturdy: Yeah, this is improved significantly in #77, which will be rebased and merged after these changes are merged.

bdorney (dismissed):

lpetre-ulb:
Before merging, let me push one more commit that I am currently testing about the delays for the phase scan (very minimal modification). It should only take a few tens of minutes, but it would speed up the phase scan by a lot.

bdorney (approved):
last commit introduced a trivial change. still good.

lpetre-ulb:
As @bdorney suggested in the previous PR (https://github.com/cms-gem-daq-project/ctp7_modules/pull/73#discussion_r238976795), the wait times in the phase scan were *very long*. @evka85 has explained me how he choose these values and how much they can be reduce :
> I just picked these delays somewhat arbitrarily, and they are definitely overkill.
> The first one (100ms) I think was to give time for the GBTX to do the phase shift or something like that, but in fact I think this is not really needed at all.
The second one is to give time for the LINK_GOOD flag to initialize after a link reset. This flag works like this:
> * every orbit (every ~90us) the sync verify command is sent to VFATs, which then reply with a "sync good" message if they receive the verify command.
> * for the LINK_GOOD to go high, it has to see 10 "sync good" messages from the VFATs in a row (in other words, the communication with the VFAT has to be good for 10 consecutive orbits). After the LINK_GOOD is high, it can only go low if there's 100 orbits without "sync_good" messages from VFATs.
> 
> So given this behavior of the LINK_GOOD flag, there must be a delay of at least 10 orbits (around 1ms) after the link reset. So in principle the 300ms could be reduced to 1ms, but I would give it some wiggle room and just wait for 5 or 10ms or so.

Following his advise, I picked 10ms of wait time. After a few tests, the phase scan still works seamlessly and so one can launch a scan with a bigger number of repetitions without waiting "forever".

jsturdy (commented):
~~Was there a reason these delays were set to 300ms before?~~
@lpetre-ulb @bdorney

Too slow, OK

jsturdy (approved):

lpetre-ulb:
> Was there a reason these delays were set to 300ms before?
> @lpetre-ulb @bdorney

I copied these from @evka85's original `gbt.py` script, but as he wrote in his email, he "picked these delays somewhat arbitrarily, and they are definitely overkill".

PR #76, opened 2018-12-14
Hot Fix: Added ADC Cached Registers w/backwards compatibility

bdorney:
## Description

After CTP7 FW version [3.7.1](https://github.com/evka85/GEM_AMC/releases/tag/v3.7.1) the `ADCX` registers have been converted to `ADCX_CACHED` and `ADCX_UPDATE` for `X = 0,1`. 

I've added a backwards compatible change to check if `ADCX_CACHED` exists, and if so it is used; otherwise `ADCX` will be used.

### Types of changes

## Motivation and Context

See discussion above.

## How Has This Been Tested?

Ran temperature monitoring measurements:
http://cmsonline.cern.ch/cms-elog/1077701

While VFAT info is not shown in the output it is *always* written to the output file.

## Checklist:


src/vfat3.cpp
```diff
@@ -260,12 +260,31 @@ void getChannelRegistersVFAT3Local(localArgs *la, uint32_t ohN, uint32_t vfatMas
...
-    } //End Case: Use ADC with external reference
-    else{ //Case: Use ADC with internal reference
-        broadcastReadLocal(la, outData, ohN, "ADC0", mask);
-    } //End Case: Use ADC with internal reference
+    //Hack
+    uint32_t fwMajor = readReg(la,"GEM_AMC.GEM_SYSTEM.RELEASE.MAJOR");
+    uint32_t fwMinor = readReg(la,"GEM_AMC.GEM_SYSTEM.RELEASE.MINOR");
+    uint32_t fwBuild = readReg(la,"GEM_AMC.GEM_SYSTEM.RELEASE.BUILD");
+
+    if( fwMajor >= 3 && fwMinor >= 7 && fwBuild >= 1 )
```
bdorney, line 268: I'm not sure if this is the best approach.  The method used here:

https://github.com/cms-gem-daq-project/ctp7_modules/blob/9911586985520c6ad5c2e6206708ccf7c8f13f24/src/calibration_routines.cpp#L1176-L1184

Is applicable since the above is a broadcast read and not a single chip read.  If anyone has a better idea on the implementation let me know

src/vfat3.cpp
```diff
@@ -260,12 +260,31 @@ void getChannelRegistersVFAT3Local(localArgs *la, uint32_t ohN, uint32_t vfatMas
...
-    if(useExtRefADC){ //Case: Use ADC with external reference
-        broadcastReadLocal(la, outData, ohN, "ADC1", mask);
-    } //End Case: Use ADC with external reference
-    else{ //Case: Use ADC with internal reference
-        broadcastReadLocal(la, outData, ohN, "ADC0", mask);
-    } //End Case: Use ADC with internal reference
+    //Hack
+    uint32_t fwMajor = readReg(la,"GEM_AMC.GEM_SYSTEM.RELEASE.MAJOR");
+    uint32_t fwMinor = readReg(la,"GEM_AMC.GEM_SYSTEM.RELEASE.MINOR");
+    uint32_t fwBuild = readReg(la,"GEM_AMC.GEM_SYSTEM.RELEASE.BUILD");
```
jsturdy, line 266: Is the FW version a single readable register?
If so, you can do just one read and unpack in the check function (or create a `FWVersion` object that has comparison operators defined)

src/vfat3.cpp
jsturdy: > Is applicable since the above is a broadcast read and not a single chip read. If anyone has a better idea on the implementation let me know

This statement/question is not clear; you are asking if the linked check would be sufficient to replace the `fw` check implemented in L268?  You're asking if the cached version checking is appropriate for a "broadcast" type read?

src/vfat3.cpp
bdorney: Sadly unlike the `1.X.Y` series there is no single readable register.  However @evka85 may be able to implement this in a future PR.

src/vfat3.cpp
bdorney: > > Is applicable since the above is a broadcast read and not a single chip read. If anyone has a better idea on the implementation let me know
> 
> This statement/question is not clear; you are asking if the linked check would be sufficient to replace the `fw` check implemented in L268? You're asking if the cached version checking is appropriate for a "broadcast" type read?

I'm wondering if there is a better solution than line 268.  I originally wanted to do something like lines 1176 to 1184 but this action is hidden inside a broadcast read so I cannot simply check if the register exists.  I could in principle check if the first unmasked VFAT in the mask has that register existing...and then do the broadcast read...but that too seems "clunky".

Feedback would be appreciated.

src/vfat3.cpp
jsturdy:
```xml
      <node id="RELEASE" address="0x3">
        <node id="MAJOR" address="0x0" mask="0x00ff0000" permission="r"
              description="Firmware release major version"
              fw_signal="std_logic_vector(to_unsigned(version_major, 8))"/>
        <node id="MINOR" address="0x0" mask="0x0000ff00" permission="r"
              description="Firmware release minor version"
              fw_signal="std_logic_vector(to_unsigned(version_minor, 8))"/>
        <node id="BUILD" address="0x0" mask="0x000000ff" permission="r"
              description="Firmware release build version"
              fw_signal="std_logic_vector(to_unsigned(version_build, 8))"/>
        <node id="DATE" address="0x1" permission="r"
              description="Firmware release date (read as hex)"
              fw_signal="firmware_date"/>
</node>
```
So all you need to do is get the **address** of *any* of the version fields (i.e., not the `DATE` field), and do a single read on that address.  Then you extract the corresponding bits.
Though the bit mapping may be changing, so I guess this operation can wait until that's been finalized...

src/vfat3.cpp
jsturdy: If you want, you *could* check for the existence of the `_CACHED` register name at the highest (a higher?) level, and set a flag `use_cached` which drives the functionality later.
 * That could even be done in the linked code, which is doing this check inside a loop over VFATs, and clearly **they** won't ever have different behaviour
 * Even more so, this check needs only to be done *once* in **any** routine, since it is AMC wide, i.e., different front end links won't have different behaviour

src/vfat3.cpp
jsturdy: You could further define a struct to hold the version information and define comparison operators on this

bdorney: But who should own that struct and where should it be initialized/filled/compared?

Anytime you wanted to read the VFAT ADC it seems like it would slow down everything to constantly read the FW version, parse it, and compare it.  Okay I guess at most it's one additional read plus some parsing but still that would incorporate a slow down.

Moreover...are we ever going to go backwards with the FW?  Perhaps this whole PR is a m00t point and we should just be using the cached & update registers from now on.

src/vfat3.cpp
bdorney: I guess that would done outside this function and a `use_cached` argument should be implemented to control the behavior.  But before doing this we should clarify if we are ever going to support the backwards FW or not.

In 3.7.X or up the trigger data is moved to the miniPOD connector.  Right now only one test stand has this routed correctly and all other test stands are assuming trigger data comes in on the CXP2 connector.  This is probably not a problem...would just need to order fibers but there are some cases/tests that *need* this.  So keeping a 3 CXP+MiniPod FW version and a 2 CXP (GBT+Trig) version may be advantageous.  I just don't know how difficult this is to maintain for @evka85

bdorney:
Superceded by https://github.com/cms-gem-daq-project/ctp7_modules/pull/91

PR #77, opened 2019-01-07
cmsgemos feature port for AMC

jsturdy:
## Description
Start porting functionality from `cmsgemos` to `ctp7_modules`
* Includes functionality from `HwGenericAMC`
* Includes some reworking of the `Makefile` for adapting to source -> objects -> shared objects make process
  * Multiple objects built into single shared object (`CTP7` "module")
  * Retain current number of "modules", but allow distinct features to be scoped per file, e.g., `TTC`, DAQ, etc.
* Further functionality will be added in a separate PR for the OH specific code and adaptations for the blaster interface

### Types of changes

## Motivation and Context
Starting the necessary movement to have the core `cmsgemos` SW compatible with the RPC architecture

## How Has This Been Tested?
* Libraries compiled
* Copied to CTP7
* Test function executed
* Final testing needed, but machine failure in 904 prevented this

### Testing notes
* The *exact* version of the linked libraries matters, i.e., @mexanick and I discovered a problem that was due to the module being compiled against a version of the `xhal` library that was not on the card.
* At runtime this resulted in undefined behaviour due to the change in the LMDB structure that was not present.
* The interplay of all this needs to be finalized, with the versioning of the libraries and installed packages on the CTP7 (work in progress with `smart` package manager)

## Checklist:
- [X] I have ~added~ *run* tests to cover my changes.


include/utils.h
```diff
@@ -21,16 +23,34 @@
 #include <iterator>
 #include <cstdio>
 
-memsvc_handle_t memsvc; /// \var global memory service handle required for registers read/write operations
+extern memsvc_handle_t memsvc; /// \var global memory service handle required for registers read/write operations
```
mexanick, line 26: Where this variable is defined?

include/utils.h
```diff
@@ -21,16 +23,34 @@
...
     RPCMsg *response; /*!< RPC response message */
-};
+} LocalArgs;
+
+/*!
+ * \brief returns a set up LocalArgs structure
+ */
+LocalArgs getLocalArgs(RPCMsg *response);
+
+// FIXME: to be replaced with the above function when the struct is properly implemented
```
mexanick, line 42: Is the macro below tested and proven to be working? I personally don't see much difference with the previous attempt to implement a function, which failed to allocate the memory correctly...

jsturdy: The failure was in the return of the pointer to the struct, I believe, so the macro just wraps the call until one can get the setup function to work properly.  The macro behaves no differently than having the lines that are currently in every other callback

include/utils.h
jsturdy: `utils.cpp`

mexanick (dismissed):

jsturdy:
I may prefer to finally merge this once #75 is merged, as I expect the changes in the `Makefile` to be conflicting, and I'll need to sort them out after rebasing this branch on top of those changes

jsturdy:
Rebased following the merge of #75, ready for re-review
@mexanick @bdorney

bdorney:
Before this is merged could we have the current branch of `develop` c323274d9344ebb922500832a53c9a2af11eede4 set as a legacy branch.  It will take some time for the python API to be rebuilt and the test suite to be updated and so the sustained operation measurements (and maybe some QC8 things) will need a "stable" point to work off of in the short term (for `$USER`s) while the developers update the toolset.

mexanick:
#79 should do this. Please review

mexanick (dismissed):
Cursory approval, as the multiple changes requires hardware testing/debugging

src/amc/daq.cpp
```diff
@@ -0,0 +1,524 @@
...
+  disableDAQLinkLocal(&la);
+}
+
+void setZS(const RPCMsg *request, RPCMsg *response)
+{
+  // struct localArgs la = getLocalArgs(response);
+  GETLOCALARGS(response);
+  bool enable = request->get_word("enable");
+  setZSLocal(&la,enable);
+}
```
bdorney, line 245: So I'm a bit confused by methods of this nature.  They appear to call a local method which in theory is going to perform just 1 (or a few) read/write transaction.  This will come in with a huge time penalty due to the TCP/IP transaction time.

I thought the goal was to use uhal for these "simple" actions and then rpc methods for register intensive actions.  The local methods above seem to make sense (as they could be called by some "master" RPC function to do some complex task, but having a remote method that just does one thing seems strange).

include/amc/ttc.h
```diff
@@ -0,0 +1,165 @@
+/*!
+ * \file amc/ttc.h
```
bdorney, line 2: If this is supposed to be the home of `TTC` modules shouldn't we migrate those functions in `calibration_routines.*` that deal with `TTC.GENERATOR` to this file?

src/daq_monitor.cpp
```diff
@@ -141,7 +141,7 @@ void getmonTRIGGEROHmain(const RPCMsg *request, RPCMsg *response)
   auto rtxn = lmdb::txn::begin(env, nullptr, MDB_RDONLY);
   auto dbi = lmdb::dbi::open(rtxn, nullptr);
 
-  struct localArgs la = {.rtxn = rtxn, .dbi = dbi, .response = response};  
+  struct localArgs la = {.rtxn = rtxn, .dbi = dbi, .response = response};
```
bdorney, line 144: this will eventually be replaced by `GETLOCALARGS`?

src/utils.cpp
```diff
@@ -1,5 +1,34 @@
...
+    std::string gem_path       = std::getenv("GEM_PATH");
+    std::string lmdb_data_file = gem_path+"/address_table.mdb";
+    env.open(lmdb_data_file.c_str(), 0, 0664);
+    auto rtxn = lmdb::txn::begin(env, nullptr, MDB_RDONLY);
+    auto dbi  = lmdb::dbi::open(rtxn, nullptr);
+    struct localArgs la = {.rtxn     = rtxn,
+                           .dbi      = dbi,
+                           .response = response};
+    return la;
+}
```
bdorney, line 18: Who is supposed to be doing this? `GETLOCALARGS` or this function?  I saw some discussion above.  Not sure if it's appropriate to discuss there or here.

src/amc/daq.cpp
jsturdy: Current implementation was a direct copy/paste from `HwGenericAMC`.
I fully expect to not utilize these single transaction RPC methods anywhere and to remove them during a future cleanup.

include/amc/ttc.h
jsturdy: Almost definitely (cf. future cleanup)

src/daq_monitor.cpp
jsturdy: No legacy code was modified, but, yes, during the aforementioned cleanup, all these will be updated.

src/utils.cpp
jsturdy: This function, however current usage and implementation do not work, so the macro is the cleanup placeholder

include/amc/ttc.h
jsturdy: And in that case, I'd probably block #63 and redo that as part of said cleanup

bdorney: I would be fine with that.  The git commit history that #63 introduces is somewhat messy (multiple commits adding/removing the same text).  Might want to make a comment there to @AndrewLevin then.

jsturdy:
OK, @mexanick, @bdorney, I'm currently satisfied with the content of this PR and won't push anything other than requested fixes.
Blaster interface and VFAT3 Reed--Muller ChipID grow from the tip of this PR

Makefile
```diff
@@ -56,50 +45,192 @@ default:
...
+# lib/calibration_routines.so: src/calibration_routines.cpp
+# 	$(CXX) $(CFLAGS) -std=c++1y -O3 -pthread $(INC) $(LDFLAGS) -fPIC -shared -Wl,-soname,calibration_routines.so -o $@ $< -lwisci2c -lxhal -llmdb -l:utils.so -l:extras.so -l:optohybrid.so -l:vfat3.so -l:amc.so
+
+# lib/gbt.so: src/gbt.cpp include/gbt.h include/hw_constants.h include/hw_constants_checks.h include/moduleapi.h include/memhub.h include/utils.h lib/utils.so
+# 	$(CXX) $(CFLAGS) -std=c++1y -O3 -pthread $(INC) $(LDFLAGS) -fPIC -shared -Wl,-soname,gbt.so -o $@ $< -l:memhub.so -l:utils.so
+
+.PHONY: test
+### local (PC) test functions, need standard gcc toolchain, dirs, and flags
+test: test/tester.cpp
+	g++ -O0 -g3 -fno-inline -o test/$@ $< $(INC) $(LDFLAGS) -lwiscrpcsvc
```
mexanick, line 227: If the new way of compilation is properly tested, I would clean up the commented part here, starting from line 143. If not - we need to test and fix it =)

src/utils.cpp
```diff
@@ -249,25 +277,36 @@ uint32_t applyMask(uint32_t data, uint32_t mask) {
 }
 
 uint32_t readReg(localArgs * la, const std::string & regName) {
+  LOGGER->log_message(LogManager::WARNING, stdsprintf("Called readReg on %s", regName.c_str()));
```
mexanick: Do we really want all these log messages till l309 to be of "WARNING" level? I would make them rather "DEBUG" or even "TRACE"

mexanick (changes requested):

src/amc/sca.cpp
```diff
@@ -0,0 +1,23 @@
+/*!
+ * \file amc/sca.cpp
+ * \brief AMC SCA methods for RPC modules
+ */
+
+#include "amc/sca.h"
+
+void scaModuleResetLocal(localArgs* la)
+{
+  writeReg(la, "GEM_AMC.SLOW_CONTROL.SCA.CTRL.MODULE_RESET", 0x1);
```
bdorney, line 10: This will cause a reset to all links; do we want to either add to this function a mask to be used with `GEM_AMC.SLOW_CONTROL.SCA.CTRL.SCA_RESET_ENABLE_MASK` or do we want to add a separate function to set this mask?  Probably the latter and not the former.

src/utils.cpp
jsturdy: Oops, in fact, these should all be removed I think, we tracked down the issue that I was trying to debug with them, but must have accidentally committed them

src/amc/ttc.cpp
```diff
@@ -0,0 +1,907 @@
...
+#include "amc/ttc.h"
+
+#include <ios>
+#include <chrono>
+#include <thread>
+#include <iomanip>
+
+void ttcModuleResetLocal(localArgs* la)
+{
+  // writeReg(la, "GEM_AMC.TTC.CTRL.MODULE_RESET", 0x1);
```
bdorney, line 15: Why comment this out?

Also what about the associated TTC Generator functions presently in `calibration_routines`.  Shouldn't they be moved here/expanded upon?

Makefile
jsturdy: yeah, this is further cleaned up in [this](https://github.com/jsturdy/ctp7_modules/compare/feature/hw-feature-port...jsturdy:feature/reedmuller-chipid) branch, where the Reed--Muller additions are coming

src/utils.cpp
bdorney: Indeed this will fill the log file with *every* read transaction...

src/utils.cpp
```diff
@@ -249,25 +277,36 @@ uint32_t applyMask(uint32_t data, uint32_t mask) {
...
     	//response->set_string("error", std::string("No read permissions"));
     	LOGGER->log_message(LogManager::ERROR, stdsprintf("No read permissions for %s", regName.c_str()));
       return 0xdeaddead;
     }
     uint32_t data[1];
     uint32_t address,mask;
     address = stoll(tmp[0]);
+    LOGGER->log_message(LogManager::WARNING, stdsprintf("Address obtained 0x%x", address));
     mask = stoll(tmp[2]);
+    LOGGER->log_message(LogManager::WARNING, stdsprintf("Mask obtained 0x%x", mask));
```
bdorney: Same comment as above, these should probably not be at the `WARNING` level.  Maybe `INFO` but most likely `DEBUG` or below.

src/amc/sca.cpp
jsturdy: This is just a port of the functionality in `cmsgemos`, which was done based on the V2 FW.
Absolutely, improving the interface will be done in the followup, and the discussion of behaviour can be made there. (One could there ask whether resetting the SCA module should be accompanied by *any* link resets, or whether the link resets should *always* be separate)

src/amc/ttc.cpp
jsturdy: Probably, as mentioned several times above, this PR is not intending to modify any existing functionality.

src/utils.cpp
jsturdy: Should all be gone now.

jsturdy:
bump @mexanick

mexanick (dismissed):

mexanick:
PR approved, please rebase on top of the current develop before merging

jsturdy:
> PR approved, please rebase on top of the current develop before merging

Will do, though will require another approval

PR #94, opened 2019-02-28
Clean up Makefile for improved build

jsturdy:
## Description

* Add generic rules
* Add better dependency tracking
* Fix bug building executables with `-shared`

### Types of changes

## Motivation and Context
This change simplifies the `Makefile`

## How Has This Been Tested?
Has been in all of my current development branches, I just cut it out and made it a separate change.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -21,9 +21,11 @@ include $(BUILD_HOME)/$(Package)/config/mfZynq.mk
 include $(BUILD_HOME)/$(Package)/config/mfCommonDefs.mk
 include $(BUILD_HOME)/$(Package)/config/mfRPMRules.mk
 
-IncludeDirs = ${BUILD_HOME}/$(Package)/include
+PackageBase = $(BUILD_HOME)/$(Package)
+ProjectBase = $(BUILD_HOME)/$(Project)
+IncludeDirs = $(PackageBase)/include
 IncludeDirs+= /opt/xhal/include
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -50,187 +56,109 @@ preprpm: default
...
-# TARGET_LIBS += lib/optical.so
-# TARGET_LIBS += lib/utils.so
-# TARGET_LIBS += lib/extras.so
-# TARGET_LIBS += lib/amc.so
-# TARGET_LIBS += lib/daq_monitor.so
-# TARGET_LIBS += lib/vfat3.so
-# TARGET_LIBS += lib/optohybrid.so
-# TARGET_LIBS += lib/calibration_routines.so
+## Variable that allows the generic rule to link properly against the dependencies listed above
+EXTRA_LINKS:=
```
bdorney: ~~is this supposed to be empty?~~

I see it's assigned below...although why does it appear first here as empty?

jsturdy: I think I had placed the empty assignment here because I thought the subsequent rule needed its definition, though I just tested with L74 commented out and things still are done correctly, so I've changed it

jsturdy:
Actually, while testing the change based on your comment, I realized that the object files were being deleted automatically (deemed by `make` as "intermediate")

bdorney (approved):

PR #95, opened 2019-03-01
Add Reed--Muller decoding function for VFAT3 ChipID

jsturdy:
## Description
Adds function `uint16_t decodeChipID(uint32_t encChipID)` to decode Reed--Muller encoded chip IDs

### Types of changes

## Motivation and Context
Necessary to decode the Reed--Muller encoded VFAT3 chip IDs

## How Has This Been Tested?
* ~Hasn't been tested with hardware, due to previous lack of compatible hybrids.~
  * Tested on 4 new hybrids, performs as expected
* Encoding and decoding has been tested with command line function.

### DIFF
--- a/src/vfat3.cpp
+++ b/src/vfat3.cpp
@@ -545,30 +545,18 @@ void statusVFAT3s(const RPCMsg *request, RPCMsg *response) {
 uint16_t decodeChipID(uint32_t encChipID)
 {
   // can the generator be static to limit creation/destruction of resources?
-  // convert to c++ new[]/delete[] ?
-  // convert to c++ smart ponters?
-  static int *encoded = 0;
-  static int *decoded = 0;
-  // static std::shared_ptr<int> encoded = 0;
-  // static std::shared_ptr<int> decoded = 0;
   static reedmuller rm = 0;
+  static std::unique_ptr<int[]> encoded = nullptr;
+  static std::unique_ptr<int[]> decoded = nullptr;
 
   if ((!(rm = reedmuller_init(2, 5)))
-      // || (!(encoded = (int*) calloc(rm->n, sizeof(int))))
-      // || (!(decoded = (int*) calloc(rm->k, sizeof(int))))) {
-      || (!(encoded = new int[rm->n]))
-      || (!(decoded = new int[rm->k]))
-      // || (!(encoded = std::make_shared<int[]>(rm->n)))
-      // || (!(decoded = std::make_shared<int[]>(rm->k)))
+      || (!(encoded = std::make_unique<int[]>(rm->n)))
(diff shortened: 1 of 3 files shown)


src/vfat3.cpp
```diff
@@ -539,6 +542,90 @@ void statusVFAT3s(const RPCMsg *request, RPCMsg *response) {
     rtxn.abort();
 }
 
+uint16_t decodeChipID(uint32_t encChipID)
```
mexanick, line 545: While I understand the intention of returning a smaller data type, the standard data word size in the `RPCSvc` is 32 bit. I would then operate `uint32_t` throughout.

src/vfat3.cpp
```diff
@@ -539,6 +542,90 @@ void statusVFAT3s(const RPCMsg *request, RPCMsg *response) {
...
+{
+  // can the generator be static to limit creation/destruction of resources?
+  // convert to c++ new[]/delete[] ?
+  // convert to c++ smart ponters?
+  static int *encoded = 0;
+  static int *decoded = 0;
+  // static std::shared_ptr<int> encoded = 0;
```
mexanick: Consider removing commented out lines as not needed.

mexanick (changes requested):
Few minor comments to address. Additionally, I think it might be useful to have the `warning` or `debug` log issued upon bit-flip encountered, if possible. (not sure whether it can/should be accomplished in this method...

src/vfat3.cpp
jsturdy: This function is local only, and it is expected that the thing actually using it will put the data in the appropriate return type for any RPC message.

mexanick: I see your point, but my personal preference would be to return `uint32_t` to avoid confusions and possible errors in the code which uses this function. However it is up to you whether to address this comment or not. I should note that in case of using it in multiple instance of other methods, the conversion will have to be performed in multiple places which is code dups and not the best coding practice...

src/vfat3.cpp
jsturdy: Fixed by enabling `std=c++14` in the Zynq mfDefs, and using `unique_ptr`

jsturdy:
> not sure whether it can/should be accomplished in this method...

This method will throw if 4 bit flips have occurred, fewer than that, and it simply corrects, more than that and it may throw or not, the algorithm is not guaranteed for more than 4.
It can not alert that fewer than 4 bit flips have occurred, and we probably don't care.

Again, it is expected that the function actually calling this decoder will handle the errors and emit them as desired.

jsturdy:
Rebased and ready for sign-off @bdorney @mexanick

mexanick (approved):
I approuve, but the users/developers should beware of 16bit return of `decodeChipID` method

PR #96, opened 2019-03-01
[chore] Fix warnings about signed vs. unsigned comparisons

jsturdy:
## Description
Clean up bad code leading to compiler warnings.

### Types of changes
- [x] Cleanup (non-breaking change which fixes an warning)

### DIFF
--- a/src/amc.cpp
+++ b/src/amc.cpp
@@ -188,7 +188,7 @@ std::vector<uint32_t> sbitReadOutLocal(localArgs *la, uint32_t ohN, uint32_t acq
         }
 
         acquisitionTime=difftime(time(NULL),startTime);
-        if(acquisitionTime > acquireTime){
+        if(uint32_t(acquisitionTime) > acquireTime){
             acquire=false;
         }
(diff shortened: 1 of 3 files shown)


bdorney (dismissed):

src/amc.cpp
```diff
@@ -188,7 +188,7 @@ std::vector<uint32_t> sbitReadOutLocal(localArgs *la, uint32_t ohN, uint32_t acq
         }
 
         acquisitionTime=difftime(time(NULL),startTime);
-        if(acquisitionTime > acquireTime){
+        if(uint32_t(acquisitionTime) > acquireTime){
```
mexanick, line 191: `cpplint` would shit bricks seeing the c-style type conversion )))

mexanick (approved):

src/amc.cpp
jsturdy: I agree this *should* be fixed elsewhere, but didn't want to spend the time tracking down the "right" way :-)

Issue #93, opened 2019-02-26
[feature request] Expose low level SCA functionality


jsturdy:
## Brief summary of issue
This feature will allow low-level functionality of the SCA to be exposed in the SW.

* readback/control of the ADC
* control over I2C of resets
* other possibilities
  * control of the SPI interface?
  * control of the GPIO interface?
  * control of the JTAG interface? (probably not)
  * control of the DAC interface?

### Types of issue
- [x] Feature request (request for change which adds functionality)

evka85:
I think these functions would be nice to have available via RPC:
  * readback/control of ADC
  * control of GPIO -- this is used to control the FPGA and VFATs resets

I would say that JTAG and SPI are not that critical, but may be nice to have. The SPI interface is only used to talk to the VTRX/VTTX, and actually we haven't used it before because normally it's not necessary to change the VTRX/VTTX parameters, but the idea is that with radiation the laser driver may degrade and in that case it may be necessary to increase the laser driver modulation current to improve the signal. I2C channels 1 and 5 are connected to the VTTXs.
JTAG is mostly a debugging tool I would say and I think it would be best if this is not used during operation, because we've seen some strange problems in CSC when accessing JTAG during operation. For debugging one can always just use the python script, so this is not critical. Possible uses are: reading sysmon, and loading firmware, although these functions can be done in other ways: sysmon can be read in OH firmware and then through the OH slow control, and the firmware can be loaded through the PROMless method.

Other SCA modules like SPI, DAC are not used on our OH (disconnected), and are therefore disabled in the SCA configuration. Actually I2C is also disabled by default, so if you want to play with it you'll have to enable it by writing to CRB, CRC, CRD control registers. This can be changed in the CTP7 firmware such that they are enabled by default

jsturdy:
> I think these functions would be nice to have available via RPC:
> * readback/control of ADC
> * control of GPIO -- this is used to control the FPGA and VFATs resets

OK, sounds good!

> The `SPI` interface is only used to talk to the `VTRX`/`VTTX`

I assume from context you meant `I2C` here?

> OH firmware and then through the OH slow control

Right, this is already done, so I don't see a reason to duplicate this functionality through the SCA, unless there is a performance benefit...

> Actually I2C is also disabled by default, so if you want to play with it you'll have to enable it by writing to CRB, CRC, CRD control registers. This can be changed in the CTP7 firmware such that they are enabled by default

[...]

PR #97, opened 2019-03-01
SCA command control functions

jsturdy:
## Description

Initial placeholders for SCA interface functionality, as discussed in #93 

* Add local-only functions to control the use of the various interfaces:
  * Generic `sendSCACommand` and then interfaces to the different blocks that call this command
    * `sendSCACommandWithReply` for transactions that expect returned data
  * SCA `CTRL` interface
    * Generic control registers B, C, and D
    * Chip ID and SEU counters
  * SCA `ADC` interface placeholder
  * SCA `GPIO` interface placeholder
  * SCA `I2C` interface placeholder
* Add enums
  * Per interface
    * for the different channels (e.g., ADC channels)
    * for the different commands
  * Different SCA communication channels
  * Different bits in the SCA command error word

### Types of changes

## Motivation and Context
Need to be able to reliably interface with the various interfaces on the SCA for GPIO/ADC/I2C

## How Has This Been Tested?
Has not been, for now just placeholders and framework for future development.

### DIFF
--- a/include/amc/sca.h
+++ b/include/amc/sca.h
@@ -8,12 +8,99 @@
 #define AMC_SCA_H
 
 #include "utils.h"
+#include "amc/sca_enums.h"
 
 /*!
  * \defgroup sca SCA module functionality
  */
 
+/*!
+ * \brief Prepare data for use with the SCA communication interfaces
+ *
+ * \details SCA TX/RX data is transmitted using the HDLC protocol, which is 16-bit length, and sent LSB to MSB.
+ * In the HDLC packet, it is sent/received as [<16:31><0:15>].
+ * The GEM_AMC firmware stores it as [<7:0><15:8><23:16><31:24>]
+ *
+ * \param data is the data to be converted to the appropriate ordering
+ */
+uint32_t formatSCAData(uint32_t const& data);
+
+/*!
+ * \brief Execute a command using the SCA interface
+ * \details Generic command to drive commands to all SCA modules
+ *  *
+ * \param la Local arguments structure
+ * \param ch channel to communicate with
+ * \param cmd command to send, referenced in sca_enums.h
+ * \param len length of the data to send, available 1,2,4
+ * \param data to send to the CTRL command
+ * \param linkMask bit list of OptoHybrids to send the commands to
+ */
+void sendSCACommand(localArgs* la, uint8_t const& ch, uint8_t const& cmd, uint8_t const& len, uint32_t data, uint16_t const& linkMask=0xfff);
+
+/*!
+ * \brief Execute a command using the SCA interface, and read the reply from the SCA
+ * \param la Local arguments structure
+ * \param ch channel to communicate with
+ * \param cmd command to send, referenced in sca_enums.h
+ * \param len length of the data to send, available 1,2,4
+ * \param data to send to the CTRL command
+ * \param linkMask bit list of OptoHybrids to send the commands to
+ */
+std::vector<uint32_t> sendSCACommandWithReply(localArgs* la, uint8_t const& ch, uint8_t const& cmd, uint8_t const& len, uint32_t data, uint16_t const& linkMask=0xfff);
+
+/*!
+ * \brief Execute a command using the SCA CTRL interface
+ * \details CTRL contains
+ *  * generic SCA R/W control registers
+ *  * Chip ID (on channel 0x14, ADC)
+ *  * SEU counter and reset (on channel 0x13, JTAG)
+ * \param la Local arguments structure
+ * \param cmd which command to send, referenced in sca_enums.h
+ * \param linkMask bit list of OptoHybrids to send the commands to
+ * \param len length of the data to send, default is 0x1
+ * \param data to send to the CTRL command, default is 0x0
+ */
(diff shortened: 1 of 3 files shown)


include/amc/sca.h
```diff
@@ -8,12 +8,102 @@
...
+ * \param cmd which command to send, referenced in sca_enums.h
+ * \param len length of the data to send, available 1,2,4
+ * \param data to send to the ADC command
+ * \param linkMask bit list of OptoHybrids to send the commands to
+ */
+std::vector<uint32_t> scaADCCommand(localArgs* la, SCAADCChannelT const& ch, uint8_t const& len, uint32_t data, uint16_t const& linkMask=0xfff);
+
 /*** CTRL submodule ***/
 /*!
  * \brief Reset the SCA module
```
mexanick, line 106: Please update doc with parameters

include/amc/sca_enums.h
```diff
@@ -0,0 +1,269 @@
...
+
+  /**
+   * \brief SCA Resets
+   * Types of resets able to be sent via the SCA
+   */
+  struct SCAResetType {
+    enum ESCAResetType {
+      SCAReset  = 0x0, ///< Reset the SCA
+      HardReset = 0x1, ///< Use the SCA to send a TTC-style HardReset command
+    } SCAResetTupe;
```
mexanick: `SCAResetTupe`-->`SCAResetType`

include/amc/sca_enums.h
```diff
@@ -0,0 +1,269 @@
...
+      CTRL_W_CRD   = 0x06, ///< Write to CTRL register D
+      CTRL_R_CRB   = 0x03, ///< Read from CTRL register B
+      CTRL_R_CRC   = 0x05, ///< Read from CTRL register C
+      CTRL_R_CRD   = 0x07, ///< Read from CTRL register D
+      CTRL_R_SEU   = 0xF1, ///< Read from the SEU monitor (channel 0x13)
+      CTRL_C_SEU   = 0xF0, ///< Reset the SEU monitor (channel 0x13)
+    } CTRLCommand;
+  };
+
+  struct I2CCommand {  //I2CCommand settings
```
mexanick, line 73: Missed doc

include/amc/sca_enums.h
```diff
@@ -0,0 +1,269 @@
...
+      GPIO_W_INTS      = 0x70, ///< Write GPIO interrupt vector register, length 4, read length 2
+      GPIO_R_INTS      = 0x71, ///< Read GPIO interrupt vector register, length 4, read length 2
+      GPIO_W_CLKSEL    = 0x80, ///< Write GPIO clock select register, length 4, read length 2
+      GPIO_R_CLKSEL    = 0x81, ///< Read GPIO clock select register, length 4, read length 2
+      GPIO_W_EDGESEL   = 0x90, ///< Write GPIO clock edge select register, length 4, read length 2
+      GPIO_R_EDGESEL   = 0x91, ///< Read GPIO clock edge select register, length 4, read length 2
+    } GPIOCommand;
+  };
+
+  struct ADCCommand {  //ADCCommand settings
```
mexanick, line 134: Missed doc

include/amc/sca_enums.h
```diff
@@ -0,0 +1,269 @@
...
+      // ADC commands V1??
+      ADC_V1_GO       = 0xB2, ///< Start of ADC conversion
+      ADC_V1_W_INSEL  = 0x30, ///< Write ADC register INSEL
+      ADC_V1_R_INSEL  = 0x31, ///< Read ADC register INSEL
+      ADC_V1_W_CURREN = 0x40, ///< Write ADC register
+      ADC_V1_R_CURREN = 0x41, ///< Read ADC register
+    } ADCCommand;
+  };  // struct ADCCommand
+
+      /**
```
mexanick: Indentation should be fixed

mexanick (changes requested):

jsturdy:
@mexanick, also rebased and I think all your comments addressed.
Would still like a cursory review from @evka85 to ensure that the framework makes sense from the FW perspective

evka85:
This looks very nice! Although it would be nice to introduce a few most common high level functions like:
  * read all ADC channels
  * send a manual hard reset to a single chamber

jsturdy:
> This looks very nice! Although it would be nice to introduce a few most common high level functions like:

Absolutely, this was just to have the necessary components before handing the expansion task off.
Just wanted to have the core set up in the "right" way, i.e., if certain actions need to happen for **every** type of transaction, putting those in the wrappers (e.g., "disable monitoring")

mexanick (dismissed):

evka85:
Oh.. yeah disable monitoring I think should just always be done, so maybe even put it in sendScaCommand function. But if this way of ADC readout tests ok (which it does on GE2/1), then I would like to take the firmware based ADC monitoring out completely, because its potential to interfere with other functions as well as non-portability between GE1/1 and GE2/1..

Actually after taking a second look, I think you overlooked setting the GEM_AMC.SLOW_CONTROL.SCA.MANUAL_CONTROL.LINK_ENABLE_MASK, which defines to which OHs the command will go. It's a bitmask, I think you should just write linkMask to it on sendSCACommand()

jsturdy:
@evka85, OK, I added the monitoring on/off at the level of the wrappers (one layer abstracted from the `sendSCACommand` to minimize the number of times it is toggled)

I also added the per-link setting of the module reset, but in the python tool this is protected with a "check if the register exists", but perhaps this is only relevant for old firmware?
If it needs to be protected, that can be added by the next person.

PR #98, opened 2019-03-05
Fix error in Makefile

jsturdy:
Dependent objects were not being built, e.g., `amc.o` was built, but not `amc/ttc.o`.
As such, they were not being triggered by the shared object rule that was designed to
add them to the `amc.so` shared object.

## Description
Modify the `.so` pattern rule to do a `.SECONDEXPANSION` and dynamically include all object files matching the stem, e.g., the rule will build this automatically based on
```
$(PackageLibDir)/%.so: $$(filter $$(PackageObjectDir)/$$*$$(pc).o, $(TargetObjects))
```
expanding into
```
lib/amc.so: amc.o amc/ttc.o amc/daq.o amc/sca.o
```

### Types of changes

## Motivation and Context
Compartmentalized objects to simplify code were not being included into the associated shared object.

## How Has This Been Tested?
Tested on rebased Reed--Muller branch; compiles as expected, symbols are properly linked

## Checklist:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -67,16 +67,15 @@ TargetObjects:= $(patsubst %.d,%.o,$(Dependencies))
 
 TargetLibraries:= memhub memory optical utils extras amc daq_monitor vfat3 optohybrid calibration_routines gbt
 
-## Variable that allows the generic rule to link properly against the dependencies listed above
-# EXTRA_LINKS:=
-
 # Everything links against these three
 BASE_LINKS = -lxhal -llmdb -lwisci2c
 
-## Generic shared object creation rule
-$(PackageLibraryDir)/%.so: $(PackageObjectDir)/%.o
+## Generic shared object creation rule, need to accomodate cases where we have lib.o lib/sub.o
+pc:=%
+.SECONDEXPANSION:
+$(PackageLibraryDir)/%.so: $$(filter $(PackageObjectDir)/$$*$$(pc).o, $(TargetObjects))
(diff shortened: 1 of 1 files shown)


jsturdy:
I need one more change here, as `optohybrid` is not listed as a dependency of `vfat3`, but `vfat3` needs `broadcastReadLocal` which is coming from `optohybrid`

Makefile
```diff
@@ -151,7 +150,7 @@ _all: build
 ### local (PC) test functions, need standard gcc toolchain, dirs, and flags
 .PHONY: test
 test: test/tester.cpp
-	g++ -O0 -g3 -fno-inline -o test/$@ $< $(INC) $(LDFLAGS) -lwiscrpcsvc
+	g++ -O0 -g3 -fno-inline -o test/$@ $< $(INC) $(LDFLAGS) -L/opt/wiscrpcsvc/lib -lwiscrpcsvc
```
mexanick, line 153: `-L...` is not needed, it is defined in the dependent libraries at the beginning of the file. Use `$(Libraries) -lwiscrpcsvc` instead

mexanick (changes requested):

Makefile
jsturdy: Not for the PC executables, which should be linked against *only* `x86_64` libraries, so we would maybe want to make sure to exclude the `arm` dirs from any link commands just in case the test executable links against a library with the same name

Issue #85, opened 2019-02-04
[feature-request] Define behaviour of error conditions, handling, and passing


jsturdy:
## Brief summary of issue
This issue will attempt to normalize and define how errors are handled inside the CTP7 modules and corresponding RPC error keys.

### Types of issue
- [X] Discussion
- [X] Feature request (request for change which adds functionality)

## Considerations
* If an RPC message `error` key is set, then the thing making the RPC call should expect that there was a failure that couldn't be handled inside the method itself.
* Most error handling should probably be done before the RPC callback, and either forwarded back to the RPC method (which would then set the `error` key), or handled based on the error in the function itself.
* Outside, in the calling code, handling of RPC `error` key messages should be done; in `cmsgemos` my current implementation is throwing an exception if the `error` key exists, and populating the exception with the message.
  * This has the benefit that it requires no special processing of the `error` message string (it simply assumes that things that could be handled were, and now it just needs to alert that things are not right)
  * It has the drawback that if the `error` message encodes something that could be parsed in the calling code and trigger some action, one would then have to parse the exception (not really a problem)
  * The question I would raise would be, why would we want this behaviour?  What types of routines would benefit from this type of fine-grained handling of errors that the generic exception wouldn't?

mexanick:
Considering that the trigger for this discussion is a very specific method and the error is set during the register access, I would not remove the setting of the error key in the register access methods. A better (and rather simple) solution to me would be to provide a method to remove/revoke key setting. This requires small changes to the RPCSvc code which has to be agreed (and propagated to) UW.

bdorney:
> Considering that the trigger for this discussion is a very specific method and the error is set during the register access, I would not remove the setting of the error key in the register access methods.

Well there was a discussion in the design of the GBT functions directly; and in some sense I think the design of those functions was limited because of how the `error` key is inserted and the lack of error handling in the repo.  I'm referring to [this discussion](https://github.com/cms-gem-daq-project/ctp7_modules/pull/75#discussion_r240734378).

It seems like having to check if an error key should be removed (particularly when this amount of checks could grow inside nested loops) is not ideal and could introduce unwanted processing time on the Zynq.

Rather than relying on UW to change the `RPCSvc` code (which they may reject) I think we could just add to the `readReg/readAddress` and `writeReg/writeAddress` methods a flag `errorsWillBeRaised` (or shorter name).  This flag would default to `True` and then the job of the developer is specifically to disable it when not desired.

[...]

PR #99, opened 2019-03-07
New Feature: Adding vfatMask argument to getVFAT3ChipIDsLocal and a check for sync

bdorney:
## Description

`getVFAT3ChipIDsLocal(...)` now takes a vfatMask argument which defaults to no vfat's masked.  It will also check for sync before querying chipID.  It will skip masked VFATs.

### Types of changes

## Motivation and Context

Need a way to skip VFATs that are incommunicado.

## How Has This Been Tested?

Tested on QC7 setup with a vfatmask of `0x8`

```python
In [1]: from gempython.tools.vfat_user_functions_xhal import HwVFAT

In [2]: vfatBoard = HwVFAT("eagle60",4)
Open pickled address table if available  /opt/cmsgemos/etc/maps/amc_address_table_top.pickle...
Initializing AMC eagle60
In [5]: chipIDs = vfatBoard.getAllChipIDs(mask=0x8,rawID=False)

In [6]: for vfatN,ID in enumerate(chipIDs):
    print(vfatN,hex(ID))
   ...: 
(0, '0xfbbL')
(1, '0xc440L')
(2, '0xe200L')
(3, '0x0L')
(4, '0x1705L')
(5, '0x3300L')
(6, '0x4610L')
(7, '0x3300L')
(8, '0x2001078L')
(9, '0x3742L')
(10, '0x800f68L')
(11, '0x2ec0L')
(12, '0xd100L')
(13, '0x2ec0L')
(14, '0x401775L')
(15, '0xf69L')
(16, '0xea80L')
(17, '0x3300L')
(18, '0x4841L')
(19, '0xd748L')
(20, '0xd100L')
(21, '0x3300L')
(22, '0xd100L')
(23, '0x40afeL')
```

## Checklist:

- [x] I have added tests to cover my changes.

### DIFF
--- a/src/vfat3.cpp
+++ b/src/vfat3.cpp
@@ -554,11 +554,29 @@ uint16_t decodeChipID(uint32_t encChipID)
   }
 }
 
-void getVFAT3ChipIDsLocal(localArgs * la, uint32_t ohN, bool rawID)
+void getVFAT3ChipIDsLocal(localArgs * la, uint32_t ohN, uint32_t vfatMask, bool rawID)
 {
+  //Determine the inverse of the vfatmask
+  uint32_t notmask = ~vfatMask & 0xFFFFFF;
+
+  //Check if VFATs are sync'd
+  uint32_t goodVFATs = vfatSyncCheckLocal(la, ohN);
+  if( (notmask & goodVFATs) != notmask)
+  {
+      char errBuf[200];
+      sprintf(errBuf,"One of the unmasked VFATs is not Synced. goodVFATs: %x\tnotmask: %x",goodVFATs,notmask);
+      la->response->set_string("error",errBuf);
+      return;
+  }
+
   std::string regName;
 
   for(int vfatN = 0; vfatN < 24; vfatN++) {
+    // Check if vfat is masked
+    if(!((notmask >> vfatN) & 0x1)){
+        continue;
+    } //End check if VFAT is masked
+
     char regBase [100];
     sprintf(regBase, "GEM_AMC.OH.OH%i.GEB.VFAT%i.HW_CHIP_ID",ohN, vfatN);
 
@@ -594,11 +612,12 @@ void getVFAT3ChipIDs(const RPCMsg *request, RPCMsg *response)
   // struct localArgs la = getLocalArgs(response);
   GETLOCALARGS(response);
 
-  uint32_t ohN = request->get_word("ohN");
-  bool rawID   = request->get_word("rawID");
+  uint32_t ohN      = request->get_word("ohN");
+  uint32_t vfatMask = request->get_word("vfatMask");
+  bool rawID        = request->get_word("rawID");
   LOGGER->log_message(LogManager::DEBUG, "Reading VFAT3 chipIDs");
 
-  getVFAT3ChipIDsLocal(&la, ohN, rawID);
+  getVFAT3ChipIDsLocal(&la, ohN, vfatMask, rawID);
 
   rtxn.abort();
 }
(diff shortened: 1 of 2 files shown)


src/vfat3.cpp
```diff
@@ -554,15 +554,34 @@ uint16_t decodeChipID(uint32_t encChipID)
...
     char regBase [100];
     sprintf(regBase, "GEM_AMC.OH.OH%i.GEB.VFAT%i.HW_CHIP_ID",ohN, vfatN);
 
     regName = std::string(regBase);
+    // Check if vfat is masked
+    if(!((notmask >> vfatN) & 0x1)){
+        la->response->set_word(regName,0xdeaddead);
+        continue;
+    } //End check if VFAT is masked
+
```
jsturdy, line 584: Is the worry here that if a VFAT isn't sync'd, `readReg` will fail, causing an error?
Why not add a `try/catch`, or is there a legitimate reason we keep adding `vfatMask` everywhere?

My main issue with this proliferation is that the tools should really not care about which VFATs are connected/working but should be able to determine this, and the user should be verifying that the results are as expected

bdorney: While a `try/catch` would be best presently there's no error handling beyond the rpc response word here; and we have not implemented exceptions (although issues #85 and #109 seem to start this discussion).

So [readReg](https://github.com/cms-gem-daq-project/ctp7_modules/blob/a7b0b98b4ffad8c76ff698a4f3b6bcf6b6e4cac0/src/utils.cpp#L283-L316) was changed to no longer set the error key in the RPC response.  But [writeReg](https://github.com/cms-gem-daq-project/ctp7_modules/blob/a7b0b98b4ffad8c76ff698a4f3b6bcf6b6e4cac0/src/utils.cpp#L385-L426) still does.  While this is a read instance the comment appears to be about `vfatMask` usage in general.

There's no way to remove an RPC key once it's set so the calling local function cannot remove the `error` word that `writeReg` sets.  And it's unclear to me if `readReg` will forever keep those lines where the error key is set commented out.

Additionally it fills up the log file with extraneous error messages that are not there, e.g. in the case a setup is not fully equipped with VFATs.

> My main issue with this proliferation is that the tools should really not care about which VFATs are connected/working but should be able to determine this, 

 Yes the tools could automatically determine this via [getOHVFATMaskLocal](https://github.com/cms-gem-daq-project/ctp7_modules/blob/a7b0b98b4ffad8c76ff698a4f3b6bcf6b6e4cac0/src/amc.cpp#L45-L57).  But it has been shown numerous times that the `SYNC_ERR_CNT` registers are not a failproof method to determine communication status.  Only attempting actual slow control transaction with the VFAT in question is fool proof.  But yes we could implement a call of `getOHVFATMaskLocal` here.

> and the user should be verifying that the results are as expected

It's been my experience that the users are not able to, even with significant training, ensure this 100% of the time.

jsturdy: OK, will have to return to this during exception handling then

> While this is a read instance the comment appears to be about vfatMask usage in general.

Quite

> But yes we could implement a call of getOHVFATMaskLocal here.

I'm not fully advocating for *this* per se, but that could be a solution (in a multiple operation, the first `try` could indicate whether a given VFAT is communicable or not and determine whether subsequent actions are made on the given VFAT).
I would strongly advocate that any call to a "allVFATs" function should, as its contract, expect in response 24x "number of things" returned, e.g., `readAllChipIDs` should expect 24 chipIDs (valid or invalid).

jsturdy (approved):

src/vfat3.cpp
bdorney: > I would strongly advocate that any call to a "allVFATs" function should, as its contract, expect in response 24x "number of things" returned, e.g., readAllChipIDs should expect 24 chipIDs (valid or invalid).

Okay we are on the same page, I'll try to keep this in mind when developing further.

Issue #101, opened 2019-03-08
[feature request] Make sbitRateScanParallelLocal() Multi Link


bdorney:
## Brief summary of issue

The function:

https://github.com/cms-gem-daq-project/ctp7_modules/blob/187a9452db5d164d16a102e9302a0dce2ad34414/src/calibration_routines.cpp#L626

right now just measures the rate for one OH.  However it could measure the rate for *all* OH's simultaneously since it's reading a set of counters that are specific to each OH.

### Types of issue

- [ ] Bug report (report an issue with the code)
- [ ] Feature request (request for change which adds functionality)

## Expected Behavior

The `ohN` and `vfatmask`  arguments could be dropped and replaced with an `ohMask` argument defaulting to `0x3ff`:

https://github.com/cms-gem-daq-project/ctp7_modules/blob/187a9452db5d164d16a102e9302a0dce2ad34414/src/calibration_routines.cpp#L626

### Automatically determining the VFAT Mask 
The following block could be dropped:

https://github.com/cms-gem-daq-project/ctp7_modules/blob/187a9452db5d164d16a102e9302a0dce2ad34414/src/calibration_routines.cpp#L632-L639

And the vfatMask per OH could be determined by using:

https://github.com/cms-gem-daq-project/ctp7_modules/blob/187a9452db5d164d16a102e9302a0dce2ad34414/src/amc.cpp#L44

An example implementation is:

https://github.com/cms-gem-daq-project/ctp7_modules/blob/187a9452db5d164d16a102e9302a0dce2ad34414/src/amc.cpp#L100-L111

### Storing the original channel map

Then you'd need to repeat the following block for each unmasked OH.

https://github.com/cms-gem-daq-project/ctp7_modules/blob/187a9452db5d164d16a102e9302a0dce2ad34414/src/calibration_routines.cpp#L641-L650

### Determining the Rate Monitor Addresses

For each unmasked OH you'd need to determine the rate monitor addresses:

https://github.com/cms-gem-daq-project/ctp7_modules/blob/187a9452db5d164d16a102e9302a0dce2ad34414/src/calibration_routines.cpp#L652-L659

And set the sbit counters to count for one second:

https://github.com/cms-gem-daq-project/ctp7_modules/blob/187a9452db5d164d16a102e9302a0dce2ad34414/src/calibration_routines.cpp#L665-L666

### Changing the DAC Loop

Then for each unmasked OH in the `ohMask` this loop would have to be executed:

[...]

PR #102, opened 2019-03-11
Feature sbit rate multilink

mexanick:
Implementing #101 
## Description

All the (not masked) optohybrids will be engaged in the calibration. In summary, up to 12 optohybrids are supported. The caller code should be modified to provide the `ohMask` instead of `ohN` and `maskOh`(`vfatMask`), the `vfatMask` will be determined automatically at runtime as suggested in the initial issue.
Returning data types will be similar to the single link version but expanded to hold the data from 12 OH. In case some of the OHs are masked, the data at the positions reserved to them will be random garbage. Indexing when data assigned:
```cpp
int idx = ohN*(dacMax-dacMin+1)/dacStep + (dacVal-dacMin)/dacStep;
outDataDacVal[idx] = dacVal;
outDataTrigRateOverall[idx] = readRawAddress(ohTrigRateAddr[ohN][24], la->response);
for(int vfat=0; vfat<24; ++vfat){
    if ( !( (notmask >> vfat) & 0x1)) continue;
    idx = ohN*24*(dacMax-dacMin+1)/dacStep + vfat*(dacMax-dacMin+1)/dacStep+(dacVal-dacMin)/dacStep;
    outDataTrigRatePerVFAT[idx] = readRawAddress(ohTrigRateAddr[ohN][vfat], la->response);
}
```
Indexing in the caller code:
```cpp
    if(isParallel){
        uint32_t outDataTrigRatePerVFAT[12*24*(dacMax-dacMin+1)/dacStep];
        uint32_t outDataDacValPerOH[12*(dacMax-dacMin+1)/dacStep];
        uint32_t outDataTrigRatePerOH[12*(dacMax-dacMin+1)/dacStep];
        sbitRateScanParallelLocal(&la, outDataDacValPerOH, outDataTrigRatePerVFAT, outDataTrigRatePerOH, ch, dacMin, dacMax, dacStep, scanReg, ohMask);

        response->set_word_array("outDataVFATRate", outDataTrigRatePerVFAT, 12*24*(dacMax-dacMin+1)/dacStep);
        response->set_word_array("outDataDacValue", outDataDacValPerOH, 12*(dacMax-dacMin+1)/dacStep);
        response->set_word_array("outDataCTP7Rate", outDataTrigRatePerOH, 12*(dacMax-dacMin+1)/dacStep);
    }
```
### Types of changes

## Motivation and Context

Required to faster and more efficient detector calibration and commissioning
## How Has This Been Tested?

Not yet tested. Requires matching changes in caller code and analyzing routines.
## Checklist:


src/calibration_routines.cpp
```diff
@@ -722,8 +747,7 @@ void sbitRateScan(const RPCMsg *request, RPCMsg *response)
     auto rtxn = lmdb::txn::begin(env, nullptr, MDB_RDONLY);
     auto dbi = lmdb::dbi::open(rtxn, nullptr);
 
-    uint32_t ohN = request->get_word("ohN");
-    uint32_t maskOh = request->get_word("maskOh");
```
bdorney, line 726: We can't actually nix these words and leave this function as is.

Down at line 774 is the "series" scan of `sbitRateScanLocal`.  I think now `sbitRateScanParallelLocal` and `sbitRateScanLocal` have diverged such that they are now incompatible.  Moreover why would you want to run something in series one VFAT at a time when you could run it in parallel.  I think we should drop `sbitRateScanLocal` from this function completely.  If there is a later need then a separate remote method could be created for `sbitRateScanLocal`.

bdorney:
@mexanick looks like this will need a rebase.

mexanick:
I'm not 100% sure whether I made the rebase in the cleanest possible way, but it is done...

bdorney:
I've validated through commit fb5cc7dc32f3c0f48e3ce08f45a502977715027f and this should be considered no longer work in progress but ready for review.

bdorney (approved):

jsturdy:
> I'm not 100% sure whether I made the rebase in the cleanest possible way, but it is done...

I suspect not, because as I try to look at the diff, I see many changes that were included in #105 
I suppose next SW dev meeting I'll have to outline the procedure I've been employing for getting up-to-date in these cases

jsturdy:
Questions for clarity:
* What do "parallel" and "series" mean in these modules?
  * multiple threads?
  * simultaneous setup and sequential readout of multiple <things>?
  * historical?
* When should a non-"multi-link" method ever be used (can not "single-link" be effected by using a single-bit mask in a "multi-link" function"?)
  * Simplifies the module code (maintainability)
  * Transfers the onus of correct setup onto the calling code

bdorney:
> Questions for clarity:
> 
> * What do "parallel" and "series" mean in these modules?
>   
>   * multiple threads?
>   * simultaneous setup and sequential readout of multiple ?
>   * historical?

This was historical.  Before @andrewpeck implemented per VFAT sbit counters in the OH FW this was being done only in the CTP7 FW which as a single trigger rate counter.  So the VFATs had to be placed into run mode one by one and done in *series*.  Whereas with the "new" (now very old) per VFAT counters this is done in parallel.

> * When should a non-"multi-link" method ever be used (can not "single-link" be effected by using a single-bit mask in a "multi-link" function"?)

I'm not sure I understand the question/word choice you're using.  Do you mean "can I use a multilink method with ohMask 0x1 just as easily as 0x3"? The answer is "yes you can."

The only time a "single" link function should be developed is if some resource cannot be used for multiple OH's simultaneously, the example of this is the `VFAT_DAQ_MONITOR` or the `SBIT_MONITOR` blocks in the FW which can only listen to one OH at a time.

Does this answer your questions?

jsturdy:
> > Questions for clarity:
> > 
> > * What do "parallel" and "series" mean in these modules?
> >   
> >   * multiple threads?
> >   * simultaneous setup and sequential readout of multiple ?
> >   * historical?
> 
> This was historical. Before @andrewpeck implemented per VFAT sbit counters in the OH FW this was being done only in the CTP7 FW which as a single trigger rate counter. So the VFATs had to be placed into run mode one by one and done in _series_. Whereas with the "new" (now very old) per VFAT counters this is done in parallel.

OK, so the series functions should be marked as obsolete and removed at some point?

> 
> > * When should a non-"multi-link" method ever be used (can not "single-link" be effected by using a single-bit mask in a "multi-link" function"?)
> 
> I'm not sure I understand the question/word choice you're using. Do you mean "can I use a multilink method with ohMask 0x1 just as easily as 0x3"? The answer is "yes you can."
> 
> The only time a "single" link function should be developed is if some resource cannot be used for multiple OH's simultaneously, the example of this is the `VFAT_DAQ_MONITOR` or the `SBIT_MONITOR` blocks in the FW which can only listen to one OH at a time.
> 
> Does this answer your questions?

I think almost, my final point would be that given that the partial answer is "yes you can", and that the more complete answer is "only when sharing a resource is not possible", I would still be in favour of eliminating "multi-link" and "single-link" nomenclature from the functions, as this can easily be incorporated into the design, as is done for the current "multi-link" things.
For truly shared "must be done one-at-a-time" things, one can simply move the loop to the top level of the function, e.g., 
```pseudocode
for oh in range(nOH):
  if ((ohMask>>oh) & 0x1):
    setup for link X
    run link X
    retrieve data from link X
    disable link X
```
In this way, the function describes what is being done, but doesn't burden anyone with details that are probably not salient (series, parallel, multi-link).  If the calling code wants to operate on all links, it is done (whether as a "multi-link"/parallel solution, or as a series solution), and if operation on only a single link is desired, the same function call is used, with a different parameter

bdorney:
> OK, so the series functions should be marked as obsolete and removed at some point?

Yes.

bdorney:
> I think almost, my final point would be that given that the partial answer is "yes you can", and that the more complete answer is "only when sharing a resource is not possible", I would still be in favour of eliminating "multi-link" and "single-link" nomenclature from the functions, as this can easily be incorporated into the design, as is done for the current "multi-link" things.

> In this way, the function describes what is being done, but doesn't burden anyone with details that are probably not salient (series, parallel, multi-link). If the calling code wants to operate on all links, it is done (whether as a "multi-link"/parallel solution, or as a series solution), and if operation on only a single link is desired, the same function call is used, with a different parameter

Sorry but I don't understand the problem/issue you're trying to solve.  Or what the concern is.  Can you give a more concrete or simpler explanation?  From the documentation it's clear what each function does.  And if it is not we should make a PR to improve this.

jsturdy:
Closes #101

jsturdy:
> > I think almost, my final point would be that given that the partial answer is "yes you can", and that the more complete answer is "only when sharing a resource is not possible", I would still be in favour of eliminating "multi-link" and "single-link" nomenclature from the functions, as this can easily be incorporated into the design, as is done for the current "multi-link" things.
> 
> > In this way, the function describes what is being done, but doesn't burden anyone with details that are probably not salient (series, parallel, multi-link). If the calling code wants to operate on all links, it is done (whether as a "multi-link"/parallel solution, or as a series solution), and if operation on only a single link is desired, the same function call is used, with a different parameter
> 
> Sorry but I don't understand the problem/issue you're trying to solve. Or what the concern is. Can you give a more concrete or simpler explanation? From the documentation it's clear what each function does. And if it is not we should make a PR to improve this.

The problem I'm specifically trying to solve is the proliferation of "ugly/duplicate" interfaces:
* `<functionality>MultiLink` (especially when separately there exists `<functionality>`), e.g., `getOHVFATMask` and `getOHVFATMaskMultiLink`, both of these could be replaced by a completely generic `getOHVFATMask` that would handle both single and multi link operation
* `<functionality>Parallel` (I understand that this is not really a problem and will be removed)
* `<functionality>Series` (I understand there aren't currently any of these)

PR #104, opened 2019-03-13
Hotfix for V2B scan issues

jsturdy:
## Description

* Update the V2B `ULTRA` scan modules to use the `CONF` node for the configuration registers
* Fix bug in `stopCalPulse2AllChannelsLocal`
  * Writing was allowed to VFAT channel 128
  * Error message was incorrectly using user specified `ch_max`
* Fix bug in the linking of the `extras` library, to ensure it links against `memhub`

### Types of changes

## Motivation and Context
V2B scans were failing due to a change in the address table structure that was not propagated.

## How Has This Been Tested?
@ElizabethRoseStarling is testing while taking scurves on the P5 chambers.

### DIFF
--- a/src/optohybrid.cpp
+++ b/src/optohybrid.cpp
@@ -277,18 +277,18 @@ void configureScanModuleLocal(localArgs * la, uint32_t ohN, uint32_t vfatN, uint
     writeRawReg(la, scanBase + ".RESET", 0x1);
 
     // write scan parameters
-    writeReg(la, scanBase + ".MODE", scanmode);
+    writeReg(la, scanBase + ".CONF.MODE", scanmode);
     if (useUltra){
-        writeReg(la, scanBase + ".MASK", mask);
+        writeReg(la, scanBase + ".CONF.MASK", mask);
     }
     else{
-        writeReg(la, scanBase + ".CHIP", vfatN);
+        writeReg(la, scanBase + ".CONF.CHIP", vfatN);
     }
-    writeReg(la, scanBase + ".CHAN", ch);
-    writeReg(la, scanBase + ".NTRIGS", nevts);
-    writeReg(la, scanBase + ".MIN", dacMin);
-    writeReg(la, scanBase + ".MAX", dacMax);
-    writeReg(la, scanBase + ".STEP", dacStep);
+    writeReg(la, scanBase + ".CONF.CHAN", ch);
+    writeReg(la, scanBase + ".CONF.NTRIGS", nevts);
+    writeReg(la, scanBase + ".CONF.MIN", dacMin);
+    writeReg(la, scanBase + ".CONF.MAX", dacMax);
+    writeReg(la, scanBase + ".CONF.STEP", dacStep);
 
     return;
 } //End configureScanModuleLocal(...)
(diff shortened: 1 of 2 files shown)


src/optohybrid.cpp
```diff
@@ -277,18 +277,18 @@ void configureScanModuleLocal(localArgs * la, uint32_t ohN, uint32_t vfatN, uint
...
-    writeReg(la, scanBase + ".CHAN", ch);
-    writeReg(la, scanBase + ".NTRIGS", nevts);
-    writeReg(la, scanBase + ".MIN", dacMin);
-    writeReg(la, scanBase + ".MAX", dacMax);
-    writeReg(la, scanBase + ".STEP", dacStep);
+    writeReg(la, scanBase + ".CONF.CHAN", ch);
+    writeReg(la, scanBase + ".CONF.NTRIGS", nevts);
+    writeReg(la, scanBase + ".CONF.MIN", dacMin);
+    writeReg(la, scanBase + ".CONF.MAX", dacMax);
+    writeReg(la, scanBase + ".CONF.STEP", dacStep);
```
bdorney, line 291: why not put `CONF` into the `scanbase` string? Or is there an instance where this would cause issues?

src/optohybrid.cpp
```diff
@@ -356,19 +356,19 @@ void printScanConfigurationLocal(localArgs * la, uint32_t ohN, bool useUltra){
...
+    map_regValues[scanBase + ".CONF.NTRIGS"] = 0;
     map_regValues[scanBase + ".MONITOR.STATUS"] = 0;
 
     if (useUltra){
-        map_regValues[scanBase + ".MASK"] = 0;
+        map_regValues[scanBase + ".CONF.MASK"] = 0;
     }
     else{
-        map_regValues[scanBase + ".CHIP"] = 0;
+        map_regValues[scanBase + ".CONF.CHIP"] = 0;
```
bdorney, line 371: same as above.

src/optohybrid.cpp
```diff
@@ -472,11 +472,11 @@ void getUltraScanResultsLocal(localArgs * la, uint32_t *outData, uint32_t ohN, u
     //Get L1A Count & num events
     uint32_t ohnL1A_0 =  readReg(la, "GEM_AMC.OH.OH" + strOhN + ".COUNTERS.T1.SENT.L1A");
     uint32_t ohnL1A   =  readReg(la, "GEM_AMC.OH.OH" + strOhN + ".COUNTERS.T1.SENT.L1A");
-    uint32_t numtrigs = readReg(la, scanBase + ".NTRIGS");
+    uint32_t numtrigs = readReg(la, scanBase + ".CONF.NTRIGS");
 
     //Print latency counts
     bool bIsLatency = false;
-    if( readReg(la, scanBase + ".MODE") == 2){
+    if( readReg(la, scanBase + ".CONF.MODE") == 2){
```
bdorney, line 479: same as above.

bdorney (commented):
clarification on `scanbase` needed, otherwise looks good.

src/optohybrid.cpp
jsturdy: `CONF` is the sub node for the configuration part, everything else is as it was previously (yes, it would cause issues)

bdorney (approved):

PR #107, opened 2019-03-26
[feature] BLASTER RAM interface

jsturdy:
## Description
This PR brings in interfaces to reading/writing configuration data into the BLASTER RAM on the CTP7

### Types of changes

## Motivation and Context
* Necessary for the BLASTER streaming interface to put a configuration into the hardware.
* Useful before for storing configurations in RAM rather than in text files.

## How Has This Been Tested?
Reading and writing to the BLASTER RAM has been verified

## Checklist:


include/hw_constants.h
```diff
@@ -13,68 +13,82 @@
...
 
     /*! \brief Mappings between elinks, GBT index and VFAT index.
      */
     namespace elinkMappings {
-        namespace ge11{
+        namespace ge11 {
             /*! \brief Mapping from VFAT index to GBT index.
              */
-            constexpr std::array<uint32_t, 24> VFAT_TO_GBT{
+            constexpr std::array<uint32_t, 24> VFAT_TO_GBT {
```
mexanick, line 79: out of curiosity, why the `uint32_t` is used here instead of `uint8_t`?  Also, considering the l91, what's the max value there? if it is less than 255, a `uint8_t` would be also sufficient

include/utils.h
```diff
@@ -158,12 +159,64 @@ uint32_t applyMask(uint32_t data, uint32_t mask);
  */
 uint32_t readReg(LocalArgs * la, const std::string & regName);
 
-/*! \fn void writeReg(LocalArgs * la, const std::string & regName, uint32_t value)
+/*!
+ *  \brief Reads a block of values to a contiguous address space.
```
mexanick: shouldn't be "from a contiguous"?..

include/utils.h
```diff
@@ -158,12 +159,64 @@ uint32_t applyMask(uint32_t data, uint32_t mask);
...
+ *  \param size number of words to read (should this just come from the register properties?
+ *  \param result Pointer to an array to hold the result
+ *  \param offset Start reading from an offset from the base address returned by regName
+ *  \returns the number of uint32_t words in the result (or better to return a std::vector?
+ */
+uint32_t readBlock(localArgs* la, const std::string& regName, uint32_t* result, const uint32_t& size, const uint32_t& offset=0);
+
+/*!
+ *  \brief Reads a block of values to a contiguous address space.
+ *  \param la Local arguments structure
```
mexanick: Is this an overload to use locally without the local arguments? If so, please state it in the method description and remove the above line

include/amc/blaster_ram_defs.h
```diff
@@ -0,0 +1,29 @@
...
+};
+
+// <name>  is the enum scoped namespace for scope::VALUE access
+// <name>T is the enum type
+
+// typedef the struct for access to the members via struct::VALUE
+typedef BLASTERSettings::BLASTERType  BLASTERType;
+
+// typedef the enum for casting and access
+typedef BLASTERSettings::BLASTERType::EBLASTERType  BLASTERTypeT;
```
mexanick, line 27: Offtop: do we have a convention on `typedef` names? I mean `MyTypeT` vs `MyType_t` vs ...
Reason: I used already `MyType_t` somewhere else... Don't have a particular preference, but I think we should synchronize.

src/extras.cpp
```diff
@@ -43,22 +66,104 @@ void mlistread(const RPCMsg *request, RPCMsg *response) {
...
       return;
     }
   }
-	response->set_word_array("data", data, count);
+  // return type?
+  response->set_word_array("data", data, count);
 }
+
+
+//////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
```
mexanick: ??

src/amc/blaster_ram.cpp
```diff
@@ -0,0 +1,501 @@
...
+    return readReg(la, "GEM_AMC.CONFIG_BLASTER.STATUS.OH_RAM_SIZE");
+  case (BLASTERType::VFAT) :
+    return readReg(la, "GEM_AMC.CONFIG_BLASTER.STATUS.VFAT_RAM_SIZE");
+  case (BLASTERType::ALL) :
+    ram_size  = getRAMMaxSize(la, BLASTERType::GBT);
+    ram_size += getRAMMaxSize(la, BLASTERType::OptoHybrid);
+    ram_size += getRAMMaxSize(la, BLASTERType::VFAT);
+    return ram_size;
+  default:
+    break;
```
mexanick, line 32: I think we should use `EMIT_RPC_ERROR` here

src/amc/blaster_ram.cpp
```diff
@@ -0,0 +1,501 @@
...
+    return base;
+  case (BLASTERType::OptoHybrid) :
+    regName = stdsprintf("GEM_AMC.CONFIG_BLASTER.RAM.OH_FPGA_OH%d",int(ohN));
+    base = getAddress(la, regName);
+    return base;
+  case (BLASTERType::VFAT) :
+    if (partN > (oh::VFATS_PER_OH-1)) {
+      std::stringstream errmsg;
+      errmsg << "Invalid GBT specified: VFAT" << partN << " > " << (oh::VFATS_PER_OH-1);
+      LOGGER->log_message(LogManager::ERROR, errmsg.str());
```
mexanick, line 75: Here and below, I would use `EMIT_RPC_ERROR` and return an error report to the host PC

src/amc/blaster_ram.cpp
```diff
@@ -0,0 +1,501 @@
...
+        std::stringstream reg;
+        reg << "GEM_AMC.CONFIG_BLASTER.RAM.VFAT_OH" << oh;
+        nwords += readBlock(la, reg.str(), blob, perblk);
+        blob += perblk;
+      }
+    }
+    return nwords;
+  }
+}
+
```
mexanick, line 253: extra line

src/amc/blaster_ram.cpp
```diff
@@ -0,0 +1,501 @@
...
+  BLASTERTypeT type = static_cast<BLASTERTypeT>(request->get_word("type"));
+  LOGGER->log_message(LogManager::DEBUG, stdsprintf("BLASTERTypeT is 0x%x", type));
+
+  uint32_t blob_sz = request->get_binarydata_size("confblob");
+  uint32_t confblob[blob_sz];
+  LOGGER->log_message(LogManager::DEBUG, stdsprintf("blob_sz is 0x%x", blob_sz));
+  request->get_binarydata("confblob", confblob, blob_sz);
+  try {
+    writeConfRAMLocal(&la, type, confblob, blob_sz);
+  } catch (std::runtime_error& e) {
```
mexanick, line 442: Again, this error probably can be worked out inside the call function

src/utils.cpp
```diff
@@ -4,17 +4,17 @@ memsvc_handle_t memsvc;
 
 struct localArgs getLocalArgs(RPCMsg *response)
 {
```
mexanick, line 6: Offtop: do we follow 1tab=shift width=2 spaces? Or 4?

include/utils.h
jsturdy: yeah :-), i saw another one in the code recently and debated changing it and pushing.
I have a second PR which connects these functions to the individual devices, so can fix there in the interest of streamlining things

include/utils.h
jsturdy: I added it as such, but have left it empty right now, since it can't do any validation (it might be the "block" transaction you were looking for regarding reading/writing configs via slow control.
I can remove this doc string in this PR (or the aforementioned future one) but will leave the function body empty for now

include/amc/blaster_ram_defs.h
jsturdy: Not sure, I copied this style from `cmsgemos` and only use this convoluted mechanism because when implemented in `cmsgemos` couldn't use `enum class`
Should probably update to `enum class` everywhere (just need to ensure it keeps the same functionality)...

src/extras.cpp
jsturdy: Excellent question! I blame my djanky keyboard... I'll fix this and the other whitespace/doc changes

src/utils.cpp
jsturdy: Unclear, I've been using 2 spaces in `C/C++` and 4 in `python` (which is what my editor enforces)
I've seen a mixed style here, and tried to make minimal disruptions

src/amc/blaster_ram.cpp
jsturdy: The one design choice I made in this interface is to have local methods `throw` and RPC callbacks handle (adding the error string to the response.
We had had some discussion in the past about "wouldn't it be good to handle exceptions better" so I took this opportunity to start.

src/amc/blaster_ram.cpp
jsturdy: I prefer this way because if you're chaining local methods in your RPC callback, you probably want to make sure you don't continue after the first error condition is met.
Or, if you know how to hand that error condition, you *can*

src/amc/blaster_ram.cpp
jsturdy: The error is (and should be) returned to the host PC in the RPC callback

include/hw_constants.h
jsturdy: Indeed, this was set up by @lpetre-ulb, and if we need some optimizations, can incorporate them at a future date.

src/amc/blaster_ram.cpp
```diff
@@ -0,0 +1,501 @@
...
+    // read blob to specific GBT RAM, as specified by ohMask, support non consecutive OptoHybrids?
+    const uint32_t perblk = gbt::GBT_SINGLE_RAM_SIZE*gbt::GBTS_PER_OH;
+    uint32_t nwords = 0x0;
+    uint32_t* blob = static_cast<uint32_t*>(gbtblob);
+    for (size_t oh = 0; oh < amc::OH_PER_AMC; ++oh) {
+      if ((0x1<<oh)&ohMask) {
+        std::stringstream reg;
+        reg << "GEM_AMC.CONFIG_BLASTER.RAM.GBT_OH" << oh;
+        nwords += readBlock(la, reg.str(), blob, perblk);
+        blob += perblk;
```
evka85, line 185: I'm not sure if it's a good idea to let the user skip some OHs and end up with a smaller blob (only selected OHs concatenated). Perhaps it would be better to fill the space between unselected OHs with zeros or something like that. And in general not sure if such a usecase is really necessary, maybe just allow reading a config of just one OH...

jsturdy: I was initially thinking this, but then wondered about the case where this (the corresponding `write` function) would (probably due to user error) overwrite some part of the RAM that was previously written with a valid configuration.
I had mentioned in our SW dev meeting that I didn't really like the "masked" operation mode, and wasn't sure how willing I would be to support it long term

evka85:
Apart from my question on the OH mask, this looks very nice!

src/amc/blaster_ram.cpp
evka85: yeah, if there isn't a compelling use case for it, I'd remove it to avoid future problems. I can see that writing a single OH config may be useful, but a combo of several OHs (but not all) I'm not so sure..

include/hw_constants.h
bdorney: I think the idea was that since the `wisc_rpc` API only supports 32 bit transactions on get/set word/word array @lpetre-ulb did it this way.  But if this is set by a blob read then yes the data type could be changed.

evka85: Don't know if it's relevant here, but just wanted to note that the CTP7 firmware also operates only in 32bit words, so BLASTER RAMs can only be written / read 32bits at a time.
So to me it seems that having 32bit datatype in software makes sense, but I may be missing something.

jsturdy: These mappings aren't written to the a register, so using a semantically correct datatype is probably the correct thing to do (see the next mapping, which is using `uint8_t`)
Things that are doing register writes are (or should) ensure that the type they are writing is compatible with a 32-bit word.

mexanick: can you add a `TODO` or `FIXME` note there?

include/utils.h
mexanick: yeah, just remove the misplaced doc string atm

src/amc/blaster_ram.cpp
mexanick: ok, the I think we can benefit of throwing macros, as this construction is used in many places and we can reduce a codebase a bit

src/utils.cpp
mexanick: need a manifesto and refactoring ;)

include/hw_constants.h
jsturdy: Done with #108

src/amc/blaster_ram.cpp
jsturdy: Yes, I agree, we should have some exception handing commonalities implemented (see #109)

mexanick (approved):

Issue #103, opened 2019-03-12
Add a wait time between link reset and VFAT link checks


bdorney:
## Brief summary of issue

Based on a discussing that @evka85 and I had we should increase the time between when a link reset is performed and when VFAT sync error counters are checked. These counters are checked once per orbit so we should allow one or more orbits to occur by before checking those counters after a link reset.

An orbit is 9.09091E-05 seconds.

### Types of issue

- [ ] Bug report (report an issue with the code)
- [x] Feature request (request for change which adds functionality)

## Expected Behavior

After line 801:

https://github.com/cms-gem-daq-project/ctp7_modules/blob/187a9452db5d164d16a102e9302a0dce2ad34414/src/daq_monitor.cpp#L798-L802

we should insert a sleep statement for at least N orbits (N=1?).

We could consider also inserting a sleep statement in between each OH (e.g. between [lines 806 & 807](https://github.com/cms-gem-daq-project/ctp7_modules/blob/187a9452db5d164d16a102e9302a0dce2ad34414/src/daq_monitor.cpp#L806-L807) but I'm not sure that's necessary (as 24 register reads should take at least some 200 micro-seconds?).

## Current Behavior

Right after a link reset is sent we immediately start checking the `SYNC_ERR_CNT` registers.

## Context (for feature requests)

Returned results by monitoring [getmonVFATLinkLocal](https://github.com/cms-gem-daq-project/ctp7_modules/blob/187a9452db5d164d16a102e9302a0dce2ad34414/src/daq_monitor.cpp#L796) may not be accurate because we aren't waiting enough time for the sync error counters to increment.

## Your Environment

* Version used: 187a9452db5d164d16a102e9302a0dce2ad34414

jsturdy:
Partially addressed by #110

PR #112, opened 2019-04-01
Bug Fix: Bad Node Names & Hard Coded LMDB Size Limit

bdorney:
## Description

Reverted #110 as it adds an unnecessary sleep to `getmonGBTLinkLocal` and also the node names used there are *invalid*.

Keeps the sleep statement that partially addresses #103 in `getmonVFATLinkLocal`.

### Types of changes

## Motivation and Context

Invalid node names introduced.

## Checklist:

### DIFF
--- a/src/daq_monitor.cpp
+++ b/src/daq_monitor.cpp
@@ -4,15 +4,11 @@
  *  \author Brian Dorney <brian.l.dorney@cern.ch>
  */
 
-#include "daq_monitor.h"
-
-#include <chrono>
-#include <thread>
-#include <string>
-
-#include "utils.h"
 #include "amc.h"
+#include "daq_monitor.h"
 #include "hw_constants.h"
+#include <string>
+#include "utils.h"
 
 void getmonTTCmainLocal(localArgs * la)
 {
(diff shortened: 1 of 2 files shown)


mexanick (approved):

src/daq_monitor.cpp
```diff
@@ -240,30 +238,29 @@ void getmonGBTLinkLocal(localArgs * la, int NOH, bool doReset)
...
     if (doReset) {
          writeReg(la, "GEM_AMC.GEM_SYSTEM.CTRL.LINK_RESET", 0x1);
-         std::this_thread::sleep_for(std::chrono::microseconds(92)); // FIXME sleep for N orbits
     }
 
     std::string regName, respName; //regName used for read/write, respName sets word in RPC response
     for (int ohN=0; ohN < NOH; ++ohN) {
         for (unsigned int gbtN=0; gbtN < gbt::GBTS_PER_OH; ++gbtN) {
             //Ready
             respName = stdsprintf("OH%i.GBT%i.READY",ohN,gbtN);
```
jsturdy, line 247: OK, any reason why the response name differs from the register name?
e.g., 
`OH0.GBT0.READY` vs `OH0.GBT0_READY` 

Fixing this expectation (whether here in the code, or in the address table itself) would simplify the programmatic handling of things...

jsturdy (changes requested):
Changes requested only to prevent merging before addressing comment

src/daq_monitor.cpp
bdorney: It was done for aesthetic reasons when reading back.

jsturdy: OK... I'd strongly suggest this gets cleaned up at some point in the near future (one way or the other)

jsturdy (approved):

PR #115, opened 2019-04-05
New Feature: Updates to DAC & GBT Phase Scans

bdorney:
## Description

Before a DAC scan is launched L1A's will now be disabled and VFATs will be taken out of run mode before configuring DAC Monitoring on all VFATs.

The GBT phase scan will no longer check the `LINK_GOOD` register as it was largely useless for checking communication status.  Now each GBT phase scan point will attempt several slow control transactions per point before declaring the phase as good.

### Types of changes

## Motivation and Context

During QC7 tests saw that it was possible to generate both bus errors and out of sync states for VFATs in cases where the VFAT was placed into run mode and then subsequent slow control actions were performed.  This caused crashes or unintentional masking of VFATs. 

Also improves reliability of phase scan results.

## How Has This Been Tested?

Extensively on QC7 setup.

## Checklist:

- [x] I have added tests to cover my changes.

### DIFF
--- a/src/calibration_routines.cpp
+++ b/src/calibration_routines.cpp
@@ -1178,6 +1178,10 @@ std::vector<uint32_t> dacScanLocal(localArgs *la, uint32_t ohN, uint32_t dacSele
     int nDacValues = (dacMax-dacMin+1)/dacStep;
     std::vector<uint32_t> vec_dacScanData(24*nDacValues); //Each element has bits [0:7] as the current dacValue, and bits [8:17] as the ADC read back value
 
+    //Put VFATs into slow control only mode
+    writeReg(la, "GEM_AMC.GEM_SYSTEM.VFAT3.SC_ONLY_MODE", 0x1);
+    broadcastWriteLocal(la, ohN, "CFG_RUN", 0x0, mask);
+
     //Configure the DAC Monitoring on all the VFATs
     configureVFAT3DacMonitorLocal(la, ohN, mask, dacSelect);
 
--- a/src/calibration_routines.cpp
+++ b/src/calibration_routines.cpp
@@ -1179,16 +1179,14 @@ std::vector<uint32_t> dacScanLocal(localArgs *la, uint32_t ohN, uint32_t dacSele
     std::vector<uint32_t> vec_dacScanData(24*nDacValues); //Each element has bits [0:7] as the current dacValue, and bits [8:17] as the ADC read back value
 
     //Put VFATs into slow control only mode
-    writeReg(la, "GEM_AMC.GEM_SYSTEM.VFAT3.SC_ONLY_MODE", 0x1);
+    writeReg(la, "GEM_AMC.TTC.CTRL.L1A_ENABLE", 0x0);
     broadcastWriteLocal(la, ohN, "CFG_RUN", 0x0, mask);
 
     //Configure the DAC Monitoring on all the VFATs
     configureVFAT3DacMonitorLocal(la, ohN, mask, dacSelect);
 
-    //Take the VFATs out of slow control only mode
-    writeReg(la, "GEM_AMC.GEM_SYSTEM.VFAT3.SC_ONLY_MODE", 0x0);
-
     //Set the VFATs into Run Mode
+    writeReg(la, "GEM_AMC.GEM_SYSTEM.VFAT3.SC_ONLY_MODE", 0x0);
     broadcastWriteLocal(la, ohN, "CFG_RUN", 0x1, mask);
     LOGGER->log_message(LogManager::INFO, stdsprintf("VFATs not in 0x%x were set to run mode", mask));
     std::this_thread::sleep_for(std::chrono::seconds(1)); //I noticed that DAC values behave weirdly immediately after VFAT is placed in run mode (probably voltage/current takes a moment to stabalize)
--- a/src/calibration_routines.cpp
+++ b/src/calibration_routines.cpp
@@ -1178,7 +1178,7 @@ std::vector<uint32_t> dacScanLocal(localArgs *la, uint32_t ohN, uint32_t dacSele
     int nDacValues = (dacMax-dacMin+1)/dacStep;
     std::vector<uint32_t> vec_dacScanData(24*nDacValues); //Each element has bits [0:7] as the current dacValue, and bits [8:17] as the ADC read back value
 
-    //Put VFATs into slow control only mode
+    //Block L1A's then take VFATs out of run mode
     writeReg(la, "GEM_AMC.TTC.CTRL.L1A_ENABLE", 0x0);
     broadcastWriteLocal(la, ohN, "CFG_RUN", 0x0, mask);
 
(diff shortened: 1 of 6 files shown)


jsturdy:
As per discussion in #116, this may not the best way to solve this.

To better understand use cases, will there ever be a case where a DAC scan is being run on VFAT(s) on one OptoHybrid connected to a given AMC, while other OptoHybrids are actively participating in "data taking"?
If so, we could consider asking @evka85 to have the `INPUT_ENABLE_MASK` also serve as a per link trigger blocker.  This would of course need to be weighed against the potential use case where a given OptoHybrid is *not* participating in data taking, but one still wants to have the front end receive L1As...

As I currently understand it, this routine itself should never be running when one is expecting the VFAT data formatter to be active, right?
So, I would instead ensure that whatever is driving should ensure that things are properly configured for the desired operational mode (i.e., much further out ensure that L1As are not being sent), rather than relying on this mechanism to potentially mask another issue.

Actually this raises another question, can you verify whether it was an active data formatter block that was the likely culprit (check if the EC counter is rolling up)?
Because if this was **not** the case, then although this solves the current problem, the underlying issue is still there and needs to be understood.

One can consider if this routine itself should **only** be run when the VFAT is in `SC Only` mode (as per #116 discussion), and if so, then this solution is likely to be appropriate (or at least, part of the appropriate solution).

bdorney:
Yeah further study and the best way forward is needed.  I'll come back to this on Monday.

bdorney:
@jsturdy @mexanick ready for review

src/calibration_routines.cpp
```diff
@@ -1178,13 +1178,15 @@ std::vector<uint32_t> dacScanLocal(localArgs *la, uint32_t ohN, uint32_t dacSele
...
+    broadcastWriteLocal(la, ohN, "CFG_RUN", 0x0, mask);
+
     //Configure the DAC Monitoring on all the VFATs
     configureVFAT3DacMonitorLocal(la, ohN, mask, dacSelect);
 
-    //Take the VFATs out of slow control only mode
-    writeReg(la, "GEM_AMC.GEM_SYSTEM.VFAT3.SC_ONLY_MODE", 0x0);
-
     //Set the VFATs into Run Mode
+    writeReg(la, "GEM_AMC.GEM_SYSTEM.VFAT3.SC_ONLY_MODE", 0x0);
```
jsturdy, line 1189: Remove? As it is writing a `0`, which should be the value beforehand?

src/calibration_routines.cpp
```diff
@@ -1178,13 +1178,15 @@ std::vector<uint32_t> dacScanLocal(localArgs *la, uint32_t ohN, uint32_t dacSele
     int nDacValues = (dacMax-dacMin+1)/dacStep;
     std::vector<uint32_t> vec_dacScanData(24*nDacValues); //Each element has bits [0:7] as the current dacValue, and bits [8:17] as the ADC read back value
 
+    //Put VFATs into slow control only mode
```
jsturdy: Update comment (or remove)

src/calibration_routines.cpp
bdorney: This is here at recommendation from @evka85.  The suggestion here, and in all other data taking routines/procedures, is to place the VFATs in a known configuration.  Granted I'm not sure how this affects the state machine of the ADC on the VFAT.  

In any case I would prefer to keep it (here and elsewhere); as in the future it might be possible that something does place VFATs into SC only mode and then we would need to remove them from that mode.

src/calibration_routines.cpp
bdorney: Done

src/utils.cpp
```diff
@@ -300,7 +300,9 @@ uint32_t readReg(localArgs * la, const std::string & regName)
     uint32_t rmask = stoull(tmp[2], nullptr, 16);
     if (memhub_read(memsvc, raddr, 1, data) != 0) {
       // response->set_string("error", std::string("memsvc error: ")+memsvc_get_last_error(memsvc));
-      LOGGER->log_message(LogManager::ERROR, stdsprintf("read memsvc error: %s", memsvc_get_last_error(memsvc)));
+      if (printReadErrors) {
+        LOGGER->log_message(LogManager::ERROR, stdsprintf("read memsvc error: %s", memsvc_get_last_error(memsvc)));
+      }
```
jsturdy: Was this shown to have any impact on the performance in the end?
If not, I would advocate for a cleaner interface that doesn't pollute the functions.

mexanick: I support @jsturdy 's comment. While excessive logging for sure will impact the performance, in this particular case I see no reason to provide an opportunity to hide the error message in the log. In other cases if needed this can (and should) be managed by log levels

bdorney: The goal here is *not* to impact performance (no test that I did showed that logging slows down or speeds up any action on the card) but to make the GBT phase scan or the repeated reg read log readable.

Otherwise the log is incomprehensible because it's full of 1000's of `read memsvc error:`'s because you're potentially having bad phases during a GBT scan or your testing if the communication is good (repeated reg read) and this makes it hard to understand *anything* else.

bdorney: So yes there *is* a reason to hide this; two reasons in fact.

jsturdy: Then I think in this case, what you want is @mexanick's second point, the level here probably doesn't need to be `ERROR` (because it surely isn't treated as such here), but `WARNING` would be sufficient.
Then in a call that definitely doesn't care about spamming the log with these messages, the log level is changed before, and reverted after the call.
In the end, this codebase started as @mexanick's baby, so I'm happy for him to make the call.  If it were `cmsgemos` I'd say revert the change and do it differently..

bdorney: So if I understand the suggestion I should pass a log level object instead.  And this would then just be given to the log manager?

jsturdy: > So if I understand the suggestion I should pass a log level object instead. And this would then just be given to the log manager?

No, one of my (possibly elided) points in this comment is that the function signature (of such low-level functions) should *not* be changed.
If I understand how the logger is working, this is what type of changes I would approve:
* Low level functions (e.g., `readReg`/`writeReg`): one of:
  1) change the level of this message to `WARNING`
  1) leave the level of this message to `ERROR` (maybe my slight semantic preference, but at some point when we go through and fix the "error" condition handling, this can be revisited, re #85/#109)
  1) remove the printing of the message entirely, and throw an exception (preference in terms of API and usage, again, for revisitation as per #85/#109)
    * forcing the calling code to handle the printing of the log message, as well as the failure
    * (but generally, I would say that any changes to this low-level functionality should be first proposed and discussed in an issue, and not added as a rider to a different feature)
* Higher-level functions (where you want to turn on/off/control this message)
  1) call to `setLogLevel(ERROR)`  (or `setLogLevel(FATAL)`, if the second method is done above) before calls to the lower level function where this message is emitted
  1) call to  `setLogLevel(DEFAULT)` (maybe one has to do a `getLogLevel` before step 1 to get this value) after you no longer want to suppress the messages
* In your higher-level function, if you want to have a parameter that enables/disables this changing of the log level (analogue of your `printReadErrors`), I have less of an issue, because it doesn't touch the core implementation of the low-level functionality (which is one of my issues with the way this change was implemented)

bdorney: Okay; I will revert this change completely; the proposed solution you've (@jsturdy) made I think we should open a separate issue for and address it there.

bdorney:
@jsturdy @mexanick changes to `readReg` were reverted.  Ready for re-review.  Also this branch (`v1.1.X`) is not protected...not sure if that's intended.

jsturdy (approved):

Issue #100, opened 2019-03-08
[feature request] Add ADC readout from SCA


jsturdy:
## Brief summary of issue
This issue is to outline what is expected from the SCA ADC readout implementation.
The behaviour of the SCA is outlined in the GBT-SCA [users manual](https://espace.cern.ch/GBT-Project/GBT-SCA/Manuals/GBT-SCA-UserManual.pdf)

Wrapper functions to the various interfaces of the SCA have already been implemented (#97), and this issue is to simply utilize the ADC interface to read back any information that we are interested in.

### Types of issue
- [x] Feature request (request for change which adds functionality)

## Expected Behavior
* Functions that read a specific ADC, for all ADCs
* Function that will read all ADCs?

## Context (for feature requests)
* GE1/1 OptoHybrid V3 SCA pin connections:
![sca adcs](https://user-images.githubusercontent.com/5197820/54019699-1aa8ae80-418c-11e9-9a68-63106443e76a.png)

jsturdy:
@ram1123, based on [this](https://github.com/ram1123/ctp7_modules/commit/7140b8d80d073f2b4e0e9e2cfb911b6754494233) I have a couple of clarifying comments.
1) The expectation is that these SCA related functions will be implemented in the `amc/sca.cpp` unit
2) They will use the interface(s) that has(have) been implemented in #97, and as discussed in #93

ram1123:
I added the functionalities af requested in this issue #100.

You can have a look at them here: https://github.com/ram1123/ctp7_modules/commit/6f2b856a18ac9f82fdbb4178a3ab875db24d4689

Please let me know if it seems fine. Then I will make a pull request.

jsturdy:
A couple of quick questions/comments (others left inline on your code):
* your diff shows that the `config` submodule changed, did you make changes here?
  * remember, before making your PR you will have to do the following (nothing else should have gone into this branch, so these should actually not be necessary):
    1) `git rebase -i feature/sca-interface` (your changes should have started from this branch, and been made back to this branch)
    1) `git submodule update`

jsturdy:
I see that you've started from `develop`, so I'll just rebase `feature/sca-interface` onto that and you can make your PR to this branch

jsturdy:
[...]

PR #120, opened 2019-05-07
#100 : added 5 modules to read  SCAADC sensors

ram1123:
Added modules to read a specific SCA ADC or the similar type of SCA SCA sensors together, like temperature, voltage, etc.

## Description

There are 31 SCA ADC sensors defined ([here](https://github.com/ram1123/ctp7_modules/blob/f7590d67eeb06326ccaafe0d4aaf28fd1f52ecf8/include/amc/sca_enums.h#L211-L264)).  As requested in issue [#100](https://github.com/cms-gem-daq-project/ctp7_modules/issues/100), to create several function to read each ADC and all ADC, I added five different functions named `readSCAADCSensor`, `readSCAADCTemperatureSensors`, `readSCAADCVoltageSensors`,  `readSCAADCSignalStrengthSensors` and `readAllSCAADCSensors` to read individual sensors,  temperature sensors,  voltage related sensors, signal strength sensors and all sensors, respectively.

### Types of changes

## How Has This Been Tested?

All added modules are tested using a macro named `blaster.cxx`.

```cpp
#include <iostream>
#include <fstream>
#include <random>
#include <iomanip>
#include <stdio.h>
#include <stdlib.h>
#include <string.h>
#include <limits.h>
#include <errno.h>
#include <xhal/rpc/wiscrpcsvc.h>

using namespace wisc;

int main(int argc, char *argv[])
{
  RPCSvc rpc;
  try {
    rpc.connect(argv[1]);
  } catch (RPCSvc::ConnectionFailedException &e) {
    printf("Caught RPCErrorException: %s\n", e.message.c_str());
    return 1;
  } catch (RPCSvc::RPCException &e) {
    printf("Caught exception: %s\n", e.message.c_str());
    return 1;
  }

#define STANDARD_CATCH                                                  \
  catch (RPCSvc::NotConnectedException &e) {                            \
    printf("Caught NotConnectedException: %s\n", e.message.c_str());    \
    return 1;                                                           \
  } catch (RPCSvc::RPCErrorException &e) {                              \
    printf("Caught RPCErrorException: %s\n", e.message.c_str());        \
    return 1;                                                           \
  } catch (RPCSvc::RPCException &e) {                                   \
    printf("Caught exception: %s\n", e.message.c_str());                \
    return 1;                                                           \
  }

#define ASSERT(x) do {                                                  \
    if (!(x)) {                                                         \
      printf("Assertion Failed on line %u: %s\n", __LINE__, #x);        \
      return 1;                                                         \
    }                                                                   \
  } while (0)

  RPCMsg req, rsp;

  try {
    ASSERT(rpc.load_module("amc", "amc v1.0.1"));
  } STANDARD_CATCH;

  //req = RPCMsg("amc.readSCAADCSensor");
  req = RPCMsg("amc.readSCAADCTemperatureSensors");

  //errno = 0;
  //uint32_t sz = strtoul(argv[2], nullptr, 16);

  //if (errno != 0 /*|| *p != '\0'*/ || sz > INT_MAX) {
  //  errno = 0;
  //  sz = strtoul(argv[2], nullptr, 10);
  //  if (errno != 0 /*|| *p != '\0'*/ || sz > INT_MAX) {
  //    sz = 0;
  //  } else {
  //    // pass
  //  }
  //} else {
  //  // pass
  //}

  errno = 0;
  uint32_t ty = strtoul(argv[2], nullptr, 16);

  if (errno != 0 /*|| *p != '\0'*/ || ty > INT_MAX) {
    ty = 0;
  } else {
    // pass
  }

  //req.set_word("ch", 	sz);
  req.set_word("ohMask",ty);

  try {
    std::cout << "calling amc.readSCAADCTemperatureSensors on " << std::string(argv[1]) << " with " << ty << "...";
    //std::cout << "calling amc.readSCAADCTemperatureSensors on " << std::string(argv[1]) << " with " << sz << " and " << ty << "...";
    rsp = rpc.call_method(req);
    std::cout << "done!" << std::endl;
    if (rsp.get_key_exists("error")) {
      std::cout << "RPC ERROR:" << rsp.get_string("error") << std::endl;
      return 0;
    }

  
  return 0;
}
```

then running its executable after setting the environment and monitoring the log.

## Checklist:

### DIFF
--- a/include/amc/sca.h
+++ b/include/amc/sca.h
@@ -174,4 +174,60 @@ void readSCAChipID(const RPCMsg *request, RPCMsg *response);
 void readSCASEUCounter(const RPCMsg *request, RPCMsg *response);
 void resetSCASEUCounter(const RPCMsg *request, RPCMsg *response);
 
+/*!
+ *  \fn void readSCAADCSensor(const RPCMsg *request, RPCMsg *response)
+ *  \brief Read the ADC channels
+ *
+ *  - ch : Name of ADC channel to read from SCAADCChannelT
+ *  - ohMask : This specifies which OH's to read from 
+ *
+ *  \param request RPC request message
+ *  \param response RPC response message
+ */
+void readSCAADCSensor(const RPCMsg *request, RPCMsg *response);
+
+/*!
+ *  \fn void readSCAADCTemperatureSensors(const RPCMsg *request, RPCMsg *response)
+ *  \brief Read all of temperature sensor channels. They are 00, 04, 07, 08 and 1F.
+ *
+ *  - ohMask : This specifies which OH's to read from 
+ *
+ *  \param request RPC request message
+ *  \param response RPC response message
+ */
+void readSCAADCTemperatureSensors(const RPCMsg *request, RPCMsg *response);
+
+/*!
+ *  \fn void readSCAADCVoltageSensors(const RPCMsg *request, RPCMsg *response)
+ *  \brief Read all the voltages channels. They are 1B, 1E, 11, 0E, 18 and 0F. 
+ *
+ *  - ohMask : This specifies which OH's to read from 
+ *
+ *  \param request RPC request message
+ *  \param response RPC response message
+ */
+void readSCAADCVoltageSensors(const RPCMsg *request, RPCMsg *response);
+
+/*!
+ *  \fn void readSCAADCSignalStrengthSensors(const RPCMsg *request, RPCMsg *response)
+ *  \brief Read the signal strength channels. They are 15, 13 and 12. 
+ *
+ *  - ohMask : This specifies which OH's to read from 
+ *
+ *  \param request RPC request message
+ *  \param response RPC response message
+ */
+void readSCAADCSignalStrengthSensors(const RPCMsg *request, RPCMsg *response);
+
+/*!
+ *  \fn void readAllSCAADCSensors(const RPCMsg *request, RPCMsg *response)
+ *  \brief Read all connected ADC channels. 
+ *
+ *  - ohMask : This specifies which OH's to read from 
+ *
+ *  \param request RPC request message
+ *  \param response RPC response message
+ */
+void readAllSCAADCSensors(const RPCMsg *request, RPCMsg *response);
+
 #endif
--- a/include/amc/sca.h
+++ b/include/amc/sca.h
@@ -177,56 +177,86 @@ void resetSCASEUCounter(const RPCMsg *request, RPCMsg *response);
 /*!
  *  \fn void readSCAADCSensor(const RPCMsg *request, RPCMsg *response)
  *  \brief Read individual SCA ADC sensor
- *  \param[in] request RPC response message
+ *  \param[in] request RPC response message with the following keys:
  *  	- `word ohMask` : This specifies which OH's to read from
  *  	- `word ch` : Name of SCA ADC sensor to read
  *
- *  \param[out] response RPC response message
- *  	- `word_array data` : Array of word contaning the information of masked/unmasked OH, link ID, SCA ADC sensor number and the data. This is formatted like: <OH masked 1, otherwise 0 27><link ID 26:24><0s 23:21><ADC ID 20:16><0's 15:12><data 11:0>
+ *  \param[out] response RPC response message with the following keys:
+ *  	- `word_array data` : ADC data is returned as an array of 32-bit words formatted as:
+ *  			      bit [27]: data present
+ *  			      bits [26:24]: link ID
+ *  			      bits [23:21]: constant 0's
+ *  			      bits [20:16]: ADC channel ID
+ *  			      bits [15:12]: constant 0's
+ *  			      bits [11:0]: ADC data
  */
 void readSCAADCSensor(const RPCMsg *request, RPCMsg *response);
 
 /*!
  *  \fn void readSCAADCTemperatureSensors(const RPCMsg *request, RPCMsg *response)
  *  \brief Read all SCA ADC temperature sensors. They are 0x00, 0x04, 0x07, and 0x08.
- *  \param[in] request RPC response message
+ *  \param[in] request RPC response message with the following keys:
  *  	- `word ohMask` : This specifies which OH's to read from
  *
- *  \param[out] response RPC response message
- *  	- `word_array data` : Array of word contaning the information of masked/unmasked OH, link ID, SCA ADC sensor number and the data. This is formatted like: <OH masked 1, otherwise 0 27><link ID 26:24><0s 23:21><ADC ID 20:16><0's 15:12><data 11:0>
+ *  \param[out] response RPC response message with the following keys:
+ *  	- `word_array data` : ADC data is returned as an array of 32-bit words formatted as:
+ *  			      bit [27]: data present
+ *  			      bits [26:24]: link ID
+ *  			      bits [23:21]: constant 0's
+ *  			      bits [20:16]: ADC channel ID
+ *  			      bits [15:12]: constant 0's
+ *  			      bits [11:0]: ADC data
  */
 void readSCAADCTemperatureSensors(const RPCMsg *request, RPCMsg *response);
 
 /*!
  *  \fn void readSCAADCVoltageSensors(const RPCMsg *request, RPCMsg *response)
  *  \brief Read all SCA ADC voltages sensors. They are 1B, 1E, 11, 0E, 18 and 0F. 
- *  \param[in] request RPC response message
+ *  \param[in] request RPC response message with the following keys:
  *  	- `word ohMask` : This specifies which OH's to read from
  *
- *  \param[out] response RPC response message
- *  	- `word_array data` : Array of word contaning the information of masked/unmasked OH, link ID, SCA ADC sensor number and the data. This is formatted like: <OH masked 1, otherwise 0 27><link ID 26:24><0s 23:21><ADC ID 20:16><0's 15:12><data 11:0>
+ *  \param[out] response RPC response message with the following keys:
+ *  	- `word_array data` : ADC data is returned as an array of 32-bit words formatted as:
+ *  			      bit [27]: data present
+ *  			      bits [26:24]: link ID
+ *  			      bits [23:21]: constant 0's
+ *  			      bits [20:16]: ADC channel ID
+ *  			      bits [15:12]: constant 0's
+ *  			      bits [11:0]: ADC data
  */
 void readSCAADCVoltageSensors(const RPCMsg *request, RPCMsg *response);
 
 /*!
  *  \fn void readSCAADCSignalStrengthSensors(const RPCMsg *request, RPCMsg *response)
  *  \brief Read the SCA ADC signal strength sensors. They are 15, 13 and 12. 
- *  \param[in] request RPC response message
+ *  \param[in] request RPC response message with the following keys:
  *  	- `word ohMask` : This specifies which OH's to read from
  *
- *  \param[out] response RPC response message
- *  	- `word_array data` : Array of word contaning the information of masked/unmasked OH, link ID, SCA ADC sensor number and the data. This is formatted like: <OH masked 1, otherwise 0 27><link ID 26:24><0s 23:21><ADC ID 20:16><0's 15:12><data 11:0>
+ *  \param[out] response RPC response message with the following keys:
+ *  	- `word_array data` : ADC data is returned as an array of 32-bit words formatted as:
+ *  			      bit [27]: data present
+ *  			      bits [26:24]: link ID
+ *  			      bits [23:21]: constant 0's
+ *  			      bits [20:16]: ADC channel ID
+ *  			      bits [15:12]: constant 0's
+ *  			      bits [11:0]: ADC data
  */
 void readSCAADCSignalStrengthSensors(const RPCMsg *request, RPCMsg *response);
 
 /*!
  *  \fn void readAllSCAADCSensors(const RPCMsg *request, RPCMsg *response)
  *  \brief Read all connected SCA ADC sensors. 
- *  \param[in] request RPC response message
+ *  \param[in] request RPC response message with the following keys:
  *  	- `word ohMask` : This specifies which OH's to read from
  *
- *  \param[out] response RPC response message
- *  	- `word_array data` : Array of word contaning the information of masked/unmasked OH, link ID, SCA ADC sensor number and the data. This is formatted like: <OH masked 1, otherwise 0 27><link ID 26:24><0s 23:21><ADC ID 20:16><0's 15:12><data 11:0>
+ *  \param[out] response RPC response message with the following keys:
+ *  	- `word_array data` : ADC data is returned as an array of 32-bit words formatted as:
+ *  			      bit [27]: data present
+ *  			      bits [26:24]: link ID
+ *  			      bits [23:21]: constant 0's
+ *  			      bits [20:16]: ADC channel ID
+ *  			      bits [15:12]: constant 0's
+ *  			      bits [11:0]: ADC data
  */
 void readAllSCAADCSensors(const RPCMsg *request, RPCMsg *response);
 
--- a/include/amc/sca.h
+++ b/include/amc/sca.h
@@ -176,57 +176,57 @@ void resetSCASEUCounter(const RPCMsg *request, RPCMsg *response);
 
 /*!
  *  \fn void readSCAADCSensor(const RPCMsg *request, RPCMsg *response)
- *  \brief Read the ADC channels
+ *  \brief Read individual SCA ADC sensor
+ *  \param[in] request RPC response message
+ *  	- `word ohMask` : This specifies which OH's to read from
+ *  	- `word ch` : Name of SCA ADC sensor to read
  *
- *  - ch : Name of ADC channel to read from SCAADCChannelT
- *  - ohMask : This specifies which OH's to read from 
- *
- *  \param request RPC request message
- *  \param response RPC response message
+ *  \param[out] response RPC response message
+ *  	- `word_array data` : Array of word contaning the information of masked/unmasked OH, link ID, SCA ADC sensor number and the data. This is formatted like: <OH masked 1, otherwise 0 27><link ID 26:24><0s 23:21><ADC ID 20:16><0's 15:12><data 11:0>
  */
 void readSCAADCSensor(const RPCMsg *request, RPCMsg *response);
 
 /*!
  *  \fn void readSCAADCTemperatureSensors(const RPCMsg *request, RPCMsg *response)
- *  \brief Read all of temperature sensor channels. They are 00, 04, 07, 08 and 1F.
+ *  \brief Read all SCA ADC temperature sensors. They are 0x00, 0x04, 0x07, and 0x08.
+ *  \param[in] request RPC response message
+ *  	- `word ohMask` : This specifies which OH's to read from
  *
- *  - ohMask : This specifies which OH's to read from 
- *
- *  \param request RPC request message
- *  \param response RPC response message
+ *  \param[out] response RPC response message
+ *  	- `word_array data` : Array of word contaning the information of masked/unmasked OH, link ID, SCA ADC sensor number and the data. This is formatted like: <OH masked 1, otherwise 0 27><link ID 26:24><0s 23:21><ADC ID 20:16><0's 15:12><data 11:0>
  */
 void readSCAADCTemperatureSensors(const RPCMsg *request, RPCMsg *response);
 
 /*!
  *  \fn void readSCAADCVoltageSensors(const RPCMsg *request, RPCMsg *response)
- *  \brief Read all the voltages channels. They are 1B, 1E, 11, 0E, 18 and 0F. 
+ *  \brief Read all SCA ADC voltages sensors. They are 1B, 1E, 11, 0E, 18 and 0F. 
+ *  \param[in] request RPC response message
+ *  	- `word ohMask` : This specifies which OH's to read from
  *
- *  - ohMask : This specifies which OH's to read from 
- *
- *  \param request RPC request message
- *  \param response RPC response message
+ *  \param[out] response RPC response message
+ *  	- `word_array data` : Array of word contaning the information of masked/unmasked OH, link ID, SCA ADC sensor number and the data. This is formatted like: <OH masked 1, otherwise 0 27><link ID 26:24><0s 23:21><ADC ID 20:16><0's 15:12><data 11:0>
  */
 void readSCAADCVoltageSensors(const RPCMsg *request, RPCMsg *response);
 
 /*!
  *  \fn void readSCAADCSignalStrengthSensors(const RPCMsg *request, RPCMsg *response)
- *  \brief Read the signal strength channels. They are 15, 13 and 12. 
+ *  \brief Read the SCA ADC signal strength sensors. They are 15, 13 and 12. 
+ *  \param[in] request RPC response message
+ *  	- `word ohMask` : This specifies which OH's to read from
  *
- *  - ohMask : This specifies which OH's to read from 
- *
- *  \param request RPC request message
- *  \param response RPC response message
+ *  \param[out] response RPC response message
+ *  	- `word_array data` : Array of word contaning the information of masked/unmasked OH, link ID, SCA ADC sensor number and the data. This is formatted like: <OH masked 1, otherwise 0 27><link ID 26:24><0s 23:21><ADC ID 20:16><0's 15:12><data 11:0>
  */
 void readSCAADCSignalStrengthSensors(const RPCMsg *request, RPCMsg *response);
 
 /*!
  *  \fn void readAllSCAADCSensors(const RPCMsg *request, RPCMsg *response)
- *  \brief Read all connected ADC channels. 
+ *  \brief Read all connected SCA ADC sensors. 
+ *  \param[in] request RPC response message
+ *  	- `word ohMask` : This specifies which OH's to read from
  *
- *  - ohMask : This specifies which OH's to read from 
- *
- *  \param request RPC request message
- *  \param response RPC response message
+ *  \param[out] response RPC response message
+ *  	- `word_array data` : Array of word contaning the information of masked/unmasked OH, link ID, SCA ADC sensor number and the data. This is formatted like: <OH masked 1, otherwise 0 27><link ID 26:24><0s 23:21><ADC ID 20:16><0's 15:12><data 11:0>
  */
 void readAllSCAADCSensors(const RPCMsg *request, RPCMsg *response);
 
--- a/include/amc/sca_enums.h
+++ b/include/amc/sca_enums.h
@@ -234,7 +234,7 @@ class SCASettings {
       // SCA ADC temperature sensors
       VTTX_CSC_PT100 = 0x00, ///< Transceiver block next to the CSC VTTX
       VTTX_GEM_PT100 = 0x04, ///< Transceiver block next to the GEM VTTX
-      SCA_PT100      = 0x07, ///< SCA temperature sensor
+      GBT0_PT100      = 0x07, ///< SCA temperature sensor
       V6_FPGA_PT100  = 0x08, ///< Virtex6 temperature sensor
 
       // SCA ADC signal strength sensors
@@ -261,6 +261,18 @@ class SCASettings {
       //ADC_CH28 = 0x1C, ///< ADC channel 28
       //ADC_CH29 = 0x1D, ///< ADC channel 29
     } ADCChannel;
+
+    //static constexpr bool useCurrentSource(EADCChannel sensor)
+    static bool useCurrentSource(EADCChannel sensor)
+    {
+       switch (sensor) {
+          case (VTTX_CSC_PT100): return true;
+	  case (VTTX_GEM_PT100): return true;
+	  case (GBT0_PT100):	 return true;
+	  case (V6_FPGA_PT100):	 return true;
+	  default:		 return false;
+       }
+    }
   };  // struct ADCChannel
 
 };
--- a/include/amc/sca_enums.h
+++ b/include/amc/sca_enums.h
@@ -231,38 +231,35 @@ class SCASettings {
       FPGA_IO_V2P5   = 15, ///< 2.5V used by FPGA I/O and VTRXs/VTTXs
       VTTX_VTRX_V2P5 = 15, ///< 2.5V used by FPGA I/O and VTRXs/VTTXs
 
-      ADC_CH00 = 0x00, ///< ADC channel 00, Temp sensor
-      ADC_CH01 = 0x01, ///< ADC channel 01, NOT CONNECTED
-      ADC_CH02 = 0x02, ///< ADC channel 02, NOT CONNECTED
-      ADC_CH03 = 0x03, ///< ADC channel 03, NOT CONNECTED
-      ADC_CH04 = 0x04, ///< ADC channel 04, Temp sensor
-      ADC_CH05 = 0x05, ///< ADC channel 05, NOT CONNECTED
-      ADC_CH06 = 0x06, ///< ADC channel 06
-      ADC_CH07 = 0x07, ///< ADC channel 07, Temp sensor
-      ADC_CH08 = 0x08, ///< ADC channel 08, Temp sensor
-      ADC_CH09 = 0x09, ///< ADC channel 09
-      ADC_CH10 = 0x0A, ///< ADC channel 10
-      ADC_CH11 = 0x0B, ///< ADC channel 11
-      ADC_CH12 = 0x0C, ///< ADC channel 12, NOT CONNECTED
-      ADC_CH13 = 0x0D, ///< ADC channel 13
-      ADC_CH14 = 0x0E, ///< ADC channel 14, 1.8V
-      ADC_CH15 = 0x0F, ///< ADC channel 15, 2.5V I/O
-      ADC_CH16 = 0x10, ///< ADC channel 16
-      ADC_CH17 = 0x11, ///< ADC channel 17, internal 1.0V
-      ADC_CH18 = 0x12, ///< ADC channel 18, VTRx rssi3
-      ADC_CH19 = 0x13, ///< ADC channel 19, VTRx rssi2
-      ADC_CH20 = 0x14, ///< ADC channel 20
-      ADC_CH21 = 0x15, ///< ADC channel 21, VTRx rssi1
-      ADC_CH22 = 0x16, ///< ADC channel 22
-      ADC_CH23 = 0x17, ///< ADC channel 23
-      ADC_CH24 = 0x18, ///< ADC channel 24, 1.5V
-      ADC_CH25 = 0x19, ///< ADC channel 25
-      ADC_CH26 = 0x1A, ///< ADC channel 26
-      ADC_CH27 = 0x1B, ///< ADC channel 27, AVCCN
-      ADC_CH28 = 0x1C, ///< ADC channel 28
-      ADC_CH29 = 0x1D, ///< ADC channel 29
-      ADC_CH30 = 0x1E, ///< ADC channel 30, AVTTN
-      ADC_CH31 = 0x1F, ///< ADC channel 31
+      // SCA ADC temperature sensors
+      VTTX_CSC 	     = 0x00, ///< Transceiver block next to the CSC VTTX
+      VTTX_GEM       = 0x04, ///< Transceiver block next to the GEM VTTX
+      SCA_PT100      = 0x07, ///< SCA temperature sensor
+      V6_FPGA_PT100  = 0x08, ///< Virtex6 temperature sensor
+
+      // SCA ADC signal strength sensors
+      VTRX_RSSI3     = 0x12, ///< ADC channel 18, VTRx rssi3
+      VTRX_RSSI2     = 0x13, ///< ADC channel 19, VTRx rssi2
+      VTRX_RSSI1     = 0x15, ///< ADC channel 21, VTRx rssi1
+
+      // Not connected SCA ADC sensors
+      //ADC_CH01 = 0x01, ///< ADC channel 01, NOT CONNECTED
+      //ADC_CH02 = 0x02, ///< ADC channel 02, NOT CONNECTED
+      //ADC_CH03 = 0x03, ///< ADC channel 03, NOT CONNECTED
+      //ADC_CH05 = 0x05, ///< ADC channel 05, NOT CONNECTED
+      //ADC_CH09 = 0x09, ///< ADC channel 09
+      //ADC_CH10 = 0x0A, ///< ADC channel 10
+      //ADC_CH11 = 0x0B, ///< ADC channel 11
+      //ADC_CH12 = 0x0C, ///< ADC channel 12, NOT CONNECTED
+      //ADC_CH13 = 0x0D, ///< ADC channel 13
+      //ADC_CH16 = 0x10, ///< ADC channel 16
+      //ADC_CH20 = 0x14, ///< ADC channel 20
+      //ADC_CH22 = 0x16, ///< ADC channel 22
+      //ADC_CH23 = 0x17, ///< ADC channel 23
+      //ADC_CH25 = 0x19, ///< ADC channel 25
+      //ADC_CH26 = 0x1A, ///< ADC channel 26
+      //ADC_CH28 = 0x1C, ///< ADC channel 28
+      //ADC_CH29 = 0x1D, ///< ADC channel 29
     } ADCChannel;
   };  // struct ADCChannel
 
--- a/include/amc/sca_enums.h
+++ b/include/amc/sca_enums.h
@@ -265,13 +265,13 @@ class SCASettings {
     //static constexpr bool useCurrentSource(EADCChannel sensor)
     static bool useCurrentSource(EADCChannel sensor)
     {
-       switch (sensor) {
-          case (VTTX_CSC_PT100): return true;
-	  case (VTTX_GEM_PT100): return true;
-	  case (GBT0_PT100):	 return true;
-	  case (V6_FPGA_PT100):	 return true;
-	  default:		 return false;
-       }
+        switch (sensor) {
+            case (VTTX_CSC_PT100):	return true;
+            case (VTTX_GEM_PT100):	return true;
+            case (GBT0_PT100):		return true;
+            case (V6_FPGA_PT100):	return true;
+            default:			return false;
+        }
     }
   };  // struct ADCChannel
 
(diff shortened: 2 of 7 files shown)


include/amc/sca_enums.h
```diff
@@ -231,38 +231,35 @@ class SCASettings {
...
-      ADC_CH24 = 0x18, ///< ADC channel 24, 1.5V
-      ADC_CH25 = 0x19, ///< ADC channel 25
-      ADC_CH26 = 0x1A, ///< ADC channel 26
-      ADC_CH27 = 0x1B, ///< ADC channel 27, AVCCN
-      ADC_CH28 = 0x1C, ///< ADC channel 28
-      ADC_CH29 = 0x1D, ///< ADC channel 29
-      ADC_CH30 = 0x1E, ///< ADC channel 30, AVTTN
-      ADC_CH31 = 0x1F, ///< ADC channel 31
+      // SCA ADC temperature sensors
+      VTTX_CSC 	     = 0x00, ///< Transceiver block next to the CSC VTTX
```
jsturdy: Oops, these should have `_PT100`, like the other temp sensors

include/utils.h
```diff
@@ -80,6 +80,14 @@ std::string serialize(xhal::utils::Node n);
     response->set_string("error", message);             \
     return error_code; }
 
+/*! \fn uint32_t bitCheck(uint32_t ohMask, int bit)
+ *  \brief return 1 if the particular bit is 1 else 0
+ *
+ *  \param ohMask: This specifies which OH's to read from 
+ *  \param bit: integer that specifies a particular bit position
+ */
+uint32_t bitCheck(uint32_t ohMask, int bit);
```
jsturdy: I would generalize this such that the input parameter is **not** related to the `OptoHybrid`, rather, it is simply a mask/word.
Functionally no change, but semantically, it is more general.

src/amc/sca.cpp
```diff
@@ -235,3 +240,145 @@ void resetSCASEUCounter(const RPCMsg *request, RPCMsg *response)
...
+  uint32_t ohMask = request->get_word("ohMask");
+  SCAADCChannelT  ch = static_cast<SCAADCChannelT>(request->get_word("ch"));
+  
+  std::vector<uint32_t> result;
+  std::vector<uint32_t> outData;
+
+  result = scaADCCommand(&la, ch, 0, 0, ohMask);
+
+  int count = 0;
+  for(auto const& val : result){
```
jsturdy: please use (note the spaces) everywhere inside functions
```c++
for (specifier) {

if (condition) {

while (condition) {
```

src/amc/sca.cpp
```diff
@@ -235,3 +240,145 @@ void resetSCASEUCounter(const RPCMsg *request, RPCMsg *response)
...
+  int count;
+  std::vector<uint32_t> result;
+  std::vector<uint32_t> outData;
+  for (SCAADCChannelT channelName = SCAADCChannel::VTTX_CSC; channelName<=SCAADCChannel::SCA_TEMP; channelName=SCAADCChannelT(channelName+1)) {
+    result = scaADCCommand(&la, channelName, 0, 0, ohMask);
+    count = 0;
+    for(auto const& val : result){
+      LOGGER->log_message(LogManager::DEBUG, stdsprintf("Reading of OH%i, SCA-ADC channel 0x%x = %i ",count, channelName, val));
+      outData.push_back((bitCheck(ohMask, count) << 27) | (count<<24) | (channelName<<16) | val);
+      count++;
```
jsturdy: prefer pre-increment, as mentioned several times previously

src/amc/sca.cpp
```diff
@@ -235,3 +240,145 @@ void resetSCASEUCounter(const RPCMsg *request, RPCMsg *response)
...
+  rtxn.abort();
+}
+
+void readAllSCAADCSensors(const RPCMsg *request, RPCMsg *response)
+{
+  GETLOCALARGS(response);
+
+  uint32_t ohMask = request->get_word("ohMask");
+
+  int count;
```
jsturdy: `count` is a rather poorly chosen name, maybe `ohIdx`?

src/utils.cpp
```diff
@@ -128,6 +128,15 @@ void readRegFromDB(const RPCMsg *request, RPCMsg *response)
   rtxn.abort();
 }
 
+uint32_t bitCheck(uint32_t ohMask, int bit)
+{
+  if ( (ohMask >> bit) & 1 ){
```
jsturdy: You should probably also have a safety check to ensure you don't shift off the end if `bit > 31`

src/amc/sca.cpp
```diff
@@ -235,3 +240,145 @@ void resetSCASEUCounter(const RPCMsg *request, RPCMsg *response)
...
+
+  uint32_t ohMask = request->get_word("ohMask");
+  SCAADCChannelT  ch = static_cast<SCAADCChannelT>(request->get_word("ch"));
+  
+  std::vector<uint32_t> result;
+  std::vector<uint32_t> outData;
+
+  result = scaADCCommand(&la, ch, 0, 0, ohMask);
+
+  int count = 0;
```
jsturdy: `count` is a rather poorly chosen name, maybe `ohIdx`? (here and elsewhere where it is used the same way)

src/amc/sca.cpp
```diff
@@ -120,7 +120,7 @@ std::vector<uint32_t> scaGPIOCommandLocal(localArgs* la, SCAGPIOCommandT const&
   return reply;
 }
 
-std::vector<uint32_t> scaADCCommand(localArgs* la, SCAADCChannelT const& ch, SCAADCCommandT const& cmd, uint8_t const& len, uint32_t data, uint16_t const& ohMask)
+std::vector<uint32_t> scaADCCommand(localArgs* la, SCAADCChannelT const& ch, uint8_t const& len, uint32_t data, uint16_t const& ohMask)
```
jsturdy: I think you can also remove `len` and `data` from this function (and your subsequent usage)
I don't think for the ADC we'll need to change these values

include/amc/sca.h
```diff
@@ -174,4 +174,60 @@ void readSCAChipID(const RPCMsg *request, RPCMsg *response);
...
 
+/*!
+ *  \fn void readSCAADCSensor(const RPCMsg *request, RPCMsg *response)
+ *  \brief Read the ADC channels
+ *
+ *  - ch : Name of ADC channel to read from SCAADCChannelT
+ *  - ohMask : This specifies which OH's to read from 
+ *
+ *  \param request RPC request message
+ *  \param response RPC response message
```
jsturdy: Here, and other new functions, state the output key name, type, and describe the format (as mentioned previously, the documentation style utilized by @lpetre-ulb in `gbt.h` is rather quite good)

include/amc/sca.h
```diff
@@ -174,4 +174,60 @@ void readSCAChipID(const RPCMsg *request, RPCMsg *response);
 void readSCASEUCounter(const RPCMsg *request, RPCMsg *response);
 void resetSCASEUCounter(const RPCMsg *request, RPCMsg *response);
 
+/*!
+ *  \fn void readSCAADCSensor(const RPCMsg *request, RPCMsg *response)
+ *  \brief Read the ADC channels
+ *
+ *  - ch : Name of ADC channel to read from SCAADCChannelT
```
jsturdy: Here, and other new functions, move this into the `\param request` documentation section, state the key name and type

jsturdy (changes requested):

include/amc/sca.h
ram1123: Is the following message fine?

```cpp
/*!
 *  \fn void readAllSCAADCSensors(const RPCMsg *request, RPCMsg *response)
 *  \brief Read all connected SCA ADC sensors. 
 *  \param[in] request RPC response message
 *  \param[out] response RPC response message
 *
 *  The method expects the following RPC keys:
 *  - `word ohMask` : This specifies which OH's to read from
 *
 *  The method returns the following RPC keys:
 *  - `word_array data` : Array of word contaning the information of masked/unmasked OH, link ID, SCA ADC sensor number and the data. This is formatted like: <OH masked 1, otherwise 0 27><link ID 26:24><0s 23:21><ADC ID 20:16><0's 15:12><data 11:0>
 */
void readAllSCAADCSensors(const RPCMsg *request, RPCMsg *response);
```

src/utils.cpp
ram1123: Is this fine?

```cpp
uint32_t bitCheck(uint32_t word, int bit)
{
  if (bit > 31)
    throw std::invalid_argument("In function: uint32_t bitCheck(uint32_t word, int bit), the shifting bit (named bit) should not be more than 31");
  if ( (word >> bit) & 1 ) {
    return 1;
  } else {
    return 0;
  }
}
```

include/amc/sca.h
jsturdy: Yes, but please group them with the `\param` block to which they belong, i.e.,
```c++
\param[in] request RPC request message, with the following keys
           * `word ohMask` : blah
\param[out] response RPC response message, with the following keys
           * `word_array data` : blah
```

src/utils.cpp
jsturdy: I'd shorten the message to:
`Invalid request to shift word by more than 31 bits`

include/amc/sca.h
```diff
@@ -174,4 +174,60 @@ void readSCAChipID(const RPCMsg *request, RPCMsg *response);
...
 
+/*!
+ *  \fn void readSCAADCSensor(const RPCMsg *request, RPCMsg *response)
+ *  \brief Read individual SCA ADC sensor
+ *  \param[in] request RPC response message
+ *  	- `word ohMask` : This specifies which OH's to read from
+ *  	- `word ch` : Name of SCA ADC sensor to read
+ *
+ *  \param[out] response RPC response message
+ *  	- `word_array data` : Array of word contaning the information of masked/unmasked OH, link ID, SCA ADC sensor number and the data. This is formatted like: <OH masked 1, otherwise 0 27><link ID 26:24><0s 23:21><ADC ID 20:16><0's 15:12><data 11:0>
```
bdorney: From the text here I don't understand how each output data word is to be decoded.  Specifically:

```
OH masked 1, otherwise 0 27
```

include/amc/sca.h
```diff
@@ -174,4 +174,60 @@ void readSCAChipID(const RPCMsg *request, RPCMsg *response);
...
+void readSCAADCSensor(const RPCMsg *request, RPCMsg *response);
+
+/*!
+ *  \fn void readSCAADCTemperatureSensors(const RPCMsg *request, RPCMsg *response)
+ *  \brief Read all SCA ADC temperature sensors. They are 0x00, 0x04, 0x07, and 0x08.
+ *  \param[in] request RPC response message
+ *  	- `word ohMask` : This specifies which OH's to read from
+ *
+ *  \param[out] response RPC response message
+ *  	- `word_array data` : Array of word contaning the information of masked/unmasked OH, link ID, SCA ADC sensor number and the data. This is formatted like: <OH masked 1, otherwise 0 27><link ID 26:24><0s 23:21><ADC ID 20:16><0's 15:12><data 11:0>
```
bdorney: same

include/amc/sca.h
```diff
@@ -174,4 +174,60 @@ void readSCAChipID(const RPCMsg *request, RPCMsg *response);
...
+void readSCAADCTemperatureSensors(const RPCMsg *request, RPCMsg *response);
+
+/*!
+ *  \fn void readSCAADCVoltageSensors(const RPCMsg *request, RPCMsg *response)
+ *  \brief Read all SCA ADC voltages sensors. They are 1B, 1E, 11, 0E, 18 and 0F. 
+ *  \param[in] request RPC response message
+ *  	- `word ohMask` : This specifies which OH's to read from
+ *
+ *  \param[out] response RPC response message
+ *  	- `word_array data` : Array of word contaning the information of masked/unmasked OH, link ID, SCA ADC sensor number and the data. This is formatted like: <OH masked 1, otherwise 0 27><link ID 26:24><0s 23:21><ADC ID 20:16><0's 15:12><data 11:0>
```
bdorney: same

include/amc/sca.h
```diff
@@ -174,4 +174,60 @@ void readSCAChipID(const RPCMsg *request, RPCMsg *response);
...
+void readSCAADCVoltageSensors(const RPCMsg *request, RPCMsg *response);
+
+/*!
+ *  \fn void readSCAADCSignalStrengthSensors(const RPCMsg *request, RPCMsg *response)
+ *  \brief Read the SCA ADC signal strength sensors. They are 15, 13 and 12. 
+ *  \param[in] request RPC response message
+ *  	- `word ohMask` : This specifies which OH's to read from
+ *
+ *  \param[out] response RPC response message
+ *  	- `word_array data` : Array of word contaning the information of masked/unmasked OH, link ID, SCA ADC sensor number and the data. This is formatted like: <OH masked 1, otherwise 0 27><link ID 26:24><0s 23:21><ADC ID 20:16><0's 15:12><data 11:0>
```
bdorney: Assume my top level comment applied to all instances of this documentation line.

include/amc/sca_enums.h
```diff
@@ -231,38 +231,35 @@ class SCASettings {
...
+      //ADC_CH12 = 0x0C, ///< ADC channel 12, NOT CONNECTED
+      //ADC_CH13 = 0x0D, ///< ADC channel 13
+      //ADC_CH16 = 0x10, ///< ADC channel 16
+      //ADC_CH20 = 0x14, ///< ADC channel 20
+      //ADC_CH22 = 0x16, ///< ADC channel 22
+      //ADC_CH23 = 0x17, ///< ADC channel 23
+      //ADC_CH25 = 0x19, ///< ADC channel 25
+      //ADC_CH26 = 0x1A, ///< ADC channel 26
+      //ADC_CH28 = 0x1C, ///< ADC channel 28
+      //ADC_CH29 = 0x1D, ///< ADC channel 29
```
bdorney, line 262: If they are not connected then we should just delete these lines and not push large comment blocks to the repo.

However if we expect that these ADC channels would be accessible on GE2/1 then they should not be uncommented but left with a note of "GE2/1 only"

src/amc/sca.cpp
```diff
@@ -120,7 +120,7 @@ std::vector<uint32_t> scaGPIOCommandLocal(localArgs* la, SCAGPIOCommandT const&
   return reply;
 }
 
-std::vector<uint32_t> scaADCCommand(localArgs* la, SCAADCChannelT const& ch, SCAADCCommandT const& cmd, uint8_t const& len, uint32_t data, uint16_t const& ohMask)
+std::vector<uint32_t> scaADCCommand(localArgs* la, SCAADCChannelT const& ch, uint16_t const& ohMask)
```
bdorney, line 124: This is a local function? e.g. it doesn't take the RPC message/response parameters?

If so would suggest to change name to follow previous naming convention; e.g. `scaADCCommandLocal`.

src/amc/sca.cpp
```diff
@@ -129,11 +129,16 @@ std::vector<uint32_t> scaADCCommand(localArgs* la, SCAADCChannelT const& ch, SCA
   // select the ADC channel
   sendSCACommand(la, SCAChannel::ADC, SCAADCCommand::ADC_W_MUX, 0x4, ch, ohMask);
 
-  // enable/disable the current sink
-  sendSCACommand(la, SCAChannel::ADC, SCAADCCommand::ADC_W_CURR, 0x4, 0x1<<ch, ohMask);
+
+  // enable the current sink if reading the temperature ADCs
+  if (ch == SCAADCChannel::VTTX_CSC_PT100 || ch == SCAADCChannel::VTTX_GEM_PT100 || ch == SCAADCChannel::SCA_PT100 || ch == SCAADCChannel::V6_FPGA_PT100 )	// Hardcoded channel numbers
```
bdorney: This could be simplified with the unordered map I suggest above, e.g.

```C++
if (map_connChan.count(ch))
{
   //do stuff
}
```

include/amc/sca_enums.h
```diff
@@ -231,38 +231,35 @@ class SCASettings {
...
+      // SCA ADC temperature sensors
+      VTTX_CSC_PT100 = 0x00, ///< Transceiver block next to the CSC VTTX
+      VTTX_GEM_PT100 = 0x04, ///< Transceiver block next to the GEM VTTX
+      SCA_PT100      = 0x07, ///< SCA temperature sensor
+      V6_FPGA_PT100  = 0x08, ///< Virtex6 temperature sensor
+
+      // SCA ADC signal strength sensors
+      VTRX_RSSI3     = 0x12, ///< ADC channel 18, VTRx rssi3
+      VTRX_RSSI2     = 0x13, ///< ADC channel 19, VTRx rssi2
+      VTRX_RSSI1     = 0x15, ///< ADC channel 21, VTRx rssi1
```
bdorney, line 243: if (ch == SCAADCChannel::VTTX_CSC_PT100 || ch == SCAADCChannel::VTTX_GEM_PT100 || ch == SCAADCChannel::SCA_PT100 || ch == SCAADCChannel::V6_FPGA_PT100 )

To this struct I would suggest to add an `std::unordered_map<int,bool>` which is filled by the `ADCChannel` constructor (which you would have to create) which would have:

```C++
map_connChan[VTTX_CSC_PT100] = true;
map_connChan[VTTX_GEM_PT100] = true;
...
...
```

This would greatly simplify some of your if statements below (e.g. my comment [here](https://github.com/cms-gem-daq-project/ctp7_modules/pull/120/files#r282377698)).

src/amc/sca.cpp
```diff
@@ -129,11 +129,16 @@ std::vector<uint32_t> scaADCCommand(localArgs* la, SCAADCChannelT const& ch, SCA
...
+    sendSCACommand(la, SCAChannel::ADC, SCAADCCommand::ADC_W_CURR, 0x4, 0x1<<ch, ohMask);
   // start the conversion and get the result
   std::vector<uint32_t> result = sendSCACommandWithReply(la, SCAChannel::ADC, SCAADCCommand::ADC_GO, 0x4, 0x1, ohMask);
 
+  // disable the current sink if reading the temperature ADCs
+  if (ch == SCAADCChannel::VTTX_CSC_PT100 || ch == SCAADCChannel::VTTX_GEM_PT100 || ch == SCAADCChannel::SCA_PT100 || ch == SCAADCChannel::V6_FPGA_PT100 )	// Hardcoded channel numbers
```
bdorney: same

bdorney (changes requested):
some changes requested to simplify code base and improve documentation.

include/amc/sca_enums.h
jsturdy: I specifically asked to **not** remove these `enum`s for now

include/amc/sca_enums.h
jsturdy: Where the purpose of the map is to do what?
If to tell whether the sensor uses a current source, it's probably cleaner to just add a function to the struct that returns `true` if the `enum` is a temperature sensor or not, i.e., something like:
```c++
bool useCurrentSource(SCAADCChannelT sensor)
{
  switch(sensor) {
    case(SCAADCChannel::VTTX_CSC_PT100: return true;
    default: return false
  }
}
```

src/amc/sca.cpp
jsturdy: This change *could* be made in a separate PR (the nomenclature is my doing, and this function is **not** the only one exhibiting this)

include/amc/sca.h
ram1123: @bdorney ,

Is this fine:

- `word_array data` : It contains 27 bit information. Bit 0 to 11 contains output data, 12 to 15 are zeros, 16 to 20 contains SCA-ADC ID, 21 to 23 are zeros, 24 to 26 contains the ohMask word and bit 27 contains information about a particular OH is masked (0) or not(1).

src/utils.cpp
```diff
@@ -128,6 +128,17 @@ void readRegFromDB(const RPCMsg *request, RPCMsg *response)
...
+uint32_t bitCheck(uint32_t word, int bit)
+{
+  if (bit > 31)
+    throw std::invalid_argument("Invalid request to shift word by more than 31 bits");
+  if ( (word >> bit) & 1 ) {
+    return 1;
+  } else {
+    return 0;
+  }
+}
```
ram1123, line 136: @jsturdy ,

Is it good to change the function `bitCheck` like below?

```cpp
uint32_t bitCheck(uint32_t word, int bit)
{
   if (bit > 31)
      throw std::invalid_argument("Invalid request to shift word by more than 31 bits");
   return (((word >> bit) & 1) ? 1 : 0);
}

Comment (jsturdy) [reply src/utils.cpp:136]: You just need to do (no need for the conditional):
```c++
return (word>>bit) & 0x1;
```
but otherwise, I think that's fine, maybe add "...shift 32-bit word by..." for clarity

jsturdy:
One other comment I missed from @bdorney regarding the PT100 sensors, the one we've labeled as `SCA_PT100` should be `GBT0_PT100`, according to Yifan

ram1123:
> One other comment I missed from @bdorney regarding the PT100 sensors, the one we've labeled as `SCA_PT100` should be `GBT0_PT100`, according to Yifan

done.

include/amc/sca.h
jsturdy: I think we can clean up the documentation by doing:
```c++
*  \param[out] response RPC response message with the following keys:
*    - `word_array data` : ADC data is returned as an array of 32-bit words.
                           Data is formatted as: <present 27><link ID 26:24><0s 23:21><ADC ID 20:16><0's 15:12><data 11:0>
```
or
```c++
*  \param[out] response RPC response message with the following keys:
*    - `word_array data` : ADC data is returned as an array of 32-bit words formatted as:
                           bit [27]: data present
                           bits [26:24]: link ID
                           bits [23:21]: constant 0's
                           bits [20:16]: ADC channel ID
                           bits [15:12]: constant 0's
                           bits [11:0]: ADC data
```

ram1123: For me 2nd option seems better. I will wait for @bdorney comment.

include/amc/sca_enums.h
ram1123: @bdorney & @jsturdy ,

Finally, what we are planning to do?

jsturdy: Implement my suggestion, but make the function, e.g.,
```c++
struct ADCChannel {
  enum EADChannel {
    // blah
  } ADCChannel;

  static constexpr bool useCurrentSource(EADCChannel sensor)
{
  switch(sensor) {
    case(VTTX_CSC_PT100: return true;
    default: return false
  }
}
```

Then when using, you'll do, e.g.,
```c++
if (SCAADCChannel::useCurrentSource(ch)) {
  // something
}
```

include/amc/sca.h
ram1123: Hi @bdorney ,

I am implementing the @jsturdy suggestion:

```cpp
*  \param[out] response RPC response message with the following keys:
*    - `word_array data` : ADC data is returned as an array of 32-bit words formatted as:
                           bit [27]: data present
                           bits [26:24]: link ID
                           bits [23:21]: constant 0's
                           bits [20:16]: ADC channel ID
                           bits [15:12]: constant 0's
                           bits [11:0]: ADC data
```

include/amc/sca_enums.h
ram1123: I implemented it like (as per @jsturdy suggestation):

```cpp
struct ADCChannel {
  enum EADChannel {
    // blah
  } ADCChannel;
    
static bool useCurrentSource(EADCChannel sensor)
    {
       switch (sensor) {
          case (VTTX_CSC_PT100): return true;
	  case (VTTX_GEM_PT100): return true;
	  case (GBT0_PT100):	 return true;
	  case (V6_FPGA_PT100):	 return true;
	  default:		 return false;
       }
    }
}
```

Then it is used as:

```cpp
  // enable the current sink if reading the temperature ADCs
  if (SCAADCChannel::useCurrentSource(ch))
    sendSCACommand(la, SCAChannel::ADC, SCAADCCommand::ADC_W_CURR, 0x4, 0x1<<ch, ohMask);
```

src/amc/sca.cpp
ram1123: I implemented it like (as per @jsturdy suggestation):

```cpp
struct ADCChannel {
  enum EADChannel {
    // blah
  } ADCChannel;
    
static bool useCurrentSource(EADCChannel sensor)
    {
       switch (sensor) {
          case (VTTX_CSC_PT100): return true;
	  case (VTTX_GEM_PT100): return true;
	  case (GBT0_PT100):	 return true;
	  case (V6_FPGA_PT100):	 return true;
	  default:		 return false;
       }
    }
}
```

Then it is used as:

```cpp
  // enable the current sink if reading the temperature ADCs
  if (SCAADCChannel::useCurrentSource(ch))
    sendSCACommand(la, SCAChannel::ADC, SCAADCCommand::ADC_W_CURR, 0x4, 0x1<<ch, ohMask);
```

src/amc/sca.cpp
ram1123: I implemented it like (as per @jsturdy suggestation):

```cpp
struct ADCChannel {
  enum EADChannel {
    // blah
  } ADCChannel;
    
static bool useCurrentSource(EADCChannel sensor)
    {
       switch (sensor) {
          case (VTTX_CSC_PT100): return true;
	  case (VTTX_GEM_PT100): return true;
	  case (GBT0_PT100):	 return true;
	  case (V6_FPGA_PT100):	 return true;
	  default:		 return false;
       }
    }
}
```

Then it is used as:

```cpp
  // enable the current sink if reading the temperature ADCs
  if (SCAADCChannel::useCurrentSource(ch))
    sendSCACommand(la, SCAChannel::ADC, SCAADCCommand::ADC_W_CURR, 0x4, 0x1<<ch, ohMask);
```

include/amc/sca.h
ram1123: implementing  @jsturdy's suggestion:

```cpp
*  \param[out] response RPC response message with the following keys:
*    - `word_array data` : ADC data is returned as an array of 32-bit words formatted as:
                           bit [27]: data present
                           bits [26:24]: link ID
                           bits [23:21]: constant 0's
                           bits [20:16]: ADC channel ID
                           bits [15:12]: constant 0's
                           bits [11:0]: ADC data
```

include/amc/sca.h
ram1123: implementing  @jsturdy's suggestion:

```cpp
*  \param[out] response RPC response message with the following keys:
*    - `word_array data` : ADC data is returned as an array of 32-bit words formatted as:
                           bit [27]: data present
                           bits [26:24]: link ID
                           bits [23:21]: constant 0's
                           bits [20:16]: ADC channel ID
                           bits [15:12]: constant 0's
                           bits [11:0]: ADC data
```

include/amc/sca.h
ram1123: implementing  @jsturdy's suggestion:

```cpp
*  \param[out] response RPC response message with the following keys:
*    - `word_array data` : ADC data is returned as an array of 32-bit words formatted as:
                           bit [27]: data present
                           bits [26:24]: link ID
                           bits [23:21]: constant 0's
                           bits [20:16]: ADC channel ID
                           bits [15:12]: constant 0's
                           bits [11:0]: ADC data
```

include/amc/sca.h
bdorney: Yes sorry for late reply; the proposed change is good.  It's much clearer now.  Thanks to you both.

jsturdy (dismissed):
I think this is OK on my side now.
@bdorney, @mexanick, @evka85?

include/amc/sca_enums.h
```diff
@@ -231,39 +231,48 @@ class SCASettings {
...
+      //ADC_CH28 = 0x1C, ///< ADC channel 28
+      //ADC_CH29 = 0x1D, ///< ADC channel 29
     } ADCChannel;
+
+    //static constexpr bool useCurrentSource(EADCChannel sensor)
+    static bool useCurrentSource(EADCChannel sensor)
+    {
+       switch (sensor) {
+          case (VTTX_CSC_PT100): return true;
+	  case (VTTX_GEM_PT100): return true;
```
bdorney: looks like you have an extra space on line 269, e.g. compare to line 270.

src/amc/sca.cpp
```diff
@@ -235,3 +240,145 @@ void resetSCASEUCounter(const RPCMsg *request, RPCMsg *response)
...
+  std::vector<uint32_t> result;
+  std::vector<uint32_t> outData;
+
+  result = scaADCCommand(&la, ch, ohMask);
+
+  int ohIdx = 0;
+  for(auto const& val : result) {
+      LOGGER->log_message(LogManager::DEBUG, stdsprintf("Value for OH%i, SCA-ADC channel 0x%x = %i ",ohIdx, ch, val)); 
+      outData.push_back((bitCheck(ohMask, ohIdx) << 27) | (ohIdx<<24) | (ch<<16) | val);
+      response->set_word_array("data",outData.data(), outData.size());
```
bdorney: Seems like you're using the `RPCMsg::set_word_array` method that's for arrays, but this is an overloaded method that also works for `std::vector` objects,  why not just use that one, e.g. `RPCMsg:: set_word_array(std::string key, const std::vector<uint32_t> &data)`?  So line 260 would become:

```C++
    response->set_word_array("data", outData);
```

Also would prefer the RPC key name was a bit less generic.  e.g. instead of `"data"` it might be more instructive to see `"scaADCData"`.  Although not sure what @jsturdy's policy on this would be moving forward.

src/amc/sca.cpp
```diff
@@ -235,3 +240,145 @@ void resetSCASEUCounter(const RPCMsg *request, RPCMsg *response)
 
   rtxn.abort();
 }
+
+void readSCAADCSensor(const RPCMsg *request, RPCMsg *response)
+{
+  GETLOCALARGS(response);
+
+  uint32_t ohMask = request->get_word("ohMask");
```
bdorney: This is provided as an input key in may places.  It may not be appropriate for this PR, but you could imagine in `amc.h`/`amc.cpp` the following function existed:

```C++
uint32_t getOHMask(localArgs *la){
   uint32 scaReady = readReg(la, "GEM_AMC.SLOW_CONTROL.SCA.STATUS.READY");
   uint32 scaError = readReg(la, "GEM_AMC.SLOW_CONTROL.SCA.STATUS.CRITICAL_ERROR");

   return (scaReady & (~scaError & 0xfff));
}
```

The function above could be adapted to have some error/warning inserted into the RPC response (although I think this is not the road @jsturdy advocated at during the meeting today), or throw a runtime error in which the calling code would have to catch, in the event `0x0` (e.g. no unmasked OH's) is determined to be the `ohMask`. 

The following could then be done everywhere:

```C++
uint32_t ohMask = 0x0;
if (request.get_key_exists("ohMask"))
{
   ohMask = request->get_word("ohMask");
} else {
   ohMask = getOHMask(la);
}
```

If it's not desired to implement this `getOHMask` function in this PR then at minimum we should do the following instead of line 248:

```C++
uint32_t ohMask = 0xfff;
{
   ohMask = request->get_word("ohMask");
}
```

e.g. default behavior if no key is supplied would be to consider all OH's.

src/amc/sca.cpp
```diff
@@ -235,3 +240,145 @@ void resetSCASEUCounter(const RPCMsg *request, RPCMsg *response)
...
+  }
+
+  rtxn.abort();
+}
+
+void readSCAADCTemperatureSensors(const RPCMsg *request, RPCMsg *response)
+{
+  GETLOCALARGS(response);
+
+  uint32_t ohMask = request->get_word("ohMask");
```
bdorney: Assume my comment [here](https://github.com/cms-gem-daq-project/ctp7_modules/pull/120/files#r283759749) applies to all instances of this line.

src/amc/sca.cpp
```diff
@@ -235,3 +240,145 @@ void resetSCASEUCounter(const RPCMsg *request, RPCMsg *response)
...
+    result = scaADCCommand(&la, channelName, ohMask);
+    ohIdx = 0;
+    for(auto const& val : result) {
+    	LOGGER->log_message(LogManager::DEBUG, stdsprintf("Temperature for OH%i, SCA-ADC channel 0x%x = %i ",ohIdx, channelName, val));
+	outData.push_back((bitCheck(ohMask, ohIdx) << 27) | (ohIdx<<24) | (channelName<<16) | val);
+	++ohIdx;
+    }
+  }
+  
+  response->set_word_array("data", outData.data(), outData.size());
```
bdorney: Assume my comment [here](https://github.com/cms-gem-daq-project/ctp7_modules/pull/120/files#r283752308) applies to all instances of your usage of the array version of `wisc::RPCMsg::set_word_array` over the `std::vector` version.

src/amc/sca.cpp
```diff
@@ -235,3 +240,145 @@ void resetSCASEUCounter(const RPCMsg *request, RPCMsg *response)
...
+  SCAADCChannelT  ch = static_cast<SCAADCChannelT>(request->get_word("ch"));
+  
+  std::vector<uint32_t> result;
+  std::vector<uint32_t> outData;
+
+  result = scaADCCommand(&la, ch, ohMask);
+
+  int ohIdx = 0;
+  for(auto const& val : result) {
+      LOGGER->log_message(LogManager::DEBUG, stdsprintf("Value for OH%i, SCA-ADC channel 0x%x = %i ",ohIdx, ch, val));
```
bdorney, line 259: Unless there is a specific reason from @jsturdy would prefer this log message show SCA ADC channel number in decimal rather than hex.  Comment applies to all instances.

src/amc/sca.cpp
```diff
@@ -235,3 +240,145 @@ void resetSCASEUCounter(const RPCMsg *request, RPCMsg *response)
...
+void readAllSCAADCSensors(const RPCMsg *request, RPCMsg *response)
+{
+  GETLOCALARGS(response);
+
+  uint32_t ohMask = request->get_word("ohMask");
+
+  int ohIdx;
+  std::vector<uint32_t> result;
+  std::vector<uint32_t> outData;
+  for (SCAADCChannelT channelName = SCAADCChannel::VTTX_CSC_PT100; channelName<=SCAADCChannel::SCA_TEMP; channelName=SCAADCChannelT(channelName+1)) {
```
bdorney, line 370: I think you missed a spot where the outer for loop should be like `for(auto const& channelName : chArr) {` with appropriate definition of `chArr`.  Otherwise this for loop may behave unexpectedly if entries are added to the `SCAADCChannel` `enum`.

bdorney (changes requested):
Code cleanup needed in some places.

src/amc/sca.cpp
ram1123: @bdorney : Yes, I will change the  response like `response->set_word_array("data", outData);`. For `"data"` to `"scaADCData"` I will wait what @jsturdy  says.

include/amc/sca_enums.h
ram1123: I don't know why on github its showing like this. While in my vi editor its fine.

bdorney: Do you have your editor configured to convert the tab character `\t` to a specific number of spaces?  If not then the reason is most likely this.

ram1123: yes... `\t` is shifts to 3 spaces. When I enter it keeps indentation. So, it might be because of this.

src/amc/sca.cpp
jsturdy: For defaults for non-mandatory keys, I prefer something, a la
```c++
    const uint32_t ohMask   = request->get_key_exists("ohMask") ? request->get_word("ohMask") : 0xfff; // or better yet, some defined constant, e.g., FULL_OH_MASK
```

src/amc/sca.cpp
jsturdy: The loop is "ok" in this case, but it **will** have *unintended* consequences (that I think I mentioned in an earlier comment), in that it will attempt to sample *every* (even not present) channels.
Sure, one *could* duplicate the `enum` class here just to get an iterable object (as was done in the other functions), but that should really be done in a better way inside the `enum` (class) itself, and also removal of the arrays defined in the other functions, to get iterators for all, temperature, voltage, and rssi channels, independently.
For now, I don't consider this a blocker

src/amc/sca.cpp
jsturdy: I strongly prefer completely generic `"data"` keys for simplicity (here and everywhere else); the calling function knows which function it's calling, so it doesn't need to give a special name to the data...

include/amc/sca_enums.h
jsturdy: You should remove *completely* any `\t` characters because of issues like this, what @bdorney was asking is if you have your editor configured to do this, which it appears you don't.
I'm not a `vi/vim` user, so I can't advise here, but you can probably search for "how to automatically convert all whitespace to space characters"

src/amc/sca.cpp
bdorney: I see, okay.  I will try to remember this moving forward.  @ram1123 please address just the `set_word_array` comment above.

src/amc/sca.cpp
bdorney: One could add `constexpr uint32_t FULL_OH_MASK = 0xfff` inside of:

https://github.com/cms-gem-daq-project/ctp7_modules/blob/b358ed18fd77ed0625b5f04082df7d19da2d7ff6/include/hw_constants.h#L18-L26

And then `amc::FULL_OH_MASK` could be used here.

bdorney: I still favor automatic determination of OH Mask though.

include/amc/sca_enums.h
bdorney: For `vim` you should add:

```vim
" show existing tab with 4 spaces width
set tabstop=4
```

To your `$HOME/.vimrc` to ensure a tab character is represented always as four space characters `\s`.  For `vi` I don't know; a google search would be needed.

ram1123: Thanks for pointing and suggestations. I will take care of this before making commit.

src/amc/sca.cpp
ram1123: Added like :

```cpp
  const uint32_t ohMask   = request->get_key_exists("ohMask") ? request->get_word("ohMask") : amc::FULL_OH_MASK;
  SCAADCChannelT  ch = static_cast<SCAADCChannelT>(request->get_key_exists("ch") ? request->get_word("ch") : 0x00);
```

src/amc/sca.cpp
ram1123: @jsturdy : Should I shift to hex?

include/amc/sca_enums.h
jsturdy: Don't think the suggestion is having the intended effect, see, e.g., [this](https://stackoverflow.com/questions/1878974/redefine-tab-as-4-spaces)

src/amc/sca.cpp
jsturdy: You probably do not want to allow a default channel selection

src/amc/sca.cpp
jsturdy: no comment (within "reasonable" time) means no preference (or on holiday, but that's not currently the case ;-) )

src/amc/sca.cpp
ram1123: @jsturdy : Should I throw an exception that no key exists?

ram1123: Changed it like:

https://github.com/cms-gem-daq-project/ctp7_modules/blob/09d87c4195e039efa16615b5169d47c19c018397/src/amc/sca.cpp#L249-L250

jsturdy: You can't (shouldn't) `throw` from the RPC methods, there's nothing to catch it there.
Just do it like you had it before you addressed the comment about `ohMask`; the initial comment applied **only** to `ohMask`

jsturdy (approved):

jsturdy:
@mexanick, any objections to merging this?

mexanick (approved):
Cursory

ram1123:
## Output values from the sensors

| Sensor ID | Info | ADC Count | Measured decimal Value| Expected value ([link](https://github.com/cms-gem-daq-project/ctp7_modules/blob/b358ed18fd77ed0625b5f04082df7d19da2d7ff6/include/amc/sca_enums.h#L211-L267))
| -------- | -------- | -------- |--- |---|
| 0x00    |    PT100    |    44    |    19 $^0C$ | -
| 0x04    |    PT100    |    44    |    19 $^0C$ |-
| 0x07    |    PT100    |    45    |    24.40 $^0C$ | -
| 0x08    |    PT100    |    44    |    19 $^0C$ | -
| 0x1F    |    SCA internal temperature | 2743    | 25 $^0C$ | -
| 0x1B    |    FPGA MGT Voltage | 1025 | 1 V | 1 V
| 0x1E    |    FPGA MGT Voltage | 1254 | 1.22 V | 1.2 V
| 0x11    |    FPGA core voltage | 1066| 1.04 V| 1.0 V
| 0x18    |    Power for GBTx and SCA | 1524 | 1.5 V| 1.5 V
| 0x0F    |    FPGA I/O power | 2620 | 2.6 V | 2.5 V
| 0x15    |    Signal strength of VTRX1 | 14 | 13 mV | -
| 0x13    |    Signal strength of VTRX1 | 14 | 13 mV |-
| 0x12    |    Signal strength of VTRX1 | 12 | 12 mV | -

## How the conversions are made

### ADC counts to voltage conversion

The output values are 12 bit ADC and the voltage value varies from 0 to 1 V. Thus, each bit corresponds to 1/2^{12} =~ 0.244 mV.

### PT100 Sensors 

- ADC count = 44
- Corresponding voltage = 44 x 0.244 mV = 10.736 mV
- We are measuring temperature using the platenium resistor from which 100 $\mu$A current is flowing
- Resistance = V/I = 10.736mV / 100 $\mu$ A = 107.36  $\Omega$
- Check the temperature corresponding to the 107.36 $\Omega$ resistance in the R - T table [1]. This corresponds to ~19 $^0 C$.


### SCA Internal temperature

- ADC count = 2743
- Corresponding voltage = 2743 X 0.244 mV = 669.68 mV
- Now look at the V-T graph. The voltage of 669.68 corresponds to ~25 $^0 C$.

upload_70214b9a453866df6201dd01a97f9573

The above V-T graph is taken from page 52 of GBT-SCA manual [2]



### Conversion of ADC for Voltage sensors

- ADC count = 1025
- As we can see from the image below there is a voltage divider having two resistor having values 3k and 1k each. So, need to multiply this by 4.

upload_65554560d042b44d9ed7369b6cea3992

- Voltage  = 1025 * 0.244 * 4 mV ~= 1V

---

## Reference

1. [https://www.intech.co.nz/products/temperature/typert/RTD-Pt100-Conversion.pdf](https://www.intech.co.nz/products/temperature/typert/RTD-Pt100-Conversion.pdf)

2. [https://espace.cern.ch/GBT-Project/GBT-SCA/Manuals/GBT-SCA-UserManual.pdf](https://espace.cern.ch/GBT-Project/GBT-SCA/Manuals/GBT-SCA-UserManual.pdf)

Issue #131, opened 2019-06-13
GE2/1 compatibility and elimination of GE1/1-specific hardcoded constants


mexanick:
## Brief summary of issue

Current version of the code suffers from various GEM detector type specific constants being hardcoded and can't be used in a straightforward manner with the GE2/1 
### Types of issue

- [ ] Bug report (report an issue with the code)
- [x] Feature request (request for change which adds functionality)

## Expected Behavior

The software should work in the very same manner between GE1/1 and GE2/1
## Current Behavior

Currently some of the methods are still working, but wasting the memory assuming GE1/1 electronics layout, others simply does not work properly due to mapping differences.

## Your Environment

* Version used: tip of develop
* Shell used: bash
* Setup: GE2/1 integration stand

PR #132, opened 2019-06-13
GE2/1 compatilibility

mexanick:
Addressing #131 
## Description

* Adding hw_constants for ge21
* Implementing constants usage
* Minor improvements w.r.t warnings and gbt logging

The changes should be transparent and does not require changes in the other software for the GE1/1 usage

### Types of changes

## Motivation and Context

The hardware constants has to be defined in one place and corresponding compile-time switch has to be used
## How Has This Been Tested?

Tested with the GE2/1 integration stand setup
## Checklist:

### DIFF
--- a/include/calibration_routines.h
+++ b/include/calibration_routines.h
@@ -54,7 +54,7 @@ struct vfat3DACAndSize{
     }
 };
 
-/*! \fn std::unordered_map<uint32_t, uint32_t> setSingleChanMask(int ohN, int vfatN, unsigned int ch, localArgs *la)
+/*! \fn std::unordered_map<uint32_t, uint32_t> setSingleChanMask(unsigned int ohN, int vfatN, unsigned int ch, localArgs *la)
  *  \brief Unmask the channel of interest and masks all the other
  *  \param ohN Optical link number
  *  \param vfatN VFAT position
@@ -62,7 +62,7 @@ struct vfat3DACAndSize{
  *  \param la Local arguments structure
  *  \return Original channel mask in a form of an unordered map <chanMaskAddr, mask>
  */
-std::unordered_map<uint32_t, uint32_t> setSingleChanMask(int ohN, int vfatN, unsigned int ch, localArgs *la);
+std::unordered_map<uint32_t, uint32_t> setSingleChanMask(unsigned int ohN, int vfatN, unsigned int ch, localArgs *la);
 
 /*! \fn void applyChanMask(std::unordered_map<uint32_t, uint32_t> map_chanOrigMask, localArgs *la)
  *  \brief Applies channel mask
--- a/include/hw_constants.h
+++ b/include/hw_constants.h
@@ -22,6 +22,12 @@ namespace amc {
         constexpr uint32_t OH_PER_AMC = 12;    ///< The number of OptoHybrids per AMC.
     }
 
+    /*! \brief GE2/1 specific namespace.
+     */
+    namespace ge21 {
+        constexpr uint32_t OH_PER_AMC = 12;    ///< The number of OptoHybrids per AMC.
+    }
+
     using namespace GEM_VARIANT;
 }
 
@@ -35,6 +41,13 @@ namespace oh {
         constexpr size_t OH_SINGLE_RAM_SIZE = 2*100; ///< Per-OH RAM size: for GE1/1 100 32-bit words of configuration per OH plus the corresponding OH local address
     }
 
+    /*! \brief GE2/1 specific namespace.
+     */
+    namespace ge21 {
+        constexpr uint32_t VFATS_PER_OH     = 12;    ///< The number of VFAT's per OptoHybrid.
+        constexpr size_t OH_SINGLE_RAM_SIZE = 2*100; ///< Per-OH RAM size: for GE1/1 100 32-bit words of configuration per OH plus the corresponding OH local address
+    }
+
     using namespace GEM_VARIANT;
 }
 
@@ -44,6 +57,12 @@ namespace vfat {
     /*! \brief GE1/1 specific namespace.
      */
     namespace ge11 {
+        constexpr size_t VFAT_SINGLE_RAM_SIZE = 74; ///< Per-VFAT RAM size: for GE1/1 24 VFATs per OH, 74 32-bit words of configuration per VFAT (147 16-bit VFAT configurations -> 32-bit words + padding)
+    }
+
+    /*! \brief GE2/1 specific namespace.
+     */
+    namespace ge21 {
         constexpr size_t VFAT_SINGLE_RAM_SIZE = 74; ///< Per-VFAT RAM size: for GE1/1 24 VFATs per OH, 74 32-bit words of configuration per VFAT (147 16-bit VFAT configurations -> 32-bit words + padding)
     }
 
@@ -68,11 +87,21 @@ namespace gbt {
         constexpr size_t GBT_SINGLE_RAM_SIZE = 92; ///< Per-GBT RAM size: for GE1/1 3 GBTx per OH, 92 32-bit words of configuration per GBT (366 8-bit GBT configurations -> 32-bit words + padding)
     }
 
+    /*! \brief GE2/1 specific namespace.
+     */
+    namespace ge21{
+        constexpr uint32_t GBTS_PER_OH       = 2;  ///< The number of GBT's per OptoHybrid (should be part of the oh namespace, as VFATS_PER_OH)
+        constexpr size_t GBT_SINGLE_RAM_SIZE = 92; ///< Per-GBT RAM size: for GE1/1 3 GBTx per OH, 92 32-bit words of configuration per GBT (366 8-bit GBT configurations -> 32-bit words + padding)
+    }
+
     using namespace GEM_VARIANT;
 
     /*! \brief Mappings between elinks, GBT index and VFAT index.
      */
     namespace elinkMappings {
+
+        /*! \brief GE1/1 specific namespace.
+         */
         namespace ge11 {
             /*! \brief Mapping from VFAT index to GBT index.
              */
@@ -84,6 +113,28 @@ namespace gbt {
              */
             constexpr std::array<uint8_t, 24> VFAT_TO_ELINK {
                 { 5, 9, 2, 3, 1, 8, 6, 6, 4, 1, 5, 4, 3, 2, 1, 0, 7, 8, 6, 7, 2, 3, 9, 8 }
+            };
+
+            /*! \brief Mapping from elink index to the 3 registers addresses in the GBT.
+	         */
+            constexpr std::array<std::array<uint16_t, 3>, 10> ELINK_TO_REGISTERS {
+                { {69, 73, 77}, {67, 71, 75}, {93, 97, 101}, {91, 95, 99}, {117, 121, 125}, {115, 119, 123}, {141, 145, 149}, {139, 143, 147}, {165, 169, 173}, {163, 167, 171} }
+            };
+        }
+
+        /*! \brief GE2/1 specific namespace.
+         */
+        namespace ge21 {
+            /*! \brief Mapping from VFAT index to GBT index.
+             */
+            constexpr std::array<uint32_t, 12> VFAT_TO_GBT {
+                { 0, 0, 0, 0, 0, 0, 1, 1, 1, 1, 1, 1 }
+            };
+
+            /*! \brief Mapping from VFAT index to the elink of its corresponding GBT.
+             */
+            constexpr std::array<uint8_t, 12> VFAT_TO_ELINK {
+                { 0, 1, 2, 3, 4, 5, 0, 1, 2, 3, 4, 5 }
             };
 
             /*! \brief Mapping from elink index to the 3 registers addresses in the GBT.
--- a/include/hw_constants.h
+++ b/include/hw_constants.h
@@ -39,13 +39,17 @@ namespace oh {
     namespace ge11 {
         constexpr uint32_t VFATS_PER_OH     = 24;    ///< The number of VFAT's per OptoHybrid.
         constexpr size_t OH_SINGLE_RAM_SIZE = 2*100; ///< Per-OH RAM size: for GE1/1 100 32-bit words of configuration per OH plus the corresponding OH local address
+        constexpr uint32_t CHANNELS_PER_OH  = 3072;  ///< The number of VFAT channels per OptoHybrid
+        constexpr uint32_t SBITS_PER_OH     = 1536;  ///< The number of S-bits per OptoHybrid
     }
 
     /*! \brief GE2/1 specific namespace.
      */
     namespace ge21 {
         constexpr uint32_t VFATS_PER_OH     = 12;    ///< The number of VFAT's per OptoHybrid.
-        constexpr size_t OH_SINGLE_RAM_SIZE = 2*100; ///< Per-OH RAM size: for GE1/1 100 32-bit words of configuration per OH plus the corresponding OH local address
+        constexpr size_t OH_SINGLE_RAM_SIZE = 2*100; ///< Per-OH RAM size: for GE2/1 100 32-bit words of configuration per OH plus the corresponding OH local address
+        constexpr uint32_t CHANNELS_PER_OH  = 1536;  ///< The number of VFAT channels per OptoHybrid
+        constexpr uint32_t SBITS_PER_OH     = 768;   ///< The number of S-bits per OptoHybrid
     }
 
     using namespace GEM_VARIANT;
@@ -63,7 +67,7 @@ namespace vfat {
     /*! \brief GE2/1 specific namespace.
      */
     namespace ge21 {
-        constexpr size_t VFAT_SINGLE_RAM_SIZE = 74; ///< Per-VFAT RAM size: for GE1/1 24 VFATs per OH, 74 32-bit words of configuration per VFAT (147 16-bit VFAT configurations -> 32-bit words + padding)
+        constexpr size_t VFAT_SINGLE_RAM_SIZE = 74; ///< Per-VFAT RAM size: for GE2/1 12 VFATs per OH, 74 32-bit words of configuration per VFAT (147 16-bit VFAT configurations -> 32-bit words + padding)
     }
 
     using namespace GEM_VARIANT;
@@ -91,7 +95,7 @@ namespace gbt {
      */
     namespace ge21{
         constexpr uint32_t GBTS_PER_OH       = 2;  ///< The number of GBT's per OptoHybrid (should be part of the oh namespace, as VFATS_PER_OH)
-        constexpr size_t GBT_SINGLE_RAM_SIZE = 92; ///< Per-GBT RAM size: for GE1/1 3 GBTx per OH, 92 32-bit words of configuration per GBT (366 8-bit GBT configurations -> 32-bit words + padding)
+        constexpr size_t GBT_SINGLE_RAM_SIZE = 92; ///< Per-GBT RAM size: for GE2/1 2 GBTx per OH, 92 32-bit words of configuration per GBT (366 8-bit GBT configurations -> 32-bit words + padding)
     }
 
     using namespace GEM_VARIANT;
--- a/src/amc.cpp
+++ b/src/amc.cpp
@@ -45,12 +45,12 @@ unsigned int fw_version_check(const char* caller_name, localArgs *la)
 
 uint32_t getOHVFATMaskLocal(localArgs * la, uint32_t ohN)
 {
-    uint32_t mask = 0x0;
+    uint32_t mask = 0xffffff; //Start with all vfats masked for max VFAT/GEB amount
     for (unsigned int vfatN=0; vfatN<oh::VFATS_PER_OH; ++vfatN) { //Loop over all vfats
         uint32_t syncErrCnt = readReg(la, stdsprintf("GEM_AMC.OH_LINKS.OH%i.VFAT%i.SYNC_ERR_CNT",ohN,vfatN));
 
-        if (syncErrCnt > 0x0) { //Case: nonzero sync errors, mask this vfat
-            mask = mask + (0x1 << vfatN);
+        if (syncErrCnt == 0x0) { //Case: zero sync errors, unmask this vfat
+            mask = mask - (0x1 << vfatN);
         } //End Case: nonzero sync errors, mask this vfat
     } //End loop over all vfats
 
--- a/src/amc.cpp
+++ b/src/amc.cpp
@@ -9,6 +9,7 @@
 #include "amc/ttc.h"
 #include "amc/daq.h"
 #include "amc/blaster_ram.h"
+#include "hw_constants.h"
 
 #include <chrono>
 #include <string>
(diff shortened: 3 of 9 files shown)


jsturdy:
Is this targeting "working now with setups that may be using legacy", if so, needs to target `release/1.1.X`, which is what I understand based on the dependent PRs

include/hw_constants.h
```diff
@@ -93,6 +122,28 @@ namespace gbt {
...
+
+            /*! \brief Mapping from VFAT index to the elink of its corresponding GBT.
+             */
+            constexpr std::array<uint8_t, 12> VFAT_TO_ELINK {
+                { 0, 1, 2, 3, 4, 5, 0, 1, 2, 3, 4, 5 }
+            };
+
+            /*! \brief Mapping from elink index to the 3 registers addresses in the GBT.
+	         */
+            constexpr std::array<std::array<uint16_t, 3>, 10> ELINK_TO_REGISTERS {
```
bdorney, line 146: Unrelated to this PR; but now would be a good time to add a GE2/1 heading to the guide similar to this one:

https://github.com/cms-gem-daq-project/sw_utils/blob/develop/v3ElectronicsUserGuide.md#e-link-assignment-in-ge11

include/utils.h
```diff
@@ -56,6 +58,53 @@ static constexpr uint32_t LMDB_SIZE = 1UL * 1024UL * 1024UL * 50UL; ///< Maximum
...
+    slowCtrlErrCntVFAT operator + (const slowCtrlErrCntVFAT &vfatErrs) const
+    {
+        return slowCtrlErrCntVFAT(
+                crc+vfatErrs.crc,
+                packet+vfatErrs.packet,
+                bitstuffing+vfatErrs.bitstuffing,
+                timeout+vfatErrs.timeout,
+                axi_strobe+vfatErrs.axi_strobe,
+                sum+vfatErrs.sum,
+                nTransactions+vfatErrs.nTransactions);
```
bdorney: this is in the legacy branch already but one comment here is that I overlooked the possibility that overflow could occur in this addition here.  That is not presently handled and it might be better to change to:

```C++
        return slowCtrlErrCntVFAT(
                overflowTest(crc,vfatErrs.crc),
                overflowTest(packet,vfatErrs.packet),
                ...
                ... 
                overflowTest(nTransactions,vfatErrs.nTransactions));
```

src/calibration_routines.cpp
```diff
@@ -641,12 +642,12 @@ void sbitRateScanParallelLocal(localArgs *la, uint32_t *outDataDacVal, uint32_t
             } //End loop over optohybrids
 
             //Get the SBIT Rate Monitor Address
-            uint32_t ohTrigRateAddr[12][25]; //idx 0->23 VFAT counters; idx 24 overall rate
-            for (int ohN = 0; ohN < 12; ++ohN) {
+            uint32_t ohTrigRateAddr[amc::OH_PER_AMC][oh::VFATS_PER_OH + 1]; //idx 0->23 VFAT counters; idx 24 overall rate
+            for (unsigned int ohN = 0; ohN < amc::OH_PER_AMC; ++ohN) {
```
bdorney, line 650: Some caution should be used here when designing a loop over the number of OH's; and it's been previously overlooked (now is probably a good time to address this).  Consider the case where you're running on a 4OH FW version; or a GLIB with a quad/octal SFP FMC added, the number of OH's is not 12.  Attempting to loop over `amc::OH_PER_AMC` may actually cause an error.  For example consider lines 651 & 652 below:

```C++
                        sprintf(regBuf,"GEM_AMC.OH.OH%i.FPGA.TRIG.CNT.VFAT%i_SBITS",ohN,vfat);
                        ohTrigRateAddr[ohN][vfat] = getAddress(la, regBuf);	                        ohTrigRateAddr[ohN][vfat] = getAddress(la, regBuf);
```

The `getAddress` call here would generate an error.

When looping over the number of OH's for this reason we may want to only loop over the number of OH's in the actual FW.  This of course requires a reg read to get this value; but perhaps it's also retrievable from the xml...?

bdorney: Alternatively `amc::OH_PER_AMC` could be a constant set at compile time; and be platform (e.g. uTCA shelf/test stand) dependent.

src/calibration_routines.cpp
```diff
@@ -641,12 +642,12 @@ void sbitRateScanParallelLocal(localArgs *la, uint32_t *outDataDacVal, uint32_t
             } //End loop over optohybrids
 
             //Get the SBIT Rate Monitor Address
-            uint32_t ohTrigRateAddr[12][25]; //idx 0->23 VFAT counters; idx 24 overall rate
-            for (int ohN = 0; ohN < 12; ++ohN) {
+            uint32_t ohTrigRateAddr[amc::OH_PER_AMC][oh::VFATS_PER_OH + 1]; //idx 0->23 VFAT counters; idx 24 overall rate
```
bdorney: Trailing comment here needs to be updated.

src/calibration_routines.cpp
bdorney: My first comment above should be considered a "global comment" applied everywhere.

src/calibration_routines.cpp
```diff
@@ -630,7 +631,7 @@ void sbitRateScanParallelLocal(localArgs *la, uint32_t *outDataDacVal, uint32_t
                     {
                         //Then mask all other channels except for channel ch
                         uint32_t notmask = ~vfatmask[ohN] & 0xFFFFFF;
```
bdorney, line 637: Correct me if I'm wrong but doesn't this *also* need an update, e.g.:

```C++
uint32_t notmask = ~vfatmask[ohN] & 0xFFFFFF
```

The `0xFFFFFF` is now specific to `ge11` and `ge21` so this would need a constant no?

src/calibration_routines.cpp
```diff
@@ -670,10 +671,10 @@ void sbitRateScanParallelLocal(localArgs *la, uint32_t *outDataDacVal, uint32_t
                 LOGGER->log_message(LogManager::INFO, stdsprintf("Setting %s to %i for all optohybrids in 0x%x",scanReg.c_str(),dacVal,ohMask));
 
                 //Set the scan register value
-                for (int ohN = 0; ohN < 12; ++ohN) {
+                for (unsigned int ohN = 0; ohN < amc::OH_PER_AMC; ++ohN) {
                     if ((ohMask >> ohN) & 0x1) {
                         uint32_t notmask = ~vfatmask[ohN] & 0xFFFFFF;
```
bdorney, line 680: Same comment as [here](https://github.com/cms-gem-daq-project/ctp7_modules/pull/132/files#r293509482) basically apply this everywhere.

src/calibration_routines.cpp
```diff
@@ -793,7 +794,7 @@ void checkSbitMappingWithCalPulseLocal(localArgs *la, uint32_t *outData, uint32_
     uint32_t chanRegData_orig[3072]; //original channel register data
     uint32_t chanRegData_tmp[3072]; //temporary channel register data
     getChannelRegistersVFAT3Local(la, ohN, mask, chanRegData_orig);
-    for (int idx=0; idx < 3072; ++idx) {
+    for (unsigned int idx=0; idx < 3072; ++idx) {
```
bdorney: the loop over the channel count is now also GE1/1 and GE2/1 dependent.

e.g. you need:

`ge11::CHANNELS_PER_OH`
`ge21::CHANNELS_PER_OH`

This could also have read "per detector" but since we are using the per amc/oh terminology the above is probably better.

bdorney: Not sure where else 3072 comes up but it might be in other locations that aren't illustrated here because no changes were made in the vicinity, e.g. lines 794 and 795 also use `3072`.

src/calibration_routines.cpp
```diff
@@ -856,15 +857,15 @@ void checkSbitMappingWithCalPulseLocal(localArgs *la, uint32_t *outData, uint32_
...
 
                 //bits [10:0] is the address of the cluster
                 //bits [14:12] is the cluster size
                 //bits 15 and 11 are not used
                 uint32_t thisCluster = readRawAddress(addrSbitCluster[cluster], la->response);
-                int clusterSize = (thisCluster >> 12) & 0x7;
+                unsigned int clusterSize = (thisCluster >> 12) & 0x7;
                 uint32_t sbitAddress = (thisCluster & 0x7ff);
                 bool isValid = (sbitAddress < 1536); //Possible values are [0,(24*64)-1]
                 int vfatObserved = 7-int(sbitAddress/192)+int((sbitAddress%192)/64)*8;
```
bdorney, line 875: Is the sbit clustering and addressing system in GE1/1 the same as GE2/1? e.g. will these formulas to unpack the cluster word work for GE2/1?

bdorney: For sure there are no longer 1536, see line 869, possible cluster addresses; so this would *also* need to be protected as a constant, e.g.

```C++
ge11::MAX_SBIT_ADDR_PER_OH
ge21::MAX_SBIT_ADDR_PER_OH
```

src/calibration_routines.cpp
```diff
@@ -962,7 +963,7 @@ void checkSbitRateWithCalPulseLocal(localArgs *la, uint32_t *outDataCTP7Rate, ui
     uint32_t chanRegData_tmp[3072]; //temporary channel register data
     LOGGER->log_message(LogManager::INFO, stdsprintf("Masking all channels and disabling calpulse for vfats on ohN %i", ohN));
     getChannelRegistersVFAT3Local(la, ohN, mask, chanRegData_orig);
-    for (int idx=0; idx < 3072; ++idx) {
+    for (unsigned int idx=0; idx < 3072; ++idx) {
```
bdorney: channel count number.

src/calibration_routines.cpp
```diff
@@ -980,12 +981,12 @@ void checkSbitRateWithCalPulseLocal(localArgs *la, uint32_t *outDataCTP7Rate, ui
     uint32_t addrTtcStart = getAddress(la, "GEM_AMC.TTC.GENERATOR.CYCLIC_START");
 
     //Get Trigger addresses
-    uint32_t ohTrigRateAddr[26]; //idx 0->23 VFAT counters; idx 24 rate measured by OH FPGA; idx 25 rate measured by CTP7
-    for (int vfat=0; vfat<24; ++vfat) {
+    uint32_t ohTrigRateAddr[oh::VFATS_PER_OH + 2]; //idx 0->23 VFAT counters; idx 24 rate measured by OH FPGA; idx 25 rate measured by CTP7
```
bdorney: trailing comment needs update.

bdorney (changes requested):
some additional differences between GE1/1 and GE2/1 have been overlooked (e.g. sbit clustering); need additional `gemType` specific constants.  Target branch unclear.

bdorney:
It’s possible I missed other blocks in the code due to not appearing here. Especially in the calibration toolkit.

src/calibration_routines.cpp
```diff
@@ -7,8 +7,9 @@
 #include "optohybrid.h"
 #include <thread>
 #include "vfat3.h"
+#include "hw_constants.h"
 
-std::unordered_map<uint32_t, uint32_t> setSingleChanMask(int ohN, int vfatN, unsigned int ch, localArgs *la)
+std::unordered_map<uint32_t, uint32_t> setSingleChanMask(unsigned int ohN, unsigned int vfatN, unsigned int ch, localArgs *la)
```
lpetre-ulb, line 12: Function signature is not updated in the header file.

include/hw_constants.h
```diff
@@ -35,6 +41,13 @@ namespace oh {
         constexpr size_t OH_SINGLE_RAM_SIZE = 2*100; ///< Per-OH RAM size: for GE1/1 100 32-bit words of configuration per OH plus the corresponding OH local address
     }
 
+    /*! \brief GE2/1 specific namespace.
+     */
+    namespace ge21 {
+        constexpr uint32_t VFATS_PER_OH     = 12;    ///< The number of VFAT's per OptoHybrid.
+        constexpr size_t OH_SINGLE_RAM_SIZE = 2*100; ///< Per-OH RAM size: for GE1/1 100 32-bit words of configuration per OH plus the corresponding OH local address
```
lpetre-ulb: Trailing comment should be updates. In the following constants as well.

lpetre-ulb (changes requested):
Very few comments in addition to @bdorney's that I fully agree with.

src/calibration_routines.cpp
mexanick: I'd consider this as a separate issue as it is not really closely related to this PR. Also, IMHO a rule of thumb would be to always have a proper address table for 12 OHs in LMDB on all test stands.

src/calibration_routines.cpp
mexanick: No, it is OK here, we never touch the chips above the `oh::VFATS_PER_OH`
So having it here as a maximum value constant is fine

mexanick:
@bdorney @lpetre-ulb your comments has been addressed except first one from @bdorney which I believe deserves a separate issue

src/calibration_routines.cpp
bdorney: > IMHO a rule of thumb would be to always have a proper address table for 12 OHs in LMDB on all test stands.

How feasible is this for AMC's not the CTP7? I don't know enough about how we are doing the GLIB to know this myself.

src/calibration_routines.cpp
bdorney: > No, it is OK here, we never touch the chips above the oh::VFATS_PER_OH
So having it here as a maximum value constant is fine

I guess my point here, and also with [this comment](https://github.com/cms-gem-daq-project/ctp7_modules/pull/132#discussion_r293507971) is to ensure a standardization through the repo where all constants/loop variables/etc... of this nature are specified in a single location.

Recall the issues we were debugging in the CSC office in 904 earlier this week?  We were using one namespace protected constant to get the VFAT mask but in another location we were looping over VFATs beyond that.  I think enforcing namespace protected constants in all cases like this would avoid possible issues in the future.

include/utils.h
bdorney: This comment is still relevant I think.

src/amc.cpp
```diff
@@ -44,12 +45,12 @@ unsigned int fw_version_check(const char* caller_name, localArgs *la)
...
-    uint32_t mask = 0x0;
-    for (int vfatN=0; vfatN<24; ++vfatN) { //Loop over all vfats
+    uint32_t mask = 0xffffff; //Start with all vfats masked for max VFAT/GEB amount
+    for (unsigned int vfatN=0; vfatN<oh::VFATS_PER_OH; ++vfatN) { //Loop over all vfats
         uint32_t syncErrCnt = readReg(la, stdsprintf("GEM_AMC.OH_LINKS.OH%i.VFAT%i.SYNC_ERR_CNT",ohN,vfatN));
 
-        if (syncErrCnt > 0x0) { //Case: nonzero sync errors, mask this vfat
-            mask = mask + (0x1 << vfatN);
+        if (syncErrCnt == 0x0) { //Case: zero sync errors, unmask this vfat
+            mask = mask - (0x1 << vfatN);
```
bdorney, line 53: To increase reliability of the output of this function we could; in the case that `syncErrCnt == 0x0` add some slow control transaction on this VFAT.  This could be done with `repeatedRegRead` (with a small number of reads to keep this function "fast") or it could be done with a one-off read of say of the hardware version register or something similar.

However I'm fine with letting this go and not being addressed in this PR.

bdorney: Also, nice solution.

bdorney:
After another pass on this PR I think the `struct ADCChannel` also needs to be done for both `ge11` and `ge21` namespaces as there are probably a different number (and location) of temperature sensors between GE2/1 and GE1/1 optohybrids: 

https://github.com/cms-gem-daq-project/ctp7_modules/blob/2c5d65f667e50ca95f733608e8768bf2086fd3f8/include/amc/sca_enums.h#L211-L267

Also I would suggest a look at all `struct`s and `enum`s declared here as they may *also* need namespace protection for things like different channels on the SCA ADC used to monitor different voltages, additional channels being used, additional monitoring, different pin out, etc...

I think we should address this given the board design has been completed in this PR.  Apologies for having missed this earlier.

src/calibration_routines.cpp
bdorney: I see we added the namespace protected constants which is good; but I think my first comment is still outstanding, e.g.:

> Is the sbit clustering and addressing system in GE1/1 the same as GE2/1? e.g. will these formulas to unpack the cluster word work for GE2/1?

Could we check with @andrewpeck that this is the case; if so then no issue.  Otherwise we will obviously need to handle this.

src/gbt.cpp
```diff
@@ -154,7 +174,7 @@ void writeGBTPhase(const RPCMsg *request, RPCMsg *response)
 
 bool writeGBTPhaseLocal(localArgs *la, const uint32_t ohN, const uint32_t vfatN, const uint8_t phase)
 {
-    LOGGER->log_message(LogManager::INFO, stdsprintf("Writing the VFAT #%u phase of OH #%u.", vfatN, ohN));
+    //LOGGER->log_message(LogManager::INFO, stdsprintf("Writing %u to the VFAT #%u phase of OH #%u.", phase, vfatN, ohN));
```
bdorney: I guess this was moved to line 194?  If so could we delete this commented line?

bdorney (changes requested):
Some comments still outstanding and a look at the `sca_enums.h` is needed I think.

src/calibration_routines.cpp
jsturdy: `OH_PER_AMC` (number of supported OHs) is a property of the FW, so @mexanick's point is feasible.
One could couple the LMDB fixed to 12, with any arrays fixed to 12 (or better yet, use dynamic containers), and in an implementation with fewer supported OHs, you have a check in whatever function so that you'll know what comes out is the "right size"

However, I agree that this is a separate issue, unless this code is targeting `develop` rather than the current release tree, as `develop` will be picking up the templating and this is much better suited to be addressed at that time.

bdorney: Current target branch is `develop`

src/calibration_routines.cpp
andrewpeck: Hi Brian, the cluster formula is the same at the moment (which means there is a wasted bit).. eventually we could change it to regain the bit for other purposes but compatibility seems to be a better target right now rather than optimizing bandwidth.. we also have no identified use for the extra bits right now

src/calibration_routines.cpp
lpetre-ulb: I fully agree that such constant should specified in a single location.

However, for this precise constant, what about removing it? I think it is only used in loops between 0 and `oh::VFATS_PER_OH` so it would be safe and we would get rid of one constant in both GE1/1 and GE2/1 code.

src/gbt.cpp
lpetre-ulb: Also, do we want to keep the `INFO` log level for this message since it the phase scans produce a lot of them?

What about the general semantic of log levels? How can we change the `rpcsvc` log level on the fly or at least when `rpcsvc` is started (e.g. one may want to trace all read and write calls during a debugging session)? There is no API to change it and it is fixed at compile time... I guess this is how out of the scope of this PR though.

lpetre-ulb (changes requested):
Some changes are still required, particularly in the SCA enums that I also completely overlooked during my first review. :-(

I think all the aliases should go into a `GEM_VARIANT` protected namespace since the mapping to ADC channels and GPIOs are differents. Once (or before) the ADC readout functions will be merged in the main branch some work will be also be required to make it compatible with GE2/1.

src/calibration_routines.cpp
bdorney: If you get rid of this constant how do you determine `notmask`?  Maybe I misunderstood the proposal.  Using only `~vfatmask[ohN]` will flip all bits no, e.g. bits [25, 32].  This could be problematic no?

src/gbt.cpp
bdorney: Yeah this probably needs it's own issue.  Not something for this PR.

src/calibration_routines.cpp
lpetre-ulb: As long as `notmask` is only used in `for`-loops it wouldn't be problematic. If it is used in other contexts then I fully support the constant in `hw_constants.h`.

I think the comment also applies to `ohMask` (line 610) : https://github.com/cms-gem-daq-project/ctp7_modules/blob/d384cbb0f58b93871e9b04799d2e3a70c1c1f6f2/src/calibration_routines.cpp#L606-L615

src/gbt.cpp
mexanick: I think the log level should be set by sysmgr, but I'm not 100% sure.

mexanick:
I agree that SCA enums has to be changed, but I don't have any info on this w.r.t. GE2/1. Can anyone point me to where I can get it?

lpetre-ulb:
@mexanick, you should be able to find all the information in the GE2/1 OH specification and/or on the GE2/1 OH schematic : http://padley.rice.edu/cms/OH_GE21/ohdesign.html

src/gbt.cpp
lpetre-ulb: Currently it is set [in `rpcsvc`](https://github.com/cms-gem-daq-project/rpcsvc/blob/f75364532253c93dbadece18b8be0fac7140236d/run_client.cpp#L6) and if the level given to `log_message` is below that value, the message is not even sent to the `syslog` daemon. I don't see how the `sysmgr` could help in this case... I'm going to open an issue.

mexanick:
Should these changes still go to develop or you prefer them to go under legacy/1.1.X branch?

mexanick:
Changed the target and performed a rebase

bdorney (approved):

bdorney:
Approving these changes per discussion in meeting; but if @mexanick you could open a new issue on the repo to address the SCA enum issues we discussed for GE1/1 vs. GE2/1 issues.

include/calibration_routines.h
```diff
@@ -54,15 +54,15 @@ struct vfat3DACAndSize{
     }
 };
 
-/*! \fn std::unordered_map<uint32_t, uint32_t> setSingleChanMask(int ohN, int vfatN, unsigned int ch, localArgs *la)
+/*! \fn std::unordered_map<uint32_t, uint32_t> setSingleChanMask(unsigned int ohN, int vfatN, unsigned int ch, localArgs *la)
```
jsturdy, line 57: any reason `ohN` is changed to unsigned but not `vfatN`? (here and elsewhere)

include/hw_constants.h
```diff
@@ -22,6 +22,12 @@ namespace amc {
         constexpr uint32_t OH_PER_AMC = 12;    ///< The number of OptoHybrids per AMC.
     }
 
+    /*! \brief GE2/1 specific namespace.
+     */
```
jsturdy, line 26: should standardize a type for these constants probably...

include/hw_constants.h
jsturdy: unrelated to this PR, but related to the comment, now would be a good time to convert the guide to a paginated format, a la gitbook

src/amc.cpp
```diff
@@ -44,12 +45,12 @@ unsigned int fw_version_check(const char* caller_name, localArgs *la)
 
 uint32_t getOHVFATMaskLocal(localArgs * la, uint32_t ohN)
 {
-    uint32_t mask = 0x0;
-    for (int vfatN=0; vfatN<24; ++vfatN) { //Loop over all vfats
+    uint32_t mask = 0xffffff; //Start with all vfats masked for max VFAT/GEB amount
+    for (unsigned int vfatN=0; vfatN<oh::VFATS_PER_OH; ++vfatN) { //Loop over all vfats
```
jsturdy, line 49: (here and elsewhere) why the affinity for `unsigned int` rather than `uintX_t` (as appropriate) or simply `unsigned`?

include/calibration_routines.h
mexanick: no good reason - has to be standardized indeed. However a further refactoring should be done under a separate PR (also including other comments like that)

jsturdy: in that case i would suggest to either do all the changes for this PR, or none of the changes for this PR, e.g., revert the changes that have been introduced that are only partial...

mexanick:
Merging as is, further refactoring foreseen

Issue #129, opened 2019-06-11
[feature request] Create a remote callable method for confCalPulseLocal


bdorney:
## Brief summary of issue

After discussing with @jsturdy we will need to have a remote callable method for `confCalPulseLocal`.  This should go only in the legacy branch, e.g. `release/v1.1.X`.

### Types of issue

- [ ] Bug report (report an issue with the code)
- [x] Feature request (request for change which adds functionality)

## Expected Behavior

https://github.com/cms-gem-daq-project/ctp7_modules/blob/develop/src/calibration_routines.cpp#L39

Have a remote callable method, called `confCalPulse`, for the local function `confCalPulseLocal`.

## Current Behavior

No remote callable method exists.

## Context (for feature requests)

When trying to develop a script similar to [set_scanParams_latency.py](https://github.com/cms-gem-daq-project/cmsgemos/blob/generic-amc-RPC-v3-short-term/gempython/tools/xdaq/set_scanParams_latency.py) but for scurves via tracking data path we will need a way to start/stop calibration pulses to a given set of channels.

ram1123:
Hi @bdorney,

I added the remote callable function as below:

```cpp
void confCalPulse(const RPCMsg *request, RPCMsg *response)
{
    GETLOCALARGS(response);

    uint32_t ohN = request->get_word("ohN");
    uint32_t mask = request->get_word("mask");
    uint32_t ch = request->get_word("ch");
    bool toggleOn = request->get_word("toggleOn");
    bool currentPulse = request->get_word("currentPulse");
    uint32_t calScaleFactor = request->get_word("calScaleFactor");

    confCalPulseLocal(&la, ohN, mask, ch, toggleOn, currentPulse, calScaleFactor);
    LOGGER->log_message(LogManager::DEBUG, stdsprintf("Finished configuring the calibration pulse."));

    rtxn.abort();
}
```
Also, I modified the corresponding header file. 

Its compiling without any error. How I can run this and see if the calibration pulse is configured correctly or not?

bdorney:
Reserve time on the coffin setup and use the same approach you used when testing the SCA ADC interface.

You can read back registers using the `gem_reg.py` tool.  e.g. try to use your function to write registers and then use `gem_reg.py` to see if your changes were applied.

PR #135, opened 2019-06-21
Added remote callable function for confCalPulseLocal

ram1123:
Added remote callable function for confCalPulseLocal as described in Issue #129.

## Description

Added remote callable function for confCalPulseLocal. Along with it, I also updated the error message of in the local function confCalPulseLocal for the scenario having VFAT channel > 128.  

Previous behaviour

https://github.com/cms-gem-daq-project/ctp7_modules/blob/ab4ddd0019976031c39251b719c698873b978c45/src/calibration_routines.cpp#L45-L48

Current behaviour

```cpp
    else if (ch >= 128) { //Case: Bad Config, asked for OR of all channels
        la->response->set_string("error","confCalPulseLocal(): I was told to calpulse (on/off) channels > 128.");
        return false;
    } //End Case: Bad Config, asked for OR of all channels
```
https://github.com/ram1123/ctp7_modules/blob/ad9d0906773d139067c0b7dc05fbd918e2f17aca/src/calibration_routines.cpp#L56-L59

### Types of changes

## How Has This Been Tested?

This has been tested by running the script using its various combination of input parameter. Then read the same written register using gem_reg.py tool.

## Checklist:

- [x] All  tests passed.

### DIFF
--- a/include/calibration_routines.h
+++ b/include/calibration_routines.h
@@ -347,4 +347,17 @@ void dacScanMultiLink(const RPCMsg *request, RPCMsg *response);
  */
 void genChannelScan(const RPCMsg *request, RPCMsg *response);
 
+/*! \fn void confCalPulse(const RPCMsg *request, RPCMsg *response)
+ *  \brief Generic method for calibration pulse configuration. Local callable version of confCalPulseLocal.
+ *  \param[in] request RPC response message with the following keys:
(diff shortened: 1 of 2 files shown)


bdorney (approved):

jsturdy (approved):
cursory

PR #144, opened 2019-07-15
[feature] real time monitoring

jsturdy:
## Description
Adds configurable monitoring module to `daq_monitor` for extracting register dumps
* Adds host executable to dump these to a fixed location
* Does not add (required) CTP7 file listing registers to be monitored
* Does not add (optional) `systemd` service script nor (optional) `logrotate` script that can manage this as a service

### Types of changes

## Motivation and Context
To have some way to monitor CTP7 registers over time for correlation with DAQ events

## How Has This Been Tested?
In use on `QC8` `eagle35` for the past week


src/daq_monitor.cpp
```diff
@@ -810,6 +811,65 @@ void getmonVFATLink(const RPCMsg *request, RPCMsg *response)
...
+  std::vector<uint32_t>    keyvalues;
+  for (auto const& regName : regNames) {
+    keynames.push_back(regName.second);
+    // keyvalues.push_back(regName.second, readReg(&la, regName.second));
+    response->set_word(regName.second, readReg(&la, regName.second));
+  }
+  response->set_string_array("keynames",keynames);
+  // response->set_word_array("keyvalues", keyvalues);
+  rtxn.abort();
+}
```
bdorney, line 871: If I understand this correctly, there's a file on the CTP7 that specifies which registers should be monitored; the default is `/mnt/persistent/gemdaq/mon_registers.txt"` but this can be overwritten at runtime by inserting the key `fname`.

Then for each register in `mon_registers.txt/fname` you read the reg, store the name of the reg as a key in `keynames` and then the reg value is then stored with that key.

So each call will monitor all the registers at once.

Here the registers in `mon_registers.txt/fname` are the *full* node name?

lpetre-ulb:
I guess it is too late now but for my knowledge, is there a specific reason why you did not use IPBus for this functionality? It seems very well suited to a long list of simple reads.

src/daq_monitor.cpp
jsturdy: Indeed, in fact, the register file can be directly obtained from `gem_reg.py`, as it is in the format that is printed from any register (list) read there

jsturdy:
> I guess it is too late now but for my knowledge, is there a specific reason why you did not use IPBus for this functionality? It seems very well suited to a long list of simple reads.

It's possible that it would have been fine (performance wise, for this number of registers, probably the CTP7 module will be faster, but I don't know by how much), but depending on the registers, one may have to use a different address table (`OptoHybrid` vs `AMC`)

PR #151, opened 2019-08-26
Add `log4cplus` initialization to the RPc modules

### LINKED ISSUE #139: Replace the "`LOGGER`" logging system with a full-featured logging library, e.g. `log4cplus`
## Brief summary of issue

The "`LOGGER`" logging system is limited and tightly coupled with the `rpcsvc` daemon. In order to implement some functionalities (standalone programs, log level change at runtime, ...) and to guarantee maintainability it may be time to move to a full-featured logging system.

### Types of issue

- [ ] Bug report (report an issue with the code)
- [x] Feature request (request for change which adds functionality)

## Expected Behavior

The following points should be taken into consideration while choosing a new solution:

* The functions and functors provided inside the `ctp7_modules` should be usable in standalone programs.
* The log level should be changeable at runtime. See this issue #134.
* The ATCA `rpcsvc` equivalent might not provide the "`LOGGER`" logging system.

## Current Behavior

The logging system is tightly coupled with the `rpcsvc` daemon.

@jsturdy was trying to write standalone programs for the CTP7 and faced [linking issues](https://mattermost.web.cern.ch/cms-gem-daq/pl/fpgnfk5ycfrqteebbc95yi18by).

Moreover the templated RPC refactoring starts to decouple the `ctp7_modules` from the `rpcsvc` server so it is the perfect time to change the logging system.

## Proposal

Since `xDAQ` and `xHAL` already use the `log4cplus` library I would suggest to use it as well on the CTP7. It is already compiled for `arm` and it would provide a common logging pattern.

A minimal example which logs to the CTP7 `syslog` daemon is trivial to write :

``` c++
#include <log4cplus/logger.h>
#include <log4cplus/configurator.h>
#include <log4cplus/loggingmacros.h>

int main(int, const char**)
{
    log4cplus::initialize();
    log4cplus::PropertyConfigurator::doConfigure("log.properties");
    log4cplus::Logger logger = log4cplus::Logger::getInstance(LOG4CPLUS_TEXT("main"));

    LOG4CPLUS_WARN(logger, LOG4CPLUS_TEXT("Hello, World!"));

    return 0;
}
```

with the following `log.properties` file:

``` txt
log4cplus.appender.syslog=log4cplus::SysLogAppender
log4cplus.appender.syslog.ident=syslog
log4cplus.appender.syslog.facility=user
log4cplus.appender.syslog.layout=log4cplus::TTCCLayout

log4cplus.rootLogger=DEBUG,syslog
```

and with the following compilation command:

``` bash
$ arm-linux-gnueabihf-g++ --sysroot=${PETA_STAGE} main.cpp -llog4cplus
```

Special care will have to be taken when initializing the library inside the RPC modules since they are all loaded independently. The best place would be inside the `rpcsvc` daemon where the "`LOGGER`" logging system is initialized.

lpetre-ulb:
As proposed in #139, this PR enables the use of the `log4cplus` logging system in the RPC modules.

## Description

At module initialization, the `log4cplus` system is initialized and its configuration reloaded. The configuration is read from `/mnt/persistent/gemdaq/etc/rpcsvc_log.properties` with a fallback to an embedded  default configuration. Therefore the different `log4cplus` logging macros can now be used through the code.

The default embedded configuration sends the logs from the `INFO` log level to the local `syslog` daemon with the pattern show below. It can overridden with the configuration file (not provided).

Currently the `log4cplus` library available on the CTP7 does not allow to use move semantics since it was compiled without C++11 support. One must then be careful how the API is used. This shortcoming will be lifted after a `log4cplus` recompilation.

Also, note that these ultra-legacy/stable modules have not been touched yet:

* `memory.so`
* `extra.so`
* `optical.so`

### Types of changes

## Motivation and Context

Closes #139 

## How Has This Been Tested?

Logging statements work as expected on the ULB setup. For e.g., the following logs are produced at module initialization when the logger configuration file is missing:
``` bash
Aug 26 22:17:39 CTP7 local0.notice rpcsvc[8866]: rpcsvc: Client connected from 192.168.249.191
Aug 26 22:17:39 CTP7 local0.info rpcsvc[8866]: memory:  Memhub initialized a semaphore. Current semaphore value = 1
Aug 26 22:17:39 CTP7 user.info rpcsvc: eagle63[8866] - void initLogging() - Impossible to read the configuration file; defaulting to the embedded configuration.
[...]
```

## Checklist:


Makefile
```diff
@@ -74,8 +74,8 @@ TargetObjects:= $(patsubst %.d,%.o,$(Dependencies))
 
 TargetLibraries:= memhub memory optical utils extras amc daq_monitor vfat3 optohybrid calibration_routines gbt
 
-# Everything links against these three
-BASE_LINKS = -lxhal -llmdb -lwisci2c
+# Everything links against these four
```
jsturdy: We can drop the number here for future maintainability :-)

include/utils.h
```diff
@@ -34,6 +34,23 @@ typedef struct localArgs {
     RPCMsg *response; /*!< RPC response message */
 } LocalArgs;
 
+/*!
+ * \brief \c log4cplus configuration filename
+ */
+constexpr auto LOGGING_CONFIGURATION_FILENAME = "/mnt/persistent/gemdaq/etc/rpcsvc_log.properties";
```
jsturdy: Would we want to get this from some env variable?

include/vfat_parameters.h
```diff
@@ -1,6 +1,6 @@
 #include <string>
 #include <map>
-std::map<std::string, uint32_t> vfat_parameters = {
+static std::map<std::string, uint32_t> vfat_parameters = {
```
jsturdy: These can definitely be dropped, as our updated architecture will not be supporting VFAT2, but if this change was only done for ability to compile, then I don't have a strong preference one way or the other for this PR

src/amc.cpp
```diff
@@ -10,6 +10,9 @@
 #include "amc/daq.h"
 #include "amc/blaster_ram.h"
 
+#include <log4cplus/logger.h>
+#include <log4cplus/loggingmacros.h>
+
```
jsturdy, line 14: Any reason *not* to include these via `utils.h` (which I believe all modules will have to include at some level)?

src/utils.cpp
```diff
@@ -35,6 +42,32 @@ std::string serialize(xhal::utils::Node n)
...
+    log4cplus::Logger::getDefaultHierarchy().resetConfiguration();
+
+    std::ifstream configurationFile{LOGGING_CONFIGURATION_FILENAME};
+
+    if (configurationFile) {
+        log4cplus::PropertyConfigurator configurator{configurationFile};
+        configurator.configure();
+    }
+    // Fallback to the default embedded configuration
+    else {
```
jsturdy: `if/else` brace style

jsturdy (commented):
Couple of small comments

include/utils.h
lpetre-ulb: The path for the configuration file was agreed in the issue, but I changed it to an environment variable. It makes sense since there is now a default embedded configuration.

include/vfat_parameters.h
lpetre-ulb: It was in my commit history after some GLIB emulator tests; I'll drop the commit. It is not required for proper compilation, only to avoid a double free at exit. Removing all the VFAT2 leftovers can be addressed in another PR.

src/utils.cpp
lpetre-ulb: I couldn't find the `if/else` brace style in the [`cmsgemos` contributing guide](https://github.com/cms-gem-daq-project/cmsgemos/blob/develop/.github/CONTRIBUTING.md). I hope to have got these right now.

src/amc.cpp
lpetre-ulb: No technical reason, so I moved these includes in `utils.h`.

However, I'm usually not fond of indirectly including macros/functions through a chain of headers, especially with an header like `utils.h` which contains both local functions and indirect includes; it makes the code harder to read. Can't we envision to move all these common includes to a `common.h` header which  could then be eventually precompiled?

lpetre-ulb:
Comments are addressed. However, I think that the `utils.h` file should be refactored sooner than later in order to avoid creating a file which contains everything that has no place somewhere else.

include/utils.h
```diff
@@ -7,21 +7,23 @@
 #ifndef UTILS_H
 #define UTILS_H
 
-#include "moduleapi.h"
 //#include <libmemsvc.h>
+#include "moduleapi.h"
 #include "memhub.h"
+
 #include "lmdb_cpp_wrapper.h"
```
jsturdy, line 14: Is this the same file that is now included in `xhal`?
If so, maybe we can include from there

jsturdy (approved):
Looks good :+1:

Re the `utils.h`/`common.h`, yes, we should think of how best to do this, and yes, now (well, after this PR is merged and we start the rest of the migration) is the right time to do that

include/utils.h
lpetre-ulb: Yes, they are the same, so `lmdb_cpp_wrapper.h` should be removed from the `ctp7_modules`. However, I think we should wait for https://github.com/cms-gem-daq-project/xhal/pull/133 to be merged because the header location will change. (Also, this is technically out of the scope of this PR ;-)

jsturdy: sure :-)
was just for information :+1:

Issue #138, opened 2019-07-05
Adapt the CTP7 modules for the OH modifications


lpetre-ulb:
## Brief summary of issue

The temperature measured by the PT100 in the original version of the OH showed a large bias (more than 10°C), at least on some boards. This was traced back to three factors : 

1. The SCA current source only provides *nominally* 100 µA.
2. The PT100 does not match well the characteristics of the SCA ADC. 
3. The resistance of the routing corresponds to a few degrees in case of the PT100.

Three decisions have been taken to mitigate this issue :

1. The 1.8 V voltage measurement input has been converted to a SCA ADC current source measurement input with the help of a precision 1 kOhm resistor +/- 0.05%.
2. The PT100 has been replaced by a PT1000.
3. The routing resistance will be measured next Monday on blank PCBs.

Due to timing constraints all the OHs are not modified. Three variants will exists :

1. Original version with the PT100 and no current measurement.
2. Partially patched version with the P100, but a precision resistor for current measurement.
3. Fully patched version with the PT1000 and the precision resistor.

This information will be available in the database.

Another detected issue is the noise. On the following plot the distribution of 250 measurements of the 1 kOhm precision resistor is shown :

![Figure_1](https://user-images.githubusercontent.com/33247723/60725190-b7ac0600-9f38-11e9-8d0c-9bbc618b24f6.png)

It is clear there is a large variation, higher than the 1 LSB quantization error of the SCA ADC. In order to get sensitive data the final value should then average multiple measurements. There is obviously a trade-off between the measurement time and the acceptable uncertainty, so a configurable number of measurement would be ideal.

### Types of issue

- [ ] Bug report (report an issue with the code)
- [x] Feature request (request for change which adds functionality)

## Expected Behavior

Theses new informations should be taken into account into the RPC module :

* The `ADC_V1P8` enum value should be renamed and used according to its new real use. The `VTTX_CSC_PT100` may be renamed since there are not always PT100.
* Providing a way to specify the number of measurements per ADC input channel seems important.

lpetre-ulb:
Here is [the datasheet](https://www.heraeus.com/media/media/group/doc_group/products_1/hst/smd/us_5/SMD1206_HSTUSA.pdf) for the new PT1000; it is an Heraeus PRTD SMD 1206 (part 32207595).

[...]

PR #157, opened 2019-10-23
[bugfix] SCA readout update

jsturdy:
## Description
Several issues were present with the SCA readout:
* The changes to the `.ADC_MONITORING.MONITORING_OFF` register had not been forward propagated
  * Added a new function in `utils` ` bool regExists(localArgs * la, const std::string & regName, lmdb::val * db_res)` that is called in all the read/write functions in place of the register name lookup, as this allows other code to check whether the register exists in the LMDB, even if they don't need to create a valid `lmdb::val` object
* The output data format was not tested for 12 OptoHybrids, and was truncating

### Types of changes

## Motivation and Context
Needed to reliably monitor the sensors from the SCA

## How Has This Been Tested?
Problem discovered and tested at P5

### DIFF
--- a/include/amc/sca.h
+++ b/include/amc/sca.h
@@ -183,12 +183,13 @@ void resetSCASEUCounter(const RPCMsg *request, RPCMsg *response);
  *
  *  \param[out] response RPC response message with the following keys:
  *  	- `word_array data` : ADC data is returned as an array of 32-bit words formatted as:
- *  			      bit [27]: data present
- *  			      bits [26:24]: link ID
+ *  			      bits [31:29]: constant 0's
+ *  			      bit  [28]   : data present
+ *  			      bits [27:24]: link ID
  *  			      bits [23:21]: constant 0's
  *  			      bits [20:16]: ADC channel ID
  *  			      bits [15:12]: constant 0's
- *  			      bits [11:0]: ADC data
+ *  			      bits [11:0] : ADC data
  */
 void readSCAADCSensor(const RPCMsg *request, RPCMsg *response);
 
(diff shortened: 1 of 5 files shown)


include/amc/sca_enums.h
```diff
@@ -212,7 +212,7 @@ class SCASettings {
     enum EADCChannel { //ADCChannel settings
```
lpetre-ulb, line 212: Not technically part of this PR, but is there a specific reason why you preferred `enum`'s in `struct`'s instead of scope enumerations in this file? Using strongly-typed aliases would help preventing bugs and would allow using functions overloads (e.g. replace all `sca*Command` with a single overloaded `scaCommand`).

Also not part of this PR, the GE1/1 aliases should be updated to reflect the latest [OptoHybrid changes ](https://github.com/cms-gem-daq-project/ctp7_modules/issues/138).

src/utils.cpp
```diff
@@ -148,14 +148,23 @@ uint32_t getNumNonzeroBits(uint32_t value)
   return numNonzeroBits;
 }
 
-uint32_t getMask(localArgs * la, const std::string & regName)
+bool regExists(localArgs * la, const std::string & regName, lmdb::val * db_res)
```
lpetre-ulb, line 151: It is good to have this elementary functionality shared. As as note, the DAC scans will also have to be adapted at some point in order to avoid using low-level LMDB access throughout the code.

include/amc/sca.h
```diff
@@ -183,12 +183,13 @@ void resetSCASEUCounter(const RPCMsg *request, RPCMsg *response);
  *
  *  \param[out] response RPC response message with the following keys:
  *  	- `word_array data` : ADC data is returned as an array of 32-bit words formatted as:
- *  			      bit [27]: data present
- *  			      bits [26:24]: link ID
+ *  			      bits [31:29]: constant 0's
+ *  			      bit  [28]   : data present
+ *  			      bits [27:24]: link ID
```
lpetre-ulb, line 188: I was wondering how many OHs will be supported by the foreseen ATCA board? I imagine that it will be more than 16. As a bugfix for the legacy modules, the change it certainly more than enough, but the format should probably be improved when moving to the templated RPC modules.

src/amc/sca.cpp
```diff
@@ -18,7 +18,10 @@ uint32_t formatSCAData(uint32_t const& data)
 
 void sendSCACommand(localArgs* la, uint8_t const& ch, uint8_t const& cmd, uint8_t const& len, uint32_t data, uint16_t const& ohMask)
 {
-  // FIXME: DECIDE WHETHER TO HAVE HERE // writeReg(la,"GEM_AMC.SLOW_CONTROL.SCA.ADC_MONITORING.MONITORING_OFF",         0xffffffff);
+  // FIXME: DECIDE WHETHER TO HAVE HERE // if (regExists(la, "GEM_AMC.SLOW_CONTROL.SCA.ADC_MONITORING.MONITORING_OFF")) {
+  // FIXME: DECIDE WHETHER TO HAVE HERE //   writeReg(la,"GEM_AMC.SLOW_CONTROL.SCA.ADC_MONITORING.MONITORING_OFF",         0xffffffff);
+  // FIXME: DECIDE WHETHER TO HAVE HERE // }
+
```
lpetre-ulb, line 24: Since all calls to `sendSCACommand` need to be protected against the ADC monitoring code, wouldn't it better to directly move the protection here?

lpetre-ulb (approved):
The bugs seem to be addressed. The comments only concern future PR's and/or the migration to the templated RPC modules.

include/amc/sca_enums.h
jsturdy: Yes, I had used the `struct { enum}` because there is code in `cmsgemos` using the same, and it was foreseen that these headers are part of the files that can be included there.
I fully expect that very shortly these will be replaced with the strongly typed `enum class` syntax, as we're moving to a (more) modern compiler :-)

src/amc/sca.cpp
jsturdy: Yes, I think this was my initial proposal as well, but don't remember the reason for not implementing it.
In principle, it can also be removed at some point... but yeah... another time, another PR :-)

include/amc/sca.h
jsturdy: indeed, this format should be adapted for an expected future system before it gets into the production stream (templated RPCs)

mexanick (approved):

Issue #162, opened 2019-12-11
[cleanup] Common blocks related to supOH


jsturdy:
## Brief summary of issue
As part of the porting (#160), I've come across several regularly repeated blocks of code I'm trying to clean up.
The first is some check against the FW supported number of OptoHybrid links, e.g.,
```c++
void function(uint32_t ohMask)
  uint32_t supOH = readReg(&la, "GEM_AMC.GEM_SYSTEM.CONFIG.NUM_OF_OH");
  if (request->get_key_exists("NOH")) {
    uint32_t NOH_requested = request->get_word("NOH");
    if (NOH_requested <= supOH)
      supOH = NOH_requested;
    else
      LOG4CPLUS_WARN(logger, stdsprintf("NOH requested (%i) > NUM_OF_OH AMC register value (%i), NOH request will be disregarded",NOH_requested,supOH));
  }
  for (size_t ohN = 0; ohN < supOH; ++ohN) {
    if (!((ohMask >> ohN) & 0x1)) {
      continue;
    }

    // here be the meat
  }
}
```

I tried initially to just simplify this with the following
```c++
  uint32_t supOH = readReg(&la, "GEM_AMC.GEM_SYSTEM.CONFIG.NUM_OF_OH");
  uint32_t reqOH = getNumNonzeroBits(ohMask);
  if (reqOH <= supOH)
    supOH = reqOH;
  else
    LOG4CPLUS_WARN(logger, "Requested OptoHybrids ("<< reqOH << ") > NUM_OF_OH AMC register value ("<< supOH << "), request will be reset to register max");
  }
...
```

I believe this captures the *intention* of the original code, however it can fail in the actual implementation, e.g., if the `ohMask` is sparse, but still toggles on a link that is not supported, **or**, *worse*, if the `ohMask` is sparse, not toggling any unsupported OHs, but as a byproduct, limits the range of the loop.  When I realized this, I started trying to think of a better solution.
This is safe:
```c++
  uint32_t supOH = readReg(&la, "GEM_AMC.GEM_SYSTEM.CONFIG.NUM_OF_OH");
  if (ohMask > (0x1<<supOH))
    LOG4CPLUS_WARN(logger, "Requested OptoHybrids (0x"<< std::hex << std::setfill('0') << std::setw(8) << ohMask << std::dec << ") > NUM_OF_OH AMC register value ("<< supOH << "), request will be reset to register max");
  }
```

[...]

PR #164, opened 2019-12-17
WIP [feature] templated RPC methods migration

jsturdy:
This PR holds the code of the migration to the templated methods.
It will be built up as the individual modules are merged into my fork, and will only become non WIP after all code has found its way into that fork (so review of the independent parts can still happen in bite-size chunks).

As it stands here, the changes have been applied to the core libraries as well as the `utils` and `amc` modules.

## Description
Significant and breaking

### Types of changes

## Motivation and Context
Necessary to bring the code base into the future.

## How Has This Been Tested?
Only with the compiler so far, will envision a set of operational tests to follow before final merge.

## Checklist:
- [x] I have updated the documentation accordingly (hopefully)


include/amc.h
```diff
@@ -10,65 +10,97 @@
...
-/*! \fn uint32_t getOHVFATMaskLocal(uint32_t ohN)
- *  \brief returns the vfatMask for the optohybrid ohN
- *  \details Reads the SYNC_ERR_CNT counter for each VFAT on ohN.  If for a given VFAT the counter returns a non-zero value the given VFAT will be masked.
- *  \param la Local arguments structure
- *  \param ohN Optical link
- */
-uint32_t getOHVFATMaskLocal(localArgs * la, uint32_t ohN);
+  /*!
+   *  \brief Returns AMC FW version
+   *         In the case FW version is not 1.X or 3.X, throws std::runtime_error
```
lmoureaux:
```C++
/*!
 * \throws std::runtime_error if the FW version is not 1.X or 3.X
 */
```

include/amc.h
```diff
@@ -10,65 +10,97 @@
...
+   *  \param \c caller_name Name of method which called the FW version check FIXME necessary?
```
lmoureaux: I guess the parameter is not needed, now would be a good time to drop it or at least make it optional.

include/amc.h
```diff
@@ -10,65 +10,97 @@
...
- */
-void getOHVFATMask(const RPCMsg *request, RPCMsg *response);
+  /*!
+   *  \brief returns the vfatMask for the optohybrid ohN
+   *  \details Reads the SYNC_ERR_CNT counter for each VFAT on ohN.  If for a given VFAT the counter returns a non-zero value the given VFAT will be masked.
+   *  \param ohN Optical link identifier
+   */
+  struct getOHVFATMask : public xhal::rpc::Method
+  {
+    uint32_t operator()(const uint32_t &ohN) const;
```
lmoureaux, line 38: Why constant references for parameters? I can't think of any reason to prefer them over values (for scalars; for heavier objects there is a good reason).

include/amc.h
```diff
@@ -10,65 +10,97 @@
...
 
-/*! \fn void getOHVFATMaskMultiLink(const RPCMsg *request, RPCMsg *response)
- *  \brief As getOHVFATMask(...) but for all optical links specified in ohMask on the AMC
- *  \details Here the RPCMsg request should have a "ohMask" word which specifies which OH's to read from, this is a 12 bit number where a 1 in the n^th bit indicates that the n^th OH should be read back.  Additionally there should be a "ohVfatMaskArray" which is an array of size 12 where each element is the standard vfatMask for OH specified by the array index.
- *  \param request RPC request message
- *  \param response RPC response message
- */
-void getOHVFATMaskMultiLink(const RPCMsg *request, RPCMsg *response);
+  /*!
+   *  \brief As \c getOHVFATMask but for all optical links specified in ohMask on the AMC
```
lmoureaux, line 42: I suggest repeating the documentation instead of referring to another. If not, use `\ref` instead of `\c`.

include/amc.h
```diff
@@ -10,65 +10,97 @@
...
- *  \param la Local arguments structure
- *  \param ohN Optical link
- */
-uint32_t getOHVFATMaskLocal(localArgs * la, uint32_t ohN);
+  /*!
+   *  \brief Returns AMC FW version
+   *         In the case FW version is not 1.X or 3.X, throws std::runtime_error
+   *  \param \c caller_name Name of method which called the FW version check FIXME necessary?
+   */
+  uint32_t fw_version_check(const char* caller_name);
```
lmoureaux, line 24: Documentation: what does the return value contain? `1` or `3`?

include/amc.h
lmoureaux: Support for returning `std::bitset` would come handy here.

include/amc.h
```diff
@@ -10,65 +10,97 @@
...
+  /*!
+   *  \brief As \c getOHVFATMask but for all optical links specified in ohMask on the AMC
+   *
+   *  \param ohMask a 12 bit number where a 1 in the n^th bit indicates that the n^th OH should be read back.
+   *  \param numOH number of OHs enabled in the mask FIXME redundant
+   *  \returns a vector of the VFAT mask for the unmasked OHs
+   */
+  struct getOHVFATMaskMultiLink : public xhal::rpc::Method
+  {
+    std::vector<uint32_t> operator()(const uint32_t &ohMask=0xfff, const uint32_t &numOH=0) const;
```
lmoureaux: The default `numOH` is inconsistent with the default `ohMask`. In fact `numOH` should be dropped altogether (again `std::bitset` would be useful).

include/amc.h
```diff
@@ -10,65 +10,97 @@
...
-/*! \fn void sbitReadOut(const RPCMsg *request, RPCMsg *response)
- *  \brief readout sbits using the SBIT Monitor.  See the local callable methods documentation for details.
- *  \param request RPC response message
- *  \param response RPC response message
- */
-void sbitReadOut(const RPCMsg *request, RPCMsg *response);
+  /*!
+   *  \brief reads out s-bits from OptoHybrid ohN for a number of seconds given by acquireTime and returns them to the user.
+   *
+   *  \details The s-bit Monitor stores the 8 s-bits that are sent from the OH (they are all sent at the same time
```
lmoureaux, line 56: General remark: `\details` is not necessary as long as a new paragraph is started by leaving a line empty.

include/amc.h
```diff
@@ -10,65 +10,97 @@
...
+   *  \param nReads the number of times each register in \c regList is read
+   *
+   *  \returns an std::map<std::string,uint32_t> with the following keys, which correspond to the counters under the node "GEM_AMC.SLOW_CONTROL.VFAT3":
+   *          "CRC_ERROR_CNT"
+   *          "PACKET_ERROR_CNT"
+   *          "BITSTUFFING_ERROR_CNT"
+   *          "TIMEOUT_ERROR_CNT"
+   *          "AXI_STROBE_ERROR_CNT"
+   *          "TRANSACTION_CNT"
+   *         additionally there will be one final key "SUM" which is the sum of all counters (except TRANSACTION_CNT).
```
lmoureaux, line 106: I don't like using string keys, but it's better than RPC keys since error checking is possible. I'd prefer a custom `struct` with the appropriate (un)serializer, or maybe an `array` indexed with an `enum`.

include/amc/blaster_ram.h
```diff
@@ -8,263 +8,250 @@
 
 #include "utils.h"
 #include "amc/blaster_ram_defs.h"
+namespace amc {
+    namespace blaster {
+        /*!
+         *  \brief Returns the size of the specified RAM in the BLASTER module
+         *
+         *  \param la Local arguments structure
```
lmoureaux: This parameter was dropped (five other occurrences in this file)

include/amc/blaster_ram.h
```diff
@@ -8,263 +8,250 @@
...
- *  \param la Local arguments structure
- *  \param ohN Select which OptoHybrid the component is associated with (0--11)
- *  \param partN Select which VFAT/GBTx the configuration is for
- *         0--23 (VFAT)
- *         0--2 (GBTx)
- *  \returns address in the block RAM specified
- */
-uint32_t getRAMBaseAddr(localArgs *la, BLASTERTypeT const& type, uint8_t const& ohN, uint8_t const& partN);
+        /**
+           read functions
```
lmoureaux: If this is to be a Doxygen comment, start a group (iirc with `\{`). If not, remove the second `*`

include/amc/blaster_ram.h
```diff
@@ -8,263 +8,250 @@
...
+        **/
 
-/**
-   read functions
-**/
+        /*!
+         *  \brief Reads configuration `BLOB` from BLASTER RAM
+         *
+         *  The CTP7 has three RAMs that store configuration information for `GBT`, `OptoHybrid`, and `VFAT`
+         *  The `BLOB` provided here will conatin the configuration blob for one (or all) of the three RAMs
```
lmoureaux: contain

include/amc/blaster_ram.h
```diff
@@ -8,263 +8,250 @@
...
+         *         BLASTERType::OptoHybrid - will return all OptoHybrid configurations for a single CTP7 BLASTER RAM
+         *         BLASTERType::VFAT - will return all VFAT configurations for a single CTP7 BLASTER RAM
+         *         BLASTERType::ALL - will return the full configuration of a single CTP7 BLASTER RAM
+         *  \param blob to store the configuration `BLOB`
+         *  \param blob_sz number of 32-bit words in configuration `BLOB`.
+         *         Must be equal to the size of the RAM specified:
+         *         * if the size is more, an error will be thrown
+         *         * if the size is less, only the number of words specified will be read from the RAM (maybe)
+         *         * if ALL is specified, the size must be the total RAM size
+         *           - GBT_RAM_SIZE*N_GBTX*N_OH + OH_RAM_SIZE*N_OH + VFAT_RAM_SIZE*N_VFAT*N_OH
```
lmoureaux, line 66: Could introduce a function to calculate this.

include/amc/blaster_ram.h
```diff
@@ -8,263 +8,250 @@
...
+         *         Must not exceed GEM_AMC.CONFIG_BLASTER.STATUS.GBT_RAM_SIZE +
+         *                         GEM_AMC.CONFIG_BLASTER.STATUS.OH_RAM_SIZE +
+         *                         GEM_AMC.CONFIG_BLASTER.STATUS.VFAT_RAM_SIZE
+         */
+        struct readConfRAM : xhal::rpc::Method {
+            uint32_t operator()(BLASTERTypeT const& type, uint32_t* blob, size_t const& blob_sz);
```
lmoureaux: This method cannot be called remotely because of the pointer parameter. Return a `std::vector`? Same issue for several RPC methods after this one.

include/amc/blaster_ram_defs.h
```diff
@@ -14,6 +14,13 @@ class BLASTERSettings {
       VFAT       = 0x4, ///< VFAT RAM
       ALL        = 0x7, ///< All RAMs
     } BLASTERType;
+
+    /*!
+     * \brief With its intrusive serializer
+     */
+    template<class Message> void serialize(Message & msg) {
+      msg & EBLASTERType;
```
lmoureaux: Does this compile? `EBLASTERType` is the enum type, shouldn't it be `BLASTERType`?

lmoureaux (commented):
Cursory look at the first files in alphabetical order, mostly interface & documentation comments

include/amc.h
jsturdy: What were the use cases originally for including it?
Just to pass it to a potential `LOGGER` message?

include/amc.h
jsturdy: Guess this will have to be updated globally, as there are quite some places that now `throw`

include/amc.h
jsturdy: Yes, this was being discussed in #162 and all specifications of `numOH` will be removed

include/amc.h
jsturdy: We can always move to a more custom solution, however this works with the current implementation.
What's the concern here with a `std::string` key vs some `enum` key?

include/amc/blaster_ram.h
jsturdy: Oops, my find/replace must not have been active here yet...

include/amc/blaster_ram.h
jsturdy: such shame :-(

include/amc/blaster_ram.h
jsturdy: Oops, this was a byproduct of partial implementation...

include/amc/blaster_ram_defs.h
jsturdy: This particular commit *probably* doesn't, but the compilation is fixed in a later commit.
As I go through to fix the comments here, I'll backport that

jsturdy: What do ya know, it actually *does* compile :-D, but this still needs the update viz scoped `enum`s

include/amc.h
jsturdy: I think `\c` was an artifact, so not sure who introduced it nor what it does

include/amc.h
jsturdy: updated (not pushed)

include/amc.h
jsturdy: removed (not pushed)

include/amc/blaster_ram.h
jsturdy: updated (not pushed)

include/amc/blaster_ram.h
jsturdy: removed (not pushed)

include/amc.h
lmoureaux: `\c` displays the following word in typewriter font. `\ref` is for a reference.

include/amc.h
lmoureaux: Misspelled keys detected only at runtime. But this is also the case if one uses the wrong enum (since there would be an implicit conversion to `int` behind the scenes)

mexanick (approved):
Cursory approval for testing

jsturdy:
I opened #167 to supersede this (it contains all the changes mentioned in the description of this, and is the hopeful starting point for integrated development between `xhal`/`cmsgemos`/`ctp7_modules` in the templated RPC framework)

Issue #166, opened 2020-01-15
[feature request] Compiling for x86 misses libmemsvc


jsturdy:
## Brief summary of issue
In order to compile for `x86_64` (GLIB emulator) an equivalent of `libmemsvc` must be provided.
Currently this file does not exist in `ctp7_modules` and @lpetre-ulb has a version that has been used for generating the GLIB emulator [lpetre-ulb/gem_glib_sw](https://github.com/lpetre-ulb/gem_glib_sw)
Somehow this needs to be included in the set of GEM tools, whether added here to `ctp7_modules` or independently as a dependency elsewhere.
I and @mexanick propose that it is included here in `ctp7_modules` to reduce the number of multiplying packages.

### Types of issue
- [x] Feature request (request for change which adds functionality)

## Expected Behavior
Compiling for `x86_64` should work

## Current Behavior
Doesn't work

### Steps to Reproduce (for bugs)
```
make
```

## Possible Solution (for bugs)
Include sources for `IPBus` based `libmemsvc` to `ctp7_modules` (this was how I resolved the issue for my testing, it was added to a `server` subdirectory as seen in the `ctp7_modules.mk` file)

## Your Environment
* Version used: `366ee77`

jsturdy:
If my "possible solution" is desired, I will add those changes and push that.

lpetre-ulb:
First of all, the maintained version of the `gem-emulator` is [on the CERN GitLab](https://gitlab.cern.ch/lpetre/gem-emulator) for its great CI integration and no longer on GitHub ;)

About the `libmemsvc` wrapper, I would prefer to not include it in the `ctp7_modules`. Simply for a separation of concerns between repositories. The infrastructure/helpers would better live in `xhal` rather than in the `ctp7_modules` which implements the real actions.

I'd consider two other ideas:

1. Create a sysroot image similar to what exists for the CTP7; the `libmemsvc` wrapper would then continue to live in `gem-emulator`. The mid-term goal is to include `gem-emulator` into the `GEM_AMC` repository since the container image and the firmware need to be kept in sync. One can consider the container as the Linux image of the CTP7.

[...]

PR #167, opened 2020-01-23
[feature] templated RPC + package structure

jsturdy:
## Description
Supersedes #164, includes *all* modules migrated as well as consistency updates to the package structure.

### Types of changes

## Motivation and Context
#164

## How Has This Been Tested?
Compiles for `arm`, lacks `x86_64` due to #166


.gitlab-ci.yml
```diff
@@ -0,0 +1,346 @@
...
+## ------------------- Define up anchors for various tasks, shared across all jobs utilizing them ------------------- ##
+.common_variables: &common_variables
+  EOS_SITE_URL:           https://cmsgemdaq.cern.ch/cmsgemdaq
+  BUILD_HOME:             /builds/${CI_PROJECT_NAMESPACE}
+  XDAQ_OS:                linux
+  XDAQ_ROOT:              /opt/xdaq
+  LD_LIBRARY_PATH:        /opt/xdaq/lib:/opt/xhal/lib:/opt/reedmuller/lib:/opt/wiscrpcsvc/lib
+  PACKAGE_NAME:           ${CI_PROJECT_NAME}
+  ARTIFACTS_DIR:          artifacts
+  XILINX_SDK_PATH:        /data/bigdisk/sw/Xilinx/SDK
```
lpetre-ulb, line 43: Is it useful to export this environment variable since the CI uses the Linaro toolchain? Can `/data/bigdisk` be mounted in the GitLab CI Docker containers?

.gitlab-ci.yml
```diff
@@ -0,0 +1,346 @@
...
+  pip install -I --user "pip" "importlib" "codecov" "setuptools<38.2" "sphinxcontrib-napoleon==0.2.9"
+  sudo yum install -y wiscrpcsvc\* --enablerepo=gemos*
+  sudo yum install -y reedmuller\* --enablerepo=gemos*
+  # for building releases, need to ensure that we're using the correct stream...
+  # for a generic build, this could be done against a testing stream
+  # sudo yum install -y xhal\*     --enablerepo=gemos* ## get specific version of xhal to test against
+  . /opt/rh/devtoolset-8/enable
+
+.common_zynq_setup: &common_zynq_setup |-
+  sudo yum install -y gem-peta-stage-ctp7 --enablerepo=gemos*
```
lpetre-ulb, line 74: Where is the `gem-peta-stage-ctp7` spec file hosted?

.gitmodules
```diff
@@ -1,3 +1,3 @@
 [submodule "config"]
 	path = config
-	url = git@github.com:cms-gem-daq-project/gembuild.git
+	url = https://github.com/cms-gem-daq-project/gembuild.git
```
lpetre-ulb, line 3: Wouldn't it be better to use relative URLs? At least everyone would be able to use its favorite protocol.

.gitlab-ci.yml
```diff
@@ -0,0 +1,346 @@
...
+    - /^citest.*$/@cms-gem-daq-project/ctp7_modules
+    - /^develop$/@cms-gem-daq-project/ctp7_modules
+    - /^master$/@cms-gem-daq-project/ctp7_modules
+    - merge_requests@cms-gem-daq-project/ctp7_modules
+    - tags@cms-gem-daq-project/ctp7_modules
+  tags:
+  dependencies:
+    - package:cc7
+  after_script:
+    - 'echo "FIXME: github_changelog_generator"'
```
lpetre-ulb, line 329: Do you have a reliable way to automatically generate the changelog? I believe it would have to be manually reviewed in any case in order to get high quality information.

ctp7_modules.mk
```diff
@@ -0,0 +1,267 @@
...
+ProjectPath:=$(BUILD_HOME)/$(Project)
+
+ConfigDir:=$(ProjectPath)/config
+
+include $(ConfigDir)/mfCommonDefs.mk
+
+ifeq ($(Arch),x86_64)
+include $(ConfigDir)/mfPythonDefs.mk
+CFLAGS+=-Wall -fPIC
+ADDFLAGS=-std=c++1y -std=gnu++1y -m64
```
lpetre-ulb, line 25: Since we are now targetting `devtoolset-8` for the `x86_64` architecture, shouldn't `c++1y` become `c++14`? Also, there is no need to pass both `c++1y` and `gnu++1y` as `std` flags: https://stackoverflow.com/a/43956978

ctp7_modules.mk
```diff
@@ -0,0 +1,267 @@
...
+PackageExecDir:=$(PackagePath)/bin
+# PackageDocsDir:=$(PackagePath)/doc/_build/html
+
+CTP7_MODULES_VER_MAJOR:=$(shell $(ConfigDir)/tag2rel.sh | awk '{split($$0,a," "); print a[1];}' | awk '{split($$0,b,":"); print b[2];}')
+CTP7_MODULES_VER_MINOR:=$(shell $(ConfigDir)/tag2rel.sh | awk '{split($$0,a," "); print a[2];}' | awk '{split($$0,b,":"); print b[2];}')
+CTP7_MODULES_VER_PATCH:=$(shell $(ConfigDir)/tag2rel.sh | awk '{split($$0,a," "); print a[3];}' | awk '{split($$0,b,":"); print b[2];}')
+
+INSTALL_PREFIX?=/mnt/persistent/ctp7_modules
+
+ifeq ($(and $(XHAL_ROOT),$(BUILD_HOME)),)
```
lpetre-ulb, line 48: Cannot these variables auto-determined as for `cmsgemos`?

ctp7_modules.mk
```diff
@@ -0,0 +1,267 @@
...
+
+ifeq ($(Arch),x86_64)
+LibraryDirs+=$(XHAL_ROOT)/lib
+LibraryDirs+=$(REEDMULLER_ROOT)/lib
+LibraryDirs+=$(WISCRPC_ROOT)/lib ## for libmemsvc
+LibraryDirs+=$(XDAQ_ROOT)/lib    ## for log4cplus and xerces
+LibraryDirs+=$(CACTUS_ROOT)/lib  ## uhal for libmemsvc
+else
+## Need solution for system install as well as test install against specific new version
+LibraryDirs+=$(PETA_STAGE)/mnt/persistent/xhal/lib
```
lpetre-ulb, line 82: Aren't the CTP7 libraries installed in `/opt/xhal/lib/arm`?

include/ctp7_modules/common/daq_monitor.h
```diff
@@ -0,0 +1,185 @@
...
+   *  \brief Reads the SCA Monitoring values of all OH's (voltage and temperature); these quantities are reported in ADC units
+   *
+   *  \deprecated
+   *
+   *  \param \c ohMask A 12 bit number which specifies which OptoHybrids to read from.
+   *         Having a value of 1 in the n^th bit indicates that the n^th OptoHybrid should be considered.
+   *
+   *  \returns \c std::map
+   */
+  struct [[deprecated]] getmonOHSCAmain : xhal::common::rpc::Method
```
lpetre-ulb, line 125: Isn't it the perfect time to remove this RPC method instead of deprecating it? Maybe in a future PR?

include/ctp7_modules/common/optohybrid.h
```diff
@@ -0,0 +1,236 @@
...
+   *
+   *  \deprecated
+   *
+   *  \param \c ohN
+   *  \param \c trimConfigFile
+   *  \param \c threshConfigFile
+   *  \param \c vt1
+   *  \param \c setRunMode
+   */
+  struct [[deprecated]] configureVFATs : public xhal::common::rpc::Method
```
lpetre-ulb, line 145: Don't you think it's time to drop `v2` electronics compatibility?

spec.template
```diff
@@ -0,0 +1,186 @@
...
+
+# %pre
+
+%prep
+## if there is a Source tag that points to the tarball
+#%%setup -q
+mv %{_sourcedir}/%{_project}-%{_longpackage}-%{_version}-%{_short_release}.tbz2 ./
+tar xjf %{_project}-%{_longpackage}-%{_version}-%{_short_release}.tbz2
+
+## update extracted timestamps if doing a git build
```
lpetre-ulb, line 88: How can we do a non-Git build which is, IMHO, the best way to build a RPM? We could even use `mock` to ensure the dependencies at both compile time and test time.

src/amc.cpp
```diff
@@ -5,296 +5,243 @@
...
 
 #include <chrono>
-#include <string>
 #include <time.h>
 #include <thread>
-#include <vector>
 
-unsigned int fw_version_check(const char* caller_name, localArgs *la)
+namespace amc {
+  auto logger = log4cplus::Logger::getInstance(LOG4CPLUS_TEXT("logger"));
```
lpetre-ulb, line 25: **If** this works reliably, the logger name could maybe be named after the namespace to provide improved configurability at run-time.

src/amc.cpp
```diff
@@ -5,296 +5,243 @@
...
+}
+
+uint32_t amc::fw_version_check(const char* caller_name)
```
lpetre-ulb, line 28: Not related to this PR, but just an idea for compatibility checking. Since this function does no longer seem relevant once the v2 electronics behavior will be completely removed, we could envision to perform a check on the back-end firmware at module loading. The RPC modules would only be loadable for CTP7 firmware releases they have been validated for.

src/amc.cpp
```diff
@@ -5,296 +5,243 @@
...
 
-    //readout sbits
     time_t acquisitionTime,startTime;
     bool acquire=true;
-    startTime=time(NULL);
-    (*maxNetworkSizeReached) = false;
+    startTime=time(nullptr);
     uint32_t l1ADelay;
-    while(acquire) {
-        if ( sizeof(uint32_t) * storedSbits.size() > 65000 ) { //Max TCP/IP message is 65535
```
lpetre-ulb, line 170: Why is this check removed? If I think we all agree to say that the TCP/IP "messages" don't have length limit (and even UDP datagrams can be larger than 65kB), the `rpcsvc` daemon [limits the message size](https://github.com/uwcms/APx-rpcsvc/blob/9d40d066d0c8ab451d9623a13c629e9765049782/run_client.cpp#L7-L11) due to the CTP7 memory constraints. I worry that removing this check without further consideration and tests is a bad idea.

src/amc/blaster_ram.cpp
```diff
@@ -2,334 +2,349 @@
  *  \author Jared Sturdy <sturdy@cern.ch>
  */
 
-#include "amc/blaster_ram.h"
+#include "ctp7_modules/common/amc/blaster_ram.h"
+#include "ctp7_modules/server/amc/blaster_ram.h"
```
lpetre-ulb, line 6: Where is this file supposed to be located at? I cannot find it (nor the other includes from `ctp7_modules/server/*`) anywhere.

src/amc/daq.cpp
```diff
@@ -3,580 +3,263 @@
...
 }
 
-void setDAQLinkRunParameterLocal(localArgs* la, uint8_t const& parN, uint8_t const& rparam)
+void amc::daq::setDAQLinkRunParameter::operator()(uint8_t const& parN, uint8_t const& rparam) const
 {
   if (parN < 1 || parN > 3) {
     std::stringstream errmsg;
     errmsg << "Attempting to set DAQ link run parameter " << parN << ": outside expectation (1-3)";
-    la->response->set_string("error", errmsg.str());
+    // FIXME REMOVE // la->response->set_string("error", errmsg.str());
```
lpetre-ulb, line 220: Shouldn't this method throw instead of silently fail?

src/amc/sca.cpp
```diff
@@ -3,41 +3,54 @@
...
+  utils::writeReg("GEM_AMC.SLOW_CONTROL.SCA.MANUAL_CONTROL.SCA_CMD.SCA_CMD_EXECUTE",0x1);
 }
 
-std::vector<uint32_t> sendSCACommandWithReply(localArgs* la, uint8_t const& ch, uint8_t const& cmd, uint8_t const& len, uint32_t data, uint16_t const& ohMask)
+std::vector<uint32_t> amc::sca::sendSCACommandWithReply::operator()(uint8_t const& ch, uint8_t const& cmd, uint8_t const& len, uint32_t data, uint16_t const& ohMask) const
 {
-  // FIXME: DECIDE WHETHER TO HAVE HERE // if (regExists(la, "GEM_AMC.SLOW_CONTROL.SCA.ADC_MONITORING.MONITORING_OFF")) {
-  // FIXME: DECIDE WHETHER TO HAVE HERE //   uint32_t monMask = readReg(la,"GEM_AMC.SLOW_CONTROL.SCA.ADC_MONITORING.MONITORING_OFF");
-  // FIXME: DECIDE WHETHER TO HAVE HERE //   writeReg(la,"GEM_AMC.SLOW_CONTROL.SCA.ADC_MONITORING.MONITORING_OFF",       0xffffffff);
+  // FIXME: DECIDE WHETHER TO HAVE HERE // if (utils::regExists("GEM_AMC.SLOW_CONTROL.SCA.ADC_MONITORING.MONITORING_OFF")) {
```
lpetre-ulb, line 48: Would probably be better not to support the old CTP7 firmware releases and set a minimum requirement. See the above proposal related to the FW version check.

src/extras.cpp
```diff
@@ -3,166 +3,188 @@
...
+//     }
+
+//     xhal::common::rpc::registerMethod<extras::mfiforead>(modmgr);
+//     xhal::common::rpc::registerMethod<extras::mblockread>(modmgr);
+//     xhal::common::rpc::registerMethod<extras::mlistread>(modmgr);
+//     xhal::common::rpc::registerMethod<extras::mfifowrite>(modmgr);
+//     xhal::common::rpc::registerMethod<extras::mblockwrite>(modmgr);
+//     xhal::common::rpc::registerMethod<extras::mlistwrite>(modmgr);
+//   }
+// }
```
lpetre-ulb, line 190: Why are these RPC methods all commented in the PR? Since `module_init` does no longer exist, I fear that loading the RPC module `extras` will be impossible. Although, I agree that `extras` it might not be useful remotely.

src/memory.cpp
```diff
@@ -1,45 +1,70 @@
...
+    struct mread : public xhal::common::rpc::Method
+    {
+        std::vector<uint32_t> operator()(const uint32_t &addr, const uint32_t &count) const;
+    };
+
+    struct mwrite : public xhal::common::rpc::Method
+    {
+        void operator()(const uint32_t &addr, const std::vector<uint32_t> &data) const;
+    };
+};
```
lpetre-ulb, line 25: How are those RPC methods called if they are not declared in an header file?

src/optical.cpp
```diff
@@ -1,91 +1,118 @@
...
 
 memsvc_handle_t memsvc;
 
-void measure_input_power(const RPCMsg *request, RPCMsg *response) {
+namespace optical {
+    struct measure_input_power : public xhal::common::rpc::Method
+    {
+        std::map<std::string, std::vector<uint32_t>> operator()() const;
+    };
+};
```
lpetre-ulb, line 27: Same question as for the `memory` module here.

src/optohybrid.cpp
```diff
@@ -1,615 +1,474 @@
...
-        map_regValues[scanBase + ".CONF.MASK"] = 0;
-    }
-    else{
-        map_regValues[scanBase + ".CONF.CHIP"] = 0;
+std::map<std::string, std::vector<uint32_t>> oh::statusOH::operator()(const uint32_t &ohMask) const
+{
+  LOG4CPLUS_INFO(logger, "Reeading OH status registers");
+
+  std::array<std::string, 23 + 1> regs = {
+    "CFG_PULSE_STRETCH", // FIXME seems like a copy/paste error?
```
lpetre-ulb, line 191: Indeed, it seems. ;) This list will also have to be revised since some of those registers no longer exist.

src/utils.cpp
```diff
@@ -252,33 +332,39 @@ uint32_t readAddress(lmdb::val & db_res, RPCMsg *response)
...
-    writeAddress(db_res, value, la->response);
+  if (regExists(regName, &db_res)) {
+    utils::writeAddress(db_res, value);
   } else {
-    LOGGER->log_message(LogManager::ERROR, stdsprintf("Key: %s is NOT found", regName.c_str()));
-    la->response->set_string("error", "Register not found");
+    std::stringstream errmsg;
+    errmsg << "Key: " << regName << " was NOT found";
+    LOG4CPLUS_ERROR(logger, errmsg.str());
+    throw std::runtime_error(errmsg.str());
```
lpetre-ulb, line 345: Here, as in many other places, logging **and** throwing is probably counterproductive. If an exception is not caught, `xhal::rpc` is planned to log it. If the exception is caught and the system can recover (e.g. during a GBT phase scan), there is no need to log the errors.

lpetre-ulb (commented):
Many comments which are not a good reason to postpone this PR. However, I cannot compile it seamlessly on `gemdaq-build-xdaq15`. This is possibly an issue with the VM, but I cannot really approve the PR if I cannot successfully compile the code. :-/

So, related to build system, here is a list a issues/flaws I encountered: 
* The `build` target is not defined, while it existed in this repository and still exists in `cmsgemos`.
* It is required to set the `PETA_STAGE` environment variable while I would have expected a sesible default value if the `gem-peta-stage-ctp7` RPM is installed:
``` bash
[lpetre@gemdaq-build-xdaq15 ctp7_modules]$ make
TargetArch=arm make -f ctp7_modules.mk build
Using BUILD_HOME=/afs/cern.ch/user/l/lpetre/GEM
OS Detected: centos7
BUILD_VERSION 1.0.462.dev
PREREL_VERSION -final.dev462
make[1]: Entering directory `/afs/cern.ch/user/l/lpetre/GEM/ctp7_modules'
/afs/cern.ch/user/l/lpetre/GEM/ctp7_modules/config/mfZynq.mk:2: *** "Error: PETA_STAGE environment variable not set.".  Stop.
make[1]: Leaving directory `/afs/cern.ch/user/l/lpetre/GEM/ctp7_modules'
make: *** [arm] Error 2
```
* Does compile on `gemdaq-build-xdaq15`:
``` bash
[lpetre@gemdaq-build-xdaq15 ctp7_modules]$ make
TargetArch=arm make -f ctp7_modules.mk build
[...]
XHAL_ROOT /opt/xhal
BUILD_HOME /afs/cern.ch/user/l/lpetre/GEM
make[2]: Entering directory `/afs/cern.ch/user/l/lpetre/GEM/ctp7_modules'
mkdir -p arm/src/linux/arm
arm-linux-gnueabihf-g++ -fomit-frame-pointer -pipe -fno-common -fno-builtin -Wall -std=c++14 -march=armv7-a -mfpu=neon -mfloat-abi=hard -mthumb-interwork -mtune=cortex-a9 -DEMBED -Dlinux -D__linux__ -Dunix -fPIC --sysroot=/opt/gem-peta-stage/ctp7/ -pthread -DGEM_VARIANT="ge11" -std=c++14 -std=gnu++14 -g -O2 -c -Iinclude -I/opt/xhal/include -I/opt/wiscrpcsvc/include -I/opt/reedmuller/include -MT arm/src/linux/arm/memhub.o -MMD -MP -MF arm/src/linux/arm/memhub.Td -o arm/src/linux/arm/memhub.o src/memhub.cpp
cc1plus: fatal error: /opt/wiscrpcsvc/include/stdc-predef.h: Permission denied
[...]
```
I'm not sure why `stdc-predef.h` is searched for in `/opt/wiscrpcsvc/include`, but the directory seems to be in a bad state:
``` bash
[lpetre@gemdaq-build-xdaq15 ctp7_modules]$ ll -a /opt/wiscrpcsvc/include/
ls: cannot access /opt/wiscrpcsvc/include/.: Permission denied
ls: cannot access /opt/wiscrpcsvc/include/..: Permission denied
ls: cannot access /opt/wiscrpcsvc/include/wiscRPCMsg.h: Permission denied
ls: cannot access /opt/wiscrpcsvc/include/wiscrpcsvc.h: Permission denied
total 0
d????????? ? ? ? ?            ? .
d????????? ? ? ? ?            ? ..
-????????? ? ? ? ?            ? wiscRPCMsg.h
-????????? ? ? ? ?            ? wiscrpcsvc.h
```

Also, more generally:
* What is the purpose of passing integral types as constant references?
* Shouldn't the Doxygen commands be changed from `\` to `@` for consistency with `cmsgemos`?
* Does this PR really targets `develop`? I would have expected it to target `feature/templated-rpc-methods`.

None of this is blocking, except that I did not manager to compile the PR on the dedicated VM.

.gitlab-ci.yml
jsturdy: this was a workaround until the linaro toolchain was put into the generic container.
the bind mount worked on the specific runner i had set up, but it is indeed no longer necessary

.gitlab-ci.yml
jsturdy: hmm, good question... it doesn't exist publicly because i had hoped to get Jes to provide this package for us with each image build.
I used the following to generate the RPM:
```
rpmbuild -vv -bb -bl --buildroot=$PWD/RPMBUILD/BUILDROOT --define "_requires 0" --define "_release 1.cc7" --define "_build_requires 0" --define  "_topdir /data/bigdisk/users/sturdy/software/gem-peta-stage/rpm/RPMBUILD" /data/bigdisk/users/sturdy/software/gem-peta-stage/rpm/gem-peta-stage.spec --target "noarch" --define "debug_package %{nil}"
```

.gitmodules
jsturdy: example...?
once things are completely moved to gitlab, the CI runner will have access to clone via ssh so it could be reverted, this change was made only because the test CI job on gitlab failed to clone from github

.gitlab-ci.yml
jsturdy: the `github_changelog_generator` tool works pretty well, though at present the `gitlab` support hasn't been merged in (which isn't a problem provided a mirror to `github` always exists)

ctp7_modules.mk
jsturdy: changes here are really minimally done, i wasn't expecting to have to change the compiler away from the default.
could do the flag derivation based on compiler, provided there aren't any downsides, which would be simpler to maintain long term, or drop the option entirely and rely on the compiler default

ctp7_modules.mk
jsturdy: they actually should be, so this check/message is *probably* unnecessary

ctp7_modules.mk
jsturdy: not any more, as per some [thread](https://github.com/cms-gem-daq-project/xhal/issues/132#issuecomment-528561345) in the past about keeping things together

include/ctp7_modules/common/daq_monitor.h
jsturdy: future, this PR was really just to convert current code minimally to the templated framework

include/ctp7_modules/common/optohybrid.h
jsturdy: yes, see other comments about scope of this PR

spec.template
jsturdy: when things are humming along flawlessly with an operational system in P5, then migrating to `mock` and `cmake` could possibly be pursued (or when i'm no longer working on GEMs :-D)

src/amc.cpp
jsturdy: this was simply to get things to compile.
the promise was that the logger instantiation would be handled elsewhere in a better way...

src/amc.cpp
jsturdy: if so, that should be handled in the translation layer, not in "user" code
the code shouldn't care at all whether it's trying to send a GB of data or not

src/amc/blaster_ram.cpp
jsturdy: hmm, looks like they were not added... oops!
will force push an amended commit

src/amc/daq.cpp
jsturdy: possibly, i don't remember if i didn't consider that (trying to set an incorrect parameter) a critical error, but this function has to be reworked for new FW anyway

src/extras.cpp
jsturdy: i think precisely because `extras` wasn't supposed to be an externally facing library

src/memory.cpp
jsturdy: good question, follow-up, are they necessary?

src/optical.cpp
jsturdy: is this needed remotely?

.gitmodules
lpetre-ulb: You can use:
```
 [submodule "config"]
	path = config
	url = .../gembuild.git 
```
With relative URLs, the protocol will be the same as for the parent repository. I consider forcing the protocol when not required as bad practice.

However, in this case, it might be necessary. It depends how you setup the GitLab CI. If the GitLab CI uses the external GitHub repository it will work. If the CI runs on the GitLab mirror, the risk is to trigger a race condition between the repository updates.

.gitlab-ci.yml
lpetre-ulb: Okay. Wouldn't it be good to host all the extra spec files in a dedicated repository? I'm thinking about `gem-peta-stage-ctp7` (if necessary), `tabulate` (why not using the EPEL package?), ... The compilation doesn't need to be fully automatized since it will rarely change.

.gitlab-ci.yml
lpetre-ulb: I'm happily surprised that an automatic tools can generate a sensible and useful changelog. :-)

jsturdy:
> Many comments which are not a good reason to postpone this PR. However, I cannot compile it seamlessly on `gemdaq-build-xdaq15`. This is possibly an issue with the VM, but I cannot really approve the PR if I cannot successfully compile the code. :-/
> 

I have actually never tested this on the `xdaq15` VM because there is no explicit dependency on `xdaq15` (the **only** package which has that is `cmsgemos`, everything else simply reuses `xdaq` provided packages, e.g., `log4cplus`, and *should* be able to compile regardless of the provided version)

> So, related to build system, here is a list a issues/flaws I encountered:
> 
>   * The `build` target is not defined, while it existed in this repository and still exists in `cmsgemos`.

`build` still exists for a specific target, e.g.,
```
TargetArch=arm make build -f ctp7_modules.mk
```

> 
>   * It is required to set the `PETA_STAGE` environment variable while I would have expected a sesible default value if the `gem-peta-stage-ctp7` RPM is installed:
> 
> 
> ```shell
> [lpetre@gemdaq-build-xdaq15 ctp7_modules]$ make
> TargetArch=arm make -f ctp7_modules.mk build
> Using BUILD_HOME=/afs/cern.ch/user/l/lpetre/GEM
> OS Detected: centos7
> BUILD_VERSION 1.0.462.dev
> PREREL_VERSION -final.dev462
> make[1]: Entering directory `/afs/cern.ch/user/l/lpetre/GEM/ctp7_modules'
> /afs/cern.ch/user/l/lpetre/GEM/ctp7_modules/config/mfZynq.mk:2: *** "Error: PETA_STAGE environment variable not set.".  Stop.
> make[1]: Leaving directory `/afs/cern.ch/user/l/lpetre/GEM/ctp7_modules'
> make: *** [arm] Error 2
> ```
> 

This is now "fixed", provided the fix (`PETA_STAGE=$PETA_PATH/$BOARD`) is robust enough to handle the potential targets

>     * Does compile on `gemdaq-build-xdaq15`:
> 
> 
> ```shell
> [lpetre@gemdaq-build-xdaq15 ctp7_modules]$ make
> TargetArch=arm make -f ctp7_modules.mk build
> [...]
> XHAL_ROOT /opt/xhal
> BUILD_HOME /afs/cern.ch/user/l/lpetre/GEM
> make[2]: Entering directory `/afs/cern.ch/user/l/lpetre/GEM/ctp7_modules'
> mkdir -p arm/src/linux/arm
> arm-linux-gnueabihf-g++ -fomit-frame-pointer -pipe -fno-common -fno-builtin -Wall -std=c++14 -march=armv7-a -mfpu=neon -mfloat-abi=hard -mthumb-interwork -mtune=cortex-a9 -DEMBED -Dlinux -D__linux__ -Dunix -fPIC --sysroot=/opt/gem-peta-stage/ctp7/ -pthread -DGEM_VARIANT="ge11" -std=c++14 -std=gnu++14 -g -O2 -c -Iinclude -I/opt/xhal/include -I/opt/wiscrpcsvc/include -I/opt/reedmuller/include -MT arm/src/linux/arm/memhub.o -MMD -MP -MF arm/src/linux/arm/memhub.Td -o arm/src/linux/arm/memhub.o src/memhub.cpp
> cc1plus: fatal error: /opt/wiscrpcsvc/include/stdc-predef.h: Permission denied
> [...]
> ```
> 
> I'm not sure why `stdc-predef.h` is searched for in `/opt/wiscrpcsvc/include`, but the directory seems to be in a bad state:
> 
> ```shell
> [lpetre@gemdaq-build-xdaq15 ctp7_modules]$ ll -a /opt/wiscrpcsvc/include/
> ls: cannot access /opt/wiscrpcsvc/include/.: Permission denied
> ls: cannot access /opt/wiscrpcsvc/include/..: Permission denied
> ls: cannot access /opt/wiscrpcsvc/include/wiscRPCMsg.h: Permission denied
> ls: cannot access /opt/wiscrpcsvc/include/wiscrpcsvc.h: Permission denied
> total 0
> d????????? ? ? ? ?            ? .
> d????????? ? ? ? ?            ? ..
> -????????? ? ? ? ?            ? wiscRPCMsg.h
> -????????? ? ? ? ?            ? wiscrpcsvc.h
> ```
> 

I need to check the RPM, the permissions are set wrong, probably because of my `umask` when building the RPM...

Again, I've never tested on that VM (I've only used `gem904daq02` with locally installed versions of necessary dependencies (`xhal`, and `reedmuller` for the `PETA_STAGE` library, need to provide the equivalent `reedmuller-peta-ctp7` package...) (note about doing this in a separate comment)

> Also, more generally:
> 
>   * What is the purpose of passing integral types as constant references?

I think it was culled from your example
> 
>   * Shouldn't the Doxygen commands be changed from `\` to `@` for consistency with `cmsgemos`?
> 

No change to the style was made to this codebase, I don't really have a strong preference, but if a future `clang-tidy`/`clang-format` could apply this globally, let me know and i'll add it to the style file i'm preparing (it will be added to `gembuild` along with a couple of targets to perform the actions).

>   * Does this PR really targets `develop`? I would have expected it to target `feature/templated-rpc-methods`.
> 

indeed, i had probably misread my graph and thought that `feature/templated-rpc-methods` was behind `develop` but fast-forwardable.

> 
> None of this is blocking, except that I did not manager to compile the PR on the dedicated VM.

jsturdy:
N.B., for building against non-system installed packages, which have ARM libraries, you should do the following:

```
## install local xhal
INSTALL_PREFIX=/local/install/path make install
## install local reedmuller-c
INSTALL_PREFIX=/local/install/path make install
cd /local/install/path/opt/gem-peta-stage/ctp7
ln -s /opt/gem-peta-stage/ctp7/lib .
ln -s /opt/gem-peta-stage/ctp7/usr .
## do the same for other boards, if they exist
cd /path/to/ctp7_modules
export PETA_PATH=/opt/gem-peta-stage
make
## or make only a specific target
make arm
```

.gitlab-ci.yml
jsturdy: yeah, that could probably be done, the python packages are just downloaded from pypi and built with some appropriate flags, but i suppose that could be scripted...
i'd really really prefer to push Jes to provide the peta stage devel package (headers *and* libraries for *all* internal dependencies) for CTP7, APX, etc (and of course add the emulator "stages" as well)

ctp7_modules.mk
lpetre-ulb: I think it would be best to _chose_ a C++ version and work with it. C++98 is not C++11 which is itself not C++17 or C++20.

.gitlab-ci.yml
jsturdy: check out the [repo](https://gitlab.com/github-changelog-generator/github-changelog-generator) for some examples :-)
it parses PRs and issues, so the onus is on ensuring that the titles have meaningful content (something i've harped on in the past ;-) )

ctp7_modules.mk
lpetre-ulb: Ooh, okay, I remembered they were still installed in `/opt/<package>`. I prefer this new solution though. ;)

include/ctp7_modules/common/optohybrid.h
lpetre-ulb: Yes, sure, it is outside the scope of this PR. This is just to get a written record and make sure that the idea is shared.

spec.template
lpetre-ulb: Notice that I no longer come back with `cmake` (although I'd love to see it used here :-p). `mock`, on the other hand, should be easier to add and would ensure the correct dependencies. All Fedora packages are built with it; how difficult can it be to use it?

src/amc.cpp
lpetre-ulb: Indeed, but we are mostly limited by the `rpcsvc` daemon in this case. We could sure workaround it in the templated RPC framework, but the easiest would be to simply throw an exception if the message is too big.

src/optical.cpp
lpetre-ulb: I think we will want to monitor the RX power during operations like it is currently done during commissioning.

spec.template
jsturdy: > Notice that I no longer come back with `cmake` (although I'd love to see it used here :-p). `mock`, on the other hand, should be easier to add and would ensure the correct dependencies. All Fedora packages are built with it; how difficult can it be to use it?

the eternal question ;-)
i did try to set it up at some point, but abandoned it because it was taking more time than i was interested in spending at the time :-D

src/memory.cpp
lpetre-ulb: They are currently used in `gem_reg.py` through `librwreg.so` and `librpcman.so`, but these libraries are probably going to die with the future framework. Ideally, I would prefer the DAQ SW to not be able to perform raw reads and writes. I don't have strong opinion whether a compatibility layer should be temporarily provided.

.gitlab-ci.yml
lpetre-ulb: Indeed, I downloaded most of (all?) the Python packages from PyPI to install the ULB DAQ machine. And the packages work. If there are no other packages and that the flags are not too special, I guess knowing this is enough.

.gitlab-ci.yml
lpetre-ulb: Indeed, the examples look really nice (provided that the issues and PR have proper titles as you say). :)

src/optical.cpp
jsturdy: right, but does that functionality need to be exposed through this module, or via some other mechanism in an exported library?

src/memory.cpp
jsturdy: i'd say no, and fix what breaks

src/amc.cpp
lpetre-ulb: Too bad there is no Java-like introspection...

include/ctp7_modules/common/optohybrid.h
jsturdy: could open a new issue :-)

src/memory.cpp
lpetre-ulb: :+1:

src/optical.cpp
lpetre-ulb: I believe everything should be exported through the RPC framework. I think it can be done with this module with a different implementation for each back-end.

src/amc.cpp
jsturdy: but the problem i see with this is that the message *might* be big by design, *unless* some type of an error occurred.
so if this can happen in some types of transaction doing who knows what (returning a ton of monitoring variables, reading back configuration RAM), everything there has to be chunked somehow, because the design *is* to return all that data, and i think it is problematic to want to handle this *in* the method doing the action...

lpetre-ulb: I agree that the message size should not matter to the user and should be, if required, fragmented in the lower layers.

However, this is not how the `rpcsvc` framework/service is designed; the messages must be kept relatively small, under 8 MiB. You could probably work around it in the templated RPC framework; it would probably be easier and faster to re-write the `rpcsvc` daemon. Also, the RPC calls are not designed as a streaming service, so that the whole message must be buffered at some point... but the memory is limited in the CTP7.

lpetre-ulb:
> I have actually never tested this on the `xdaq15` VM because there is no explicit dependency on `xdaq15` (the **only** package which has that is `cmsgemos`, everything else simply reuses `xdaq` provided packages, e.g., `log4cplus`, and _should_ be able to compile regardless of the provided version)

I wrongly thought that all new developments were focusing towards `xDAQ 15` and so were done on the VM (among others because of the `log4cplus` potential issue you cite).

> > * It is required to set the `PETA_STAGE` environment variable while I would have expected a sesible default value if the `gem-peta-stage-ctp7` RPM is installed:
>
> This is now "fixed", provided the fix (`PETA_STAGE=$PETA_PATH/$BOARD`) is robust enough to handle the potential targets

:+1:

> > * What is the purpose of passing integral types as constant references?
>
> I think it was culled from your example

Right, copies are avoided in the templated RPC framework when references are used. However, this is useless for the simple types for which the reference accesses cost more than the copy itself.

> > * Shouldn't the Doxygen commands be changed from `\` to `@` for consistency with `cmsgemos`?
>
> No change to the style was made to this codebase, I don't really have a strong preference, but if a future `clang-tidy`/`clang-format` could apply this globally, let me know and i'll add it to the style file i'm preparing (it will be added to `gembuild` along with a couple of targets to perform the actions).

I don't have a strong preference either, I just remembered having seen the conversion being done somewhere. If there is a survey for `\ vs @`, I'd vote `\` (not coming from Java world... :)

> N.B., for building against non-system installed packages, which have ARM libraries, you should do the following:
> [...]

I also had to export `XHAL_ROOT` to point to the `xhal` directory in `INSTALL_PREFIX`, but thanks for the instructions! Will some documentation be available somewhere? The process to setup a local environment doesn't seem trivial (but it is already much simpler than what currently exists).

lpetre-ulb (approved):
Approved to test the modules further since the code compiles within the new framework.

src/amc.cpp
jsturdy: OK, and since this is actually such a critical feature, #168 opened to try to find an actual solution for the cases where we may need to send/receive that much data.

jsturdy:
> I also had to export `XHAL_ROOT` to point to the `xhal` directory in `INSTALL_PREFIX`, but thanks for the instructions! Will some documentation be available somewhere? The process to setup a local environment doesn't seem trivial (but it is already much simpler than what currently exists).

Ah, yeah, anything that is not system installed needs to be specified, as the defaults pick up the system versions.
I think once all these things are merged in an updated doc can be provided, being based on the comment here

mexanick (approved):

lpetre-ulb (approved):

PR #169, opened 2020-03-08
Minimal working version of the templated RPC methods

### LINKED ISSUE #170: Functions still relying on 0xdeaddead
## Brief summary of issue

The template RPC framework supports exceptions to signal failures, and the `readReg` functions were rewritten to throw on failure instead of returning `0xdeaddead`. Two places in `utils.cpp` still compare against `0xdeaddead` to check for errors: https://github.com/cms-gem-daq-project/ctp7_modules/blob/feature/templated-rpc-methods/src/utils.cpp#L527 and https://github.com/cms-gem-daq-project/ctp7_modules/blob/feature/templated-rpc-methods/src/utils.cpp#L494

### Types of issue

- [x] Bug report (report an issue with the code)
- [ ] Feature request (request for change which adds functionality)

## Expected Behavior

`0xdeaddead` doesn't have a special meaning

## Current Behavior

`repeatedRegRead` and `writeReg` will fail on `0xdeaddead`

## Possible Solution (for bugs)

Use `try`...`catch`

Noticed while reading the code

## Your Environment

* Version used: `feature/templated-rpc-methods` @ 7a6901b
* Shell used: Github

lpetre-ulb:
## Description

This PR aims at providing a minimal working set of templated RPC methods.

The following changes were made:

* Remove the (de)serializers which are going to be implemented as part of https://github.com/cms-gem-daq-project/xhal/issues/138 (PR https://github.com/cms-gem-daq-project/xhal/pull/144 in progress);
* Fix missing namespaces in `src/vfat3.cpp`, `src/utils.cpp` and `src/calibration_routines.cpp` leading to link failure at run-time;
* Fix incorrect usage of LMDB leading to SIGV as soon as the `db_res` parameter of `utils::regExists` was used. Please read the commit message if more information is required. Note that since the templated RPC migration, the LMDB database is opened and closed for every LMDB register read;
* Delete the now useless `LocalArgs` structure;
* Delete the now useless `EMIT_RPC_ERROR` macro;
* Fix #170 by removing all references to `0xdeaddead` from `src/utils.cpp`;
* Disable the `x84_64` target since the GLIB sysroot is not yet ready;
* Update the `config` submodule to get proper compilation;
* Replace `int` by `std::uint32_t` in `amc/ttc` since the former is not a (de)serializable type.

[Requires https://github.com/cms-gem-daq-project/xhal/pull/144 so it is probably better to wait before merging.]

### Types of changes

## Motivation and Context

Port `cmsgemos` to the templated RPC modules in order to, at first, remove the direct IPBus dependency.

## How Has This Been Tested?

Can successfully run a basic RPC call (i.e., `vfat3::getVFAT3ChipIDs`):
``` bash
$ cat test.cpp
#include "xhal/common/rpc/call.h"
#include "ctp7_modules/common/vfat3.h"
#include <iostream>

int main(int argc, char **argv)
{
        wisc::RPCSvc conn;
        conn.connect("eagle48");
        conn.load_module("vfat3", "vfat3 v1.0.1");

        for (auto &i: xhal::common::rpc::call<vfat3::getVFAT3ChipIDs>(conn, 0, 0x0, false))
                std::cout << i << std::endl;

        return 0;
}
$ ./run.sh
6241
6415
6210
6185
6534
[...]
```

## Checklist:

### DIFF
--- a/include/ctp7_modules/common/amc/ttc.h
+++ b/include/ctp7_modules/common/amc/ttc.h
@@ -67,7 +67,7 @@ namespace amc {
      */
     struct checkPLLLock : public xhal::common::rpc::Method
     {
-      int operator()(int readAttempts) const;
+      std::uint32_t operator()(const std::uint32_t readAttempts) const;
     };
 
     /*!
@@ -81,7 +81,7 @@ namespace amc {
      */
     struct getMMCMPhaseMean : public xhal::common::rpc::Method
     {
-      float operator()(int readAttempts) const;
+      float operator()(const std::uint32_t readAttempts) const;
     };
 
     /*!
@@ -95,7 +95,7 @@ namespace amc {
      */
     struct getGTHPhaseMean : public xhal::common::rpc::Method
     {
-      float operator()(int readAttempts) const;
+      float operator()(const std::uint32_t readAttempts) const;
     };
 
     /*!
@@ -107,7 +107,7 @@ namespace amc {
      */
     struct getMMCMPhaseMedian : public xhal::common::rpc::Method
     {
-      float operator()(int readAttempts) const;
+      float operator()(const std::uint32_t readAttempts) const;
     };
 
     /*!
@@ -119,7 +119,7 @@ namespace amc {
      */
     struct getGTHPhaseMedian : public xhal::common::rpc::Method
     {
-      float operator()(int readAttempts) const;
+      float operator()(const std::uint32_t readAttempts) const;
     };
 
     /*!
--- a/include/ctp7_modules/server/utils.h
+++ b/include/ctp7_modules/server/utils.h
@@ -78,18 +78,6 @@ namespace utils {
     void initLogging();
 
     /*!
-     *  \brief This macro is used to terminate a function if an error occurs.
-     *         It logs the message  and returns the \c error_code value.
-     *
-     *  \param \c message The \c std::string error message.
-     *  \param \c error_code Value that is passed to the \c return statement.
-     */
-#define EMIT_RPC_ERROR(message, error_code) {           \
-      LOG4CPLUS_ERROR(logger, message);                 \
-      return error_code;                                \
-    }
-
-    /*!
      *  \brief return 1 if the given bit in word is 1 else 0
      *
      *  \param \c word: an unsigned int of 32 bit
--- a/include/ctp7_modules/server/utils.h
+++ b/include/ctp7_modules/server/utils.h
@@ -35,15 +35,6 @@ log4cplus.appender.syslog.layout.ConversionPattern= %h[%i] - %M - %m
 
 namespace utils {
 
-    /*!
-     *  \brief Contains arguments required to execute an LMDB transaction
-     */
-    struct LocalArgs
-    {
-        lmdb::txn &rtxn; ///< LMDB transaction handle
-        lmdb::dbi &dbi;  ///< LMDB individual database handle
-    };
-
     static constexpr uint32_t LMDB_SIZE = 1UL * 1024UL * 1024UL * 50UL; ///< Maximum size of the LMDB object, currently 50 MiB
 
     /*!
--- a/include/ctp7_modules/server/utils.h
+++ b/include/ctp7_modules/server/utils.h
@@ -11,7 +11,6 @@
 #include <sstream>
 #include <string>
 #include <vector>
-#include <experimental/optional>
 
 extern memsvc_handle_t memsvc; /// \var global memory service handle required for registers read/write operations
 /* extern log4cplus logger; /// \var global logger */
@@ -104,9 +103,9 @@ namespace utils {
      *
      *  \param \c regName Register name
      *
-     *  \return An optional \c std::string containing the content of the LMDB register, if it exists
+     *  \return A \c std::vector<std::string> containing the parsed content of the LMDB register. Is empty if the register does not exist.
      */
-    std::experimental::optional<std::string> regExists(const std::string &regName);
+    std::vector<std::string> regExists(const std::string &regName);
 
     /*!
      *  \brief Returns an address of a given register
--- a/include/ctp7_modules/server/utils.h
+++ b/include/ctp7_modules/server/utils.h
@@ -200,7 +200,7 @@ namespace utils {
      *  \param \c breakOnFailure stop attempting to read regName before nReads is reached if a failed read occurs
      *  \param \c nReads number of times to attempt to read regName
      */
-    SlowCtrlErrCntVFAT repeatedRegRead(const std::string & regName, bool breakOnFailure=true, uint32_t nReads=1000);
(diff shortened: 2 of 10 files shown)


mexanick (dismissed):
Cursory

mexanick:
> Note that since the templated RPC migration, the LMDB database is opened and closed for every LMDB register read;

Did you have a chance to estimate the performance impact?
Would it make sense to migrate the LMDB opening to the rpcsvc and returning the copy of the handlers (although this can be quite tricky...)

include/ctp7_modules/common/amc/ttc.h
```diff
@@ -67,7 +67,7 @@ namespace amc {
      */
     struct checkPLLLock : public xhal::common::rpc::Method
     {
-      int operator()(int readAttempts) const;
+      std::uint32_t operator()(const std::uint32_t readAttempts) const;
```
jsturdy, line 70: Just to be sure, `int` wasn't being used here for some attempted error handling, was it?

include/ctp7_modules/server/utils.h
```diff
@@ -122,11 +102,18 @@ namespace utils {
     /*!
      *  \brief Returns whether or not a named register can be found in the LMDB
      *
-     *  \param \c la Local arguments structure
      *  \param \c regName Register name
-     *  \param \c db_res pointer to a lmdb::val result that the calling function requires
+     *
+     *  \return An optional \c std::string containing the content of the LMDB register, if it exists
+     */
+    std::experimental::optional<std::string> regExists(const std::string &regName);
```
jsturdy: As long as this is changing return type, why not return a `std::vector<std::string>` (optional or otherwise)?
*Most* places that seem to use it will immediately split it to do some parsing anyway, and it's easy to understand a valid result

include/ctp7_modules/common/amc/ttc.h
lpetre-ulb: No, it wasn't. Actually, the local variable in `checkPLLLock` was already an `std::uint32_t`.

include/ctp7_modules/server/utils.h
lpetre-ulb: I didn't want to assume anything about the LMDB value syntax. Also, _some_ places do not split the value (if the function is only used to check the existence of a register).

lpetre-ulb:
> > Note that since the templated RPC migration, the LMDB database is opened and closed for every LMDB register read;
> 
> Did you have a chance to estimate the performance impact?
> Would it make sense to migrate the LMDB opening to the rpcsvc and returning the copy of the handlers (although this can be quite tricky...)

Not yet, but I'll definitely do so, and open an issue if there is significant impact (that would you be my first guess).

mexanick:
If there's no last-moment comments, I'll merge this in 1 hour. All further discussions can be handled in separate issues.

lpetre-ulb:
> If there's no last-moment comments, I'll merge this in 1 hour. All further discussions can be handled in separate issues.

I would wait for https://github.com/cms-gem-daq-project/xhal/pull/144 to be finished. That PR works, but it uses a code path leading to potential disasters.

include/ctp7_modules/server/utils.h
jsturdy: > I didn't want to assume anything about the LMDB value syntax.

Everywhere the tokenizaiton is happening is already using whatever built in "assumption" (or implementation of the spec) would be made at this step.  However, *if* (for whatever reason) the spec changed, making **this** change would mean that code would only be updated in one place)

> Also, _some_ places do not split the value (if the function is only used to check the existence of a register).

If just checking a registers existence, one simply has to check whether the tokenizaiton resulted in a non-empty vector, if you really also just want something that easily converts to `bool`, you can use `optional<std::vector<std::string>>` and win with the best of both...

lpetre-ulb: Since I didn't like to use `std::experimental`, I followed your idea.The name `regExists` is now a bit awkward, but that should work for now.

mexanick (approved):
Actually, `optional` is not experimental since c++17, but I'm fine with the current solution ;)

lpetre-ulb:
> Actually, `optional` is not experimental since c++17, but I'm fine with the current solution ;)

Right, but the CTP7 cross-compiler does not even fully support C++11... And xDAQ is not certified with C++17...

mexanick:
> > Actually, `optional` is not experimental since c++17, but I'm fine with the current solution ;)
> 
> Right, but the CTP7 cross-compiler does not even fully support C++11... And xDAQ is not certified with C++17...

which one? Old Linaro packaged by Xilinx or fresh Linaro GCC 7.5 (which I believe has full support of c++17) Actually even older Linaro from Xilinx supports c++14 IIRC

lpetre-ulb:
> which one? Old Linaro packaged by Xilinx or fresh Linaro GCC 7.5 (which I believe has full support of c++17) Actually even older Linaro from Xilinx supports c++14 IIRC

The old Linaro package by Xilinx (4.9.2). It indeed supports some C++14 features, but the C++11 standard library support is not complete/compliant.

I'm not sure we want to use a modern Linaro GCC release without thorough tests. We may trigger hard to debug run-time issues when linking against system-wide C++ libraries (particularly with the exceptions). If we can ensure to link only against system-wide C libraries, that should be fine (let's be careful though, `rpcsvc` is C++, but we will probably end up recompiling it...).

jsturdy:
> > which one? Old Linaro packaged by Xilinx or fresh Linaro GCC 7.5 (which I believe has full support of c++17) Actually even older Linaro from Xilinx supports c++14 IIRC
> 
> The old Linaro package by Xilinx (4.9.2). It indeed supports some C++14 features, but the C++11 standard library support is not complete/compliant.
> 
> I'm not sure we want to use a modern Linaro GCC release without thorough tests. We may trigger hard to debug run-time issues when linking against system-wide C++ libraries (particularly with the exceptions). If we can ensure to link only against system-wide C libraries, that should be fine (let's be careful though, `rpcsvc` is C++, but we will probably end up recompiling it...).

Indeed, on the CTP7 I would stick with the compiler that everything else on the card is built with because there very well might be difficult to debug issues; otherwise I'd say use the latest `arm` toolchain compiler which is (was at the time of my last update) `9.2-2019.12`, I think (it should be installable in the docker container with the `getarm.sh` script living in `/opt/arm`)

On the PC, `devtoolset` is safe, because that's explicitly how redhat designed it.
In fact, the current xDAQ team recommendations are actually wrong, and moving forward I'd have no problem, *for our code*, even using `devtoolset-9` for both cc7 and cc8, and targetting `c++17` or `c++2a` compatibility, though for anything we do on both card and PC we might prefer to simply target `c++14`, incomplete though it may be on the `Zynq`, (unless we convince Jes to update the image to a more recent version...)

jsturdy:
> On the PC, `devtoolset` is safe, because that's explicitly how redhat designed it.

Well... *with the provisio that* the machine the code executes on *runs the same version or later* of Centos as the machine that compiled the code

PR #171, opened 2020-03-19
Feature/read write export

mexanick:
## Description

* Refactor `utils` module to achieve better server/common separation
* Add export operators for
  * readRemoteReg
  * readRemoteBlock
  * writeRemoteReg
### Types of changes

## Motivation and Context

These methods are required for `cmsgemos` code in order to remove dependency from `uHAL`
## How Has This Been Tested?

Compile-time checks passed, pending hardware access test
## Checklist:

### DIFF
--- a/include/ctp7_modules/common/utils.h
+++ b/include/ctp7_modules/common/utils.h
@@ -9,9 +9,6 @@
 
 #include "xhal/common/rpc/common.h"
 
-#include <log4cplus/logger.h>
-#include <log4cplus/loggingmacros.h>
-
 #include <ostream>
 #include <iomanip>
 #include <string>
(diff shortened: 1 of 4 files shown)


include/ctp7_modules/server/utils.h
```diff
@@ -76,6 +79,74 @@ namespace utils {
...
+
+    /*!
+     *  \brief Reads a value from register. Register mask is applied. An \c std::runtime_error is thrown if the register cannot be read.
+     *
+     *  \param \c regName Register name
+     *
+     *  \returns \c uint32_t register value
+     */
+     uint32_t readReg(const std::string &regName);
+
```
lpetre-ulb: What is the rationale behind moving this function here? Previously, the functions were organized by level in the stack: raw reads/writes, masked reads/writes and block reads/writes.

src/utils.cpp
```diff
@@ -527,6 +484,77 @@ void utils::writeBlock(const uint32_t &regAddr, const uint32_t *values, const ui
...
+uint32_t utils::readRemoteReg::operator()(const std::string &regName) const
+{
+  return readReg(regName);
+}
+
+std::vector<uint32_t> utils::readRemoteBlock::operator()(const std::string &regName, const uint32_t size, const uint32_t offset) const
+{
+  std::vector<uint32_t> res(size);
+  if (size != readBlock(regName, res.data(), res.size(), offset))
+    throw std::runtime_error("readRemoteBlock:: returned block size doesn't correspond to requested");
```
lpetre-ulb: This check should probably be moved into the `readBlock` function. But that is good enough for now, all read/write functions need a careful review anyway.

include/ctp7_modules/common/utils.h
```diff
@@ -125,5 +63,44 @@ namespace utils {
...
     };
+
+    /*!
+     *  \brief Reads a value from remote register. Exported method. Register mask is applied. An \c std::runtime_error is thrown if the register cannot be read.
+     *
+     *  \param \c regName Register name
+     *
+     *  \returns \c uint32_t register value
+     */
+    struct readRemoteReg : public xhal::common::rpc::Method
```
lpetre-ulb: I would add a `[[deprecated]]` attribute on the "remote" RPC methods since they are implemented only for backward compatibility.

lpetre-ulb (dismissed):
Non-blocking comments which can be addressed in a subsequent PR (but in a separate commit ideally).

lpetre-ulb:
After review of https://github.com/cms-gem-daq-project/cmsgemos/pull/327, `readRemoteBlock` doesn't seem to be required. Could it be not exported at all?

lpetre-ulb (approved):

PR #175, opened 2020-03-30
Remove obsolete RPC modules

lpetre-ulb:
## Description

This PR removes 3 RPC modules which were no longer in use, either not compiled or not exporting any RPC method. The modules are:

* `extras`: functions were not exported and were replaced by functions inside `utils`, 
* `memory`: did not export any used RPC method,
* `rpctest`: did not use the templated RPC framework.

This PR can be merged before #174, in which case it will implement the v2 electronics compatibility removal, or after #174, in which case, this PR will have to be rebased.

### Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to change)

## Motivation and Context

Clean up.

## How Has This Been Tested?

Code still compiles and cms-gem-daq-project/cmsgemos#327 can still be compiled with the PR.

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.

PR #174, opened 2020-03-27
Remove v2 electronics hardware compatibility

lpetre-ulb:
## Description

This PR removes the obsolete code which used to ensure compatibility with the v2 electronics hardware. 

All v2 electronics specific RPC methods have been removed, most of them were already `[[deprecated]]`. Unneeded parameters from `calibration_routines::ttcGenToggle` and `calibration_routines::ttcGenConf` have been removed as well and the documentation updated accordingly.

All code is no longer prevented to work on v2 electronics with the use of `switch` or `if` statements.

### Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to change)

## Motivation and Context

Clean-up.

## How Has This Been Tested?

Code still compiles and https://github.com/cms-gem-daq-project/cmsgemos/pull/327 can still be compiled with the PR.

## Checklist:

- [x] My code follows the code style of this project.
- [x] My change requires a change to the documentation.
- [x] I have updated the documentation accordingly.
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.

PR #177, opened 2020-03-30
Introduce the expert_tools module

lpetre-ulb:
## Description

This PR introduces the `expert_tools` RPC module. As explained in the commit and in https://github.com/cms-gem-daq-project/reg_utils/pull/50, this module is devised to contain all low-level RPC methods to be used in debugging/expert tools.

Currently those two RPC methods are implemented:

* `readRawAddress`
* `writeRawAddress`

The final design for such tools is not yet fixed, but there is an urgent need that our usual tools continue to be functional during the migration. The main advantage for introducing a new module is to clearly separate low-level, potentially dangerous, RPC methods and allow for an easy removal in case it would be deemed necessary.

Note that this PR is based and #174 and #175. It can merged either before, between or after them as desired; I'll make the necessary rebase actions.

### Types of changes

## Motivation and Context

Low-level debugging tools are mandatory. The need will be even increased while migrating to the templated RPC framework.

## How Has This Been Tested?

Tested in combination with https://github.com/cms-gem-daq-project/reg_utils/pull/50. The migrated `gem_reg.py` can be launched and used with the different usual commands without any error.

## Checklist:


mexanick:
General: I'd wait until #176 is merged and retarget there. 
Details: I see you've removed the block read/write low level methods. While they don't have to be exported, it might be interesting to keep them for future operations (although a redesign of the address table accessing mechanism to a giant struct may completely change the way it is invoked). I would not delete them at the moment (but perhaps migrate into server-only part of `expert_tools` module). One interesting application of it would be CTP7 dump or whole system dump function, which is not present at the moment in the `gem_reg` but would be nice to have.
Update: looks like I missed the block access migration to `utils` module. I'm fine with that. Let's just merge the #176 first and retarget to ease the history

mexanick (approved):

